PR #5, opened 2016-03-11
Fix travel moving player back and forth infinitely

tung:
This fixes a bug where hundreds of turns are wasted by the travel system moving the player back and forth when the player targets an unreachable space and sight-blocking obstacles occur in certain formations in-between.  The player will only be stopped if they're interrupted externally, e.g. growing hungry or being hit by a monster.  See the comment in the code for full details.

Based on DynaHack commit 02da53e (Fix travel moving player back and forth repeatedly) by me.

### DIFF
--- a/src/hack.c
+++ b/src/hack.c
@@ -929,7 +929,42 @@ boolean guess;
                     int nx = x + xdir[ordered[dir]];
                     int ny = y + ydir[ordered[dir]];
 
-                    if (!isok(nx, ny))
+                    /*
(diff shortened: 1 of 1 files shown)


davecohrs:
This seems like one of the attempts I tried to fix this bug a few years ago, but found it did not address all situations. There was one in particular, where travel got stuck between 2 traps. Unfortunately, I no longer have that thread or I would forward it.

PR #6, opened 2016-03-11
map_coloring: L's Coloured Walls and Floor v1 patch (adapted)

tung:
This implements the Coloured Walls and Floor v1 patch by L (Leon Arnott):
- http://bilious.alt.org/?296
- http://l.j-factor.com/nethack/glyphcolor.diff

This recolors various features of the dungeon:
- Walls of the Wizard's Tower are magenta.
- Other walls in Gehennom (except the Valley) are red.
- Walls of the Gnomish Mines are brown, except for room walls.
- Walls of the Astral Plane are white.
- Altars are bright blue / gray / black / red depending on alignment, except on the Astral Plane, where they are all bright magenta.

This new feature can be toggled with the new `map_coloring` option (default on).

Compared to the original patch:
- The original patch had no option and was thus unconditional.
- Floors are no longer colored, since it looks ugly.
- Walls of Sokoban are bright blue.
- The Valley of the Dead is no longer grayscale, since it would be an interface screw.
- Beehives are not colored, since it interacts poorly with the way special rooms are handled.
- Lawful altars were white in the patch, but bright blue here, since white and gray are hard to distinguish.
- Room preservation has been adapted for the new level compiler.
- One of the tiny closets in Grotto Town is fixed to be colored as room wall instead of mine wall.
- Orc Town area is considered rooms instead of mine wall.

### DIFF
--- a/dat/mines.des
+++ b/dat/mines.des
@@ -84,6 +84,12 @@ REGION:(00,00,36,16),lit,"ordinary"
 STAIR:levregion(01,03,20,19),(00,00,36,15),up
 STAIR:levregion(61,03,75,19),(00,00,36,15),down
 
+# Define the town area so that it can be colored separately from the
+# surrounding mines.
+#
+# unfilled forces it to be added to the level's rooms list.
+REGION:(02,02,34,16),lit,"ordinary",unfilled
+
 # shame we can't make polluted fountains
 FOUNTAIN:(16,09)
 FOUNTAIN:(25,09)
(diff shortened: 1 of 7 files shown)


moon-chilled:
IMO `map_coloring` should default to off, and that would probably also align more with the DT's values.

paxed:
How does this work with tiles - especially the Wiz tower walls? The others I think already have branch-specific tiles.

tung:
> How does this work with tiles - especially the Wiz tower walls? The others I think already have branch-specific tiles.

This should have no impact on tiles. It modifies `mapglyph()`, which is used almost exclusively by ASCII modes (e.g. TTY windowport, Win32 pseudo-ASCII). The only places `mapglyph()` is used by tiles is to grab special flags to e.g. mark pets with a heart icon, which is unaffected by the color that this feature changes. To windowports that render tiles, this feature is essentially invisible, except for the "map_coloring" option, which makes sense for e.g. Win32 pseudo-ASCII. Windowports that render ASCII don't need to do anything they weren't already doing; Wizard Tower walls are handled entirely inside `mapglyph()`.

Apologies if the details are hazy; I'm not at all familiar with any of the tiles code, and it's been over two months since I've even looked at this code.

webbedspace:
I am Leon and wrote the original patch, which I've decided to no longer keep online due to being for a very old NH version, but I greatly appreciate this pullreq's gesture of support.

arekkusu42:
FWIW I’ve applied this patch as recently as 3.6.2.  Greatly improves the enjoyment of my game.

copperwater:
SpliceHack and xNetHack, both based on 3.7, have up-to-date implementations of this patch. I can provide an updated pull request based on that if desired.

bhaak:
> I am Leon and wrote the original patch, which I've decided to no longer keep online due to being for a very old NH version, but I greatly appreciate this pullreq's gesture of support.

@webbedspace Many of your patches made it (in one form or another) into various forks. E.g. dilapidated armouries, tinfoil hat, and this colored wall patch or also some into vanilla as well, like the picture-painting demon and the polysink patch.

Even though your old website is preserved on the Internet Archive and the patches are for an older version, a live version is much more preferred to show where the original inspiration came from.

PR #7, opened 2016-03-11
Fix Wizard of Yendor stealing quest artifacts

tung:
This removes the biggest reason that players neglect to use their role's quest artifact.  Players can now take advantage of their quest artifact's unique properties and abilities without the downside of losing it after waking Rodney, increasing the uniqueness of different roles.

Quest nemeses are still able to steal quest artifacts.

### DIFF
--- a/src/mhitu.c
+++ b/src/mhitu.c
@@ -1342,11 +1342,13 @@ register struct attack *mattk;
 
     case AD_SAMU:
         hitmsg(mtmp, mattk);
-        /* when the Wizard or quest nemesis hits, there's a 1/20 chance
-           to steal a quest artifact (any, not just the one for the hero's
-           own role) or the Amulet or one of the invocation tools */
-        if (!rn2(20))
-            stealamulet(mtmp);
+        /* when the Wiz hits, 1/20 steals the amulet */
+        /* Quest artifact may be stolen by the quest nemesis. */
+        if (u.uhave.amulet || u.uhave.bell || u.uhave.book || u.uhave.menorah
+            || (u.uhave.questart && mtmp->data->msound == MS_NEMESIS)) {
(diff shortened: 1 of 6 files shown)


src/steal.c
```diff
@@ -506,82 +506,51 @@ register struct obj *otmp;
...
-    if (n > 1) {
-        n = rnd(n);
+    struct obj *otmp = (struct obj *) 0;
+    int real = 0, fake = 0;
+
+    /* select the artifact to steal */
+    if (u.uhave.amulet) {
+        real = AMULET_OF_YENDOR;
+        fake = FAKE_AMULET_OF_YENDOR;
+    } else if (u.uhave.questart && allow_quest_artifact) {
```
scshunt: The check on u.uhave.questart will prevent the Wizard from stealing other artifacts if you don't have your own.

src/mhitu.c
```diff
@@ -1342,11 +1342,13 @@ register struct attack *mattk;
...
         hitmsg(mtmp, mattk);
-        /* when the Wizard or quest nemesis hits, there's a 1/20 chance
-           to steal a quest artifact (any, not just the one for the hero's
-           own role) or the Amulet or one of the invocation tools */
-        if (!rn2(20))
-            stealamulet(mtmp);
+        /* when the Wiz hits, 1/20 steals the amulet */
+        /* Quest artifact may be stolen by the quest nemesis. */
+        if (u.uhave.amulet || u.uhave.bell || u.uhave.book || u.uhave.menorah
+            || (u.uhave.questart && mtmp->data->msound == MS_NEMESIS)) {
```
scshunt: This will still allow nemeses to steal other artifacts if you are carrying your own.

PR #10, opened 2016-03-12
ANSI-fying

MaddTheSane:
This moves NetHack away from K&R to ANSI C calling conventions.


include/flag.h
```diff
@@ -48,7 +48,7 @@ struct flag {
     boolean showexp;         /* show experience points */
     boolean showscore;       /* show score */
     boolean silent;          /* whether the bell rings or not */
-    boolean sortloot;        /* sort items alphabetically when looting */
```
ghost: Unrelated change.

src/dog.c
```diff
@@ -560,9 +551,9 @@ long nmv; /* number of moves */
 }
 
 /* called when you move to another level */
+/* true for ascension or final escape */
```
ghost: Unrelated change

ghost:
This will require some discussion from the DT on whether we want to implement this and thereby drop pre-ANSI compiler support (which has, up until this point, been somewhat of a point of pride for NetHack to support so many platforms). In the meanwhile, please remove all changes that aren't strictly related to the ANSIfication if you wish to help the PR along.

tung:
In case it hasn't been seen, there was also an [/r/nethack submission for this PR](https://www.reddit.com/r/nethack/comments/49woe1/nethack_to_ansi/) with a lot of relevant comments. I posted in that thread, but I'll repeat what I said about this PR here too:

> This is a good thing, because it's the reason NetHack has to declare arg types like this:
> 
> ```
> int FDECL(foo, (SHORT_P, CHAR_P, BOOLEAN_P));
> 
> int
> foo(s, c, b)
> short s;
> char c;
> boolean b;
> {
>     /* ... */
>     return 0;
> }
> ```
> 
> instead of this:
> 
> ```
> int foo(short, char, boolean);
> 
> int
> foo(short s, char c, boolean b)
> {
>     /* ... */
>     return 0;
> }
> ```
> 
> It also thwarts the ability for git to show the containing functions of diff blocks, which makes quickly figuring out where changes were made a big pain. Good stuff!
> 
> > I forgot about FDECL and related macros. I've removed most of them now.
> > 
> > > Just as well, I think `FDECL` is the reason that cscope doesn't work on the NetHack source code, and cscope is essentially god mode for C programming, so on the distant chance this actually gets merged into main line, it will help a _lot_.

include/flag.h
MaddTheSane: This was more a change to make Clang happy, as there are a few places where `sortloot` is treated more like a `char` that a boolean type (I use a custom prefix header that typedefs `boolean` to C99's `bool`).

Will be reverting anyway.

ghost:
Changes to make Clang happy are definitely encouraged, but they should be a separate PR.

MaddTheSane:
Fair enough.

PR #12, opened 2016-03-17
TTY menu enhancements

tung:
Two minor improvements to TTY menus:
1. Stop `>` from closing TTY menus if the menu is scrolled to the last page (it can still be dismissed like this if space is used). This makes scrolling with `<` and `>` safe from the risk of accidentally closing menus.
2. Omit the "a - " part of "a - whatever" from menu coloring, to match NAO343.


ghost:
Note that 2. was reverted out since it was independently fixed by @paxed.

PR #13, opened 2016-03-17
Full text scrolling for TTY text windows

tung:
Up until now, TTY text windows could only be scrolled forward, which is really inconvenient when reading anything longer than a single screen.

This new change allows the standard menu scrolling keys (`^` first page, `|` last page, `<` previous page and `>` next page) to work in TTY text windows, enabling e.g. message history and long-form documentation to be scrolled backwards and forwards.

### DIFF
--- a/win/tty/wintty.c
+++ b/win/tty/wintty.c
@@ -12,6 +12,9 @@
  * to back out the changes. */
 #define H2344_BROKEN
 
+/* Allow text windows to be scrolled with the menu scrolling keys. */
+#define TTY_FULL_TEXT_SCROLL
+
 #include "hack.h"
 
 #ifdef TTY_GRAPHICS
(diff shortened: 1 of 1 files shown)


ghost:
Is there a reason that it's hidden behind a feature macro?

tung:
> Is there a reason that it's hidden behind a feature macro?

So that platforms with extremely limited memory could use the old forward-only system as an alternative. A long time ago, I remember being able to hard-lock NetHackDS (a homebrew port of NetHack for the Nintendo DS) under some circumstances by bringing up some of the longer text windows the game made available via its help menu, for example.

As for _why_ I left that alternative in, I wrote this branch with the intent that it would be accepted by the NetHack DevTeam. To facilitate that, I have to make changes the same way the DevTeam does, with the same baseline as the DevTeam, and all I know of the DevTeam is that they try to support as many platforms as possible to the point of making technical sacrifices, e.g. the widespread use of FDECL and NDECL macros for DOS overlay support, non-standard wrapper macros for things that have existed since C89, etc.

Note that this decision was based on an assumption about how the DevTeam works, because how the DevTeam actually works is done in secret: secret development discussion and a secret bug tracker mean that I have nowhere to gauge or ask what the DevTeam's technical standards for contributions are. I could send an email, but the last email I sent the DevTeam has been ignored for about two months and counting now (it was a suggestion to make the GitHub project announce pushes to #nethack-dev, a change which takes all of five minutes), so there'd be no guarantee of a reply. IRC is a complete crapshoot: nobody DevTeam-related is ever available, except for @paxed who I might squeeze a one- or two-line reply out of before he runs off because he is perpetually busy apparently.

And that is the long-winded explanation as to why this is feature macro'd. If it were up to me, I'd have just replaced the old system wholesale, but who knows if it'd be rejected because somebody on the NetHack DevTeam wants to run it on their VMS and it doesn't work there?

PR #14, opened 2016-03-20
Fix panic when picking up object

bobbydurrett:
The new sortloop code was not properly updating the
head of the list that pointed to objects at a
location on the floor. Modified the code to pass
a pointer the where the head of the list is
stored.


ghost:
This has been fixed in 221a73fe2

bobbydurrett:
It was fixed but I think it was a different commit.  I think it was 3353ad11147045cc2351ea22aa5eb2c7357e4ac9.

PR #19, opened 2016-06-11
Fix screen resize with Linux tty

bobbydurrett:
This fixes the issue that I emailed the devteam about where the screen does not resize properly on NAO. In my own Linux environment I found that the screen resize worked if the SVR4 flag was set in unixconf.h and did not without it. I tracked this down to line 68 of ioctl.c. If SVR4 is set then USE_WIN_IOCTL is set and this enables the resize of the screen. My fix was just to add LINUX as one of the flags checked so you don't have to rely on SVR4 being set when it wouldn't really make sense. I tested this on Oracle Enterprise Linux 6.8 (Redhat 6.8) 32 bit.

### DIFF
--- a/sys/share/ioctl.c
+++ b/sys/share/ioctl.c
@@ -65,7 +65,8 @@ struct termio termio;
 
 #if defined(TIOCGWINSZ)                                    \
     && (defined(BSD) || defined(ULTRIX) || defined(AIX_31) \
-        || defined(_BULL_SOURCE) || defined(SVR4))
+        || defined(_BULL_SOURCE) || defined(SVR4) ||       \
+        defined(LINUX))
 #define USE_WIN_IOCTL
 #include "tcap.h" /* for LI and CO */
 #endif


MaddTheSane:
Some context might be good, especially for those who don't know what `SVR4` is referring to.

SVR4 stands for System V Release 4 of UNIX. It was released in 1988.

bobbydurrett:
Sure, I can give some more context. I was playing Nethack 3.6.0 on NAO using the putty terminal emulator on my Windows 7 laptop. I accidentally resized my screen making it a little smaller. Later as I was trying to put something in my bag of holding I noticed that the screen was updated in the wrong place. It seemed to have not kept track of the fact that my screen was resized. I played with screen resizing more on NAO and find that generally the screen goes black when I shrink my putty window by a line or two as well as the original problem that I noticed. It just doesn't handle putty window resizes well.

Next I went to my own versions of Nethack that I have compiled on a Linux virtual machine. I have a 3.6.0 version and a version based on the fluid 3.6.1 code that is exposed on Github. One of these two installations showed the same behavior as NAO and the other perfectly resized the screen when I shrunk my putty window. I found that the only difference between the two that mattered in terms of whether the screen resized well was the flag called SVR4 in unixconf.h. So, the NAO code could just be recompiled with this flag set and it will resize the screen correctly. It was just random luck that I had the SVR4 flag set on one of my two installs since it didn't make sense for me to set a System 5 Unix flag on a Linux system.

The flags in unixconf.h indicate which version of unix/linux that you have and presumably which features will work on the various kinds of systems. Evidently in ioctl.c there is a function called getwindowsz which determines the number of lines and columns on your screen. But only certain versions of unix must have this ability so the developers used conditional compilation flags to determine whether to include any code in getwindowsz.  The USE_WIN_IOCTL must be set for any code to be in getwindowsz. USE_WIN_IOCTL is set in the logic that I recommend changing. This is the logic in the 3.6.0 version of ioctl.c:

```
#if defined(TIOCGWINSZ)                                    \
    && (defined(BSD) || defined(ULTRIX) || defined(AIX_31) \
        || defined(_BULL_SOURCE) || defined(SVR4))
#define USE_WIN_IOCTL
```

BSD, Ultrix, AIX, and System V Release 4 are all versions of Unix and so I assume that the code in getwindowsz would work on those types of Unix.

But, my version of Linux has this ioctl functionality just as these other forms of Unix do. So, somehow we need the conditions to evaluate to true so that USE_WIN_IOCTL will be set and the code in getwindowsz will be defined.

Long story, but the screen resize is a nice feature and it isn't there right now in the 3.6.0 version on NAO and that's a shame since all the code is in place to make it happen.

Bobby

FredrIQ:
OK, just to make sure, but are you aware of the fact that dgamelaunch (what NAO uses as backend for user management) seems to handle resizing poorly? (Not entirely sure why, it should be none of its business, but I have noticed this myself, something that becomes weird even if I save back to the menu and re-enter).

bobbydurrett:
I'm not aware of the dgamelaunch issues. I was able to reproduce the symptoms that I saw on NAO on my own installation on Linux so it does not seem to be related to dgamelauch. There are two symptoms that I reproduced exactly. One is choosing an item to put into a bag of holding and seeing the wrong line being updated with a plus symbol. This happens after resizing the putty window. The second symptom is that the screen goes black after the resize. I can reproduce both without setting the SVR4 flag and both problems go away with it set. So, I'm assuming it has nothing to do with dgamelaunch but haven't tested running nethack that way.

bobbydurrett:
resizing

I ran into this again today. It looks like in putty under the Window configuration I had "Change the number of rows and columns" chosen under "When the window is resized". I've changed it to "Change the size of the font". This should force it to always be 80x24 no matter what I do to the putty window.  This might be a good workaround for my laptop where no font size fits nicely. I miss the days of running nethack in DOS mode on Windows XP. Anyway, I still recommend this pull request and still recommend setting the SVR4 flag on NAO. It's pretty cool having the game automatically handle resizing for different numbers of rows and columns and the code is there. Just need to set the flag.

nhmall:
This *should* now be resolved in 27a7538e381dbc251b195d0e14fce851bea82f23
However, due to the change in logic from opt-in to opt-out, some platforms that
really do experience trouble with the resulting code that will now be included
may have to deliberately uncomment the
#define AVOID_WIN_IOCTL
in include/unixconf.h to turn it back off.

bobbydurrett:
Great. It is a minor thing but nice to get fixed.

Bobby

bobbydurrett:
Resolved in other commits

bobbydurrett:
I retested it after 27a7538e381dbc251b195d0e14fce851bea82f23t was pushed to GitHub. It fixed the problem in a different way but I think it is good. Minor fix but nice to have the resize code working on Linux.

PR #20, opened 2016-06-30
Nethack 3.6.0.ctype fixes

barthouse:
This fixes a problem with the use of ctype utility functions (e.g. isspace).  The ctype utility functions take an integer and expect them to be in a range of -1 to 255.  If the input is out of range, the ctype library will assert in debug builds of NetHack.  If we call a utility function with a signed char then we risk hitting this assert for extended characters.  All calls to the ctype functions must ensure the argument passed is a uchar to avoid passing in a value that is out of range due to sign extension.

A 100% repro example of this problem can be found by the following:
1. change your language settings to English UK allowing the entry of extended characters (like British pound)
2. Launch new game of nethack with pet using debug version
3. Name pet whose name starts with extended character followed by maximum number of other characters (no spaces)
4. Name item in inventory same as pet
5. Drop named item
6. Move one space and wait for pet to pick up item
7. When pet picks up item, you will assertion in update_topl() when calling isspace().

You can contact me at bart@barthouse.com.


barthouse:
Looks like the changes have been incorporated and this pull request is no longer required.

PR #21, opened 2016-06-30
Nethack 3.6.0.vs2015

barthouse:
This introduces a new method for building and running NetHack on Windows.  This method only requires opening a solution file, building and running.  This method is a significant improvement over the currently supported methods.

This change is generally only adding new project files to the repository with one exception.  When building with Visual Studio 2015 and the latest SDK, we now are running into a name conflict with the C standard runtime.  To avoid this conflict, terminate() was renamed nethack_terminate().


barthouse:
One thing to note.  I placed the new project files in sys/win10.   I always thought it was odd to have the project files under win/win32 since the build method really doesn't have anything to do with windowing support.

I created the new win10 folder to isolate the new support.  Eventually, these files can be moved to be under sys/winnt (perhaps in sys/winnt/build).

barthouse:
Fixed merged conflicts.

PR #24, opened 2016-09-20
Fix ascii art in DEVEL/Developer.txt

moon-chilled:
The top line was aligned wrong, and there was a bunch of extraneous white space.  Not that big of a deal, just a small enhancement.


nhmall:
manually fixed upstream

PR #27, opened 2016-12-20
Added Tony the Tiger message on food that is not a corpse.

bobbydurrett:
Modify eat.c so that Tony the Tiger message comes out when a player polymorphed into a tiger and hallucinating eats yummy food such as a meatball. This message already came out when a hallucinating tiger player ate yummy corpses such as an orc corpse.

### DIFF
--- a/src/eat.c
+++ b/src/eat.c
@@ -1853,7 +1853,7 @@ struct obj *otmp;
                         || otmp->otyp == K_RATION
                         || otmp->otyp == C_RATION)
                         ? "bland."
-                        : Hallucination ? "gnarly!" : "delicious!");
+                        : Hallucination ? ((u.umonnum == PM_TIGER) ? "gr-r-reat" : "gnarly") : "delicious!");
         }
         break; /* default */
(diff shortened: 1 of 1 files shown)


bobbydurrett:
I thought I should add more words to clarify the point of this pull request. When I started playing Nethack 3.6.0 I noticed that the dev team had changed the messages that come out when you eat. If you eat food that you really like it is "delicious". Then I noticed that the funny message that comes out when the player is polymorphed into a Tiger and hallucinating did not come out for all kinds of food that a tiger finds to be good. So, a meatball is "delicious" to a hallucinating tiger, but an orc corpse tastes Grrrreat! or something like that. Anyway, I think this pull request makes this very minor one off case that hardly ever happens more in line with the changes to the food eating messages that came with 3.6.0.

bobbydurrett:
I fixed two issues with this. Needed an exclamation point on the meatball case I originally added. Added case for tripe rations. Hallucinating players polymorphed into tigers is very important. :)

PR #28, opened 2017-02-04
Live logging

NHTangles:
Live logging patch as discussed on #nethack-dev over the last few days.

1. Pulled livelog_write_string(), etc from 3.4.3-nao patch set.
1a. Changed the livelog separator from ':' to '\t' for consistency with xlogfile.
2. Added sysconf option to control what types of messages (if any) to log, and introduced livelog types, which are passed as an additional arg to livelog_write_string(), and compared to the sysconf-configured types to determine whether or not to actually log the message.
3. Added LIVELOG_ENABLE to config.h to allow enable/disable at compile time.
4. Added livelog_printf() as a wrapper to livelog_write_string(), which does the obvious.
5. Added code to actually log messages as per 3.4.3-nao livelogging, and also achievements, unique kills, sacrifice gifts, crownings, broken conducts (there are a lot of these), and new levels.

Please consider :)


NHTangles:
Closing this one to reopen on a feature branch with some improvements.

PR #30, opened 2017-02-17
Add hint file for tty/x11 linux build

nikolas:
This is a hint file that builds a version of NetHack that can use both
tty and X11 window modes.


nhmall:
NetHack-3.7 branch can build a binary with tty, x11, Qt, and curses included, assuming the prerequisite packages are available (see sys/unix/NewInstall.unx)

PR #31, opened 2017-03-27
Livelog

NHTangles:
This is an improved version of the previous livelog submission, and is now in use on hardfought.org.


NHTangles:
Using a hex LIVELOG value in sysconf such as 0xFF does not work. Decimal value like 255 still works fine. I'll commit a fix here when I figure it out.

PR #34, opened 2017-05-06
Achievement recording fixes

NHTangles:
This properly fixes the issues with mines and soko achievement recording, as follows:
1. record_achieve_special is mapped to corpsenm, which has a default value of -1.  Setting to 1 on the prize item and then checking for nonzero will always mark the achievement (current workaround is to also check the level at pickup time, but this has other issues outlined below).
2. only checking level and otyp when setting record_achieve_special opens the possibility that a random luckstone may be generated at mine-end, and this may be interpreted as the prize (this has possibly happened though it's hard to be conclusive as monsters pick things up and move them)
3. checking the level at pickup time is problematic if a monster picks up the prize and takes it to another level before the hero gets his hands on it (cases of this have been recorded).

### DIFF
--- a/src/bones.c
+++ b/src/bones.c
@@ -175,7 +175,14 @@ boolean restore;
             } else if (otmp->otyp == SPE_BOOK_OF_THE_DEAD) {
                 otmp->otyp = SPE_BLANK_PAPER;
                 curse(otmp);
+            /* these might still be flagged as prizes if the late
+               hero did not pick them up */
+            } else if (otmp->otyp == AMULET_OF_REFLECTION
+                       || otmp->otyp == BAG_OF_HOLDING
+                       || otmp->otyp == LUCKSTONE) {
+                otmp->record_achieve_special = 0;
             }
+                
         }
     }
 }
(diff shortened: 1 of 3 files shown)


FredrIQ:
What if the player picks up the luckstone, walks somewhere else, and is killed, and someone else finds the bones? Wouldn't your fixes make the achievement count for them? Your changes doesn't seem to address this (bones.c isn't changed)

NHTangles:
Thanks for looking, FIQ.
This case is taken care of in invent.c - The player picks up the object, the achievement is recorded, and then the record_achieve_special flag is zeroed, so the object no longer has any special significance to the player, or any subsequent players that find it in bones.

NHTangles:
After further discussion with FIQ, it would indeed be possible for a prize item to get to another level without the player picking it up (eg by a pet or another monster) and potentially be left in bones.  Added another commit just now to address this.

Xaleth (changes requested):
Remove unnecessary space in line 185 in bones.c

NHTangles:
Closing this because the issues have been otherwise addressed in the following commits:
58477b33f4e797fab8c8e5202e146a678de448d5
5a50bb64451196433380cd70a884f7791ce2a288

PR #48, opened 2017-09-15
Window ports NetHack and NetHackW changed to use a common engine.dll

barthouse:


barthouse:
Posted just for review.  Pulling out for now and will post newer version soon.

PR #50, opened 2017-09-22
Changes to support game engine reentrancy.

barthouse:
Created this pull request to allow devteam to review approach.  Like to get feedback on general approach.  Very likely we will need to tweak this before accepting.


barthouse:
Thanks FredrIQ.  I'm still new to the engine code (most of my focus has been in port support).  I was hoping we had state to reflect when the monster actions are being carried out.  Being a newb to the engine, should we be concerned about expanding the scope of the check?  For example, I noticed that the flag was being toggled at individual calls to mbhit/buzz and was not just set for the duration of use_offensive.

FredrIQ:
I looked at it a bit closer. m_using is completely redundant with context.mon_moving. If there is a case where one triggers and the other doesn't, then that means there's a bug. You should be able to replace m_using with context.mon_moving in the resist() check in zap.c and remove m_using completely.

barthouse:
Thanks FredrIQ.  I'll go ahead and make that change then.

FredrIQ:
You overlooked removing extern m_using in zap.c

barthouse:
This work is now complete.

barthouse:
I will be incorporate this work using the windows port development branches.  This pull request is no longer needed.

PR #52, opened 2017-09-25
Addressing bug with the use of static dieroll in uhitm.c.

barthouse:
In uhitm.c, we keep a module level static value “dieroll” which is used in various spots in the module.

Because the general “damage monster” routine hmon() can get called from outside uhitm.c, we can call into hmon() without actually changing the value of “dieroll”.  Hmon() in turn calls hmon_hitmon() which in turn uses “dieroll” for some decision making including calling artifact_hit() passing “dieroll”.

So, if you have a vorpal blade and you happen to chop a monsters head off (meaning that you have a “dieroll” of 1) you can then proceed to throw the vorpal blade at other monsters and assuming you hit, you will be guaranteed to chop off the monsters head since “dieroll” will not change when throwing weapons.  (“dieroll” stored in uhitm.c, only changes on certain attacks).

Not that throwing vorpal blades makes a lot of sense …. But demonstrates the bug.

This changes does beg the question in a couple of spots as to what dieroll should we use when we are dropping objects on monsters in stuck traps and when throwing cream pies.

### DIFF
--- a/doc/fixes36.1
+++ b/doc/fixes36.1
@@ -443,11 +443,13 @@ fix buffer overflow in wizard mode for '#' command when 'extmenu' option is on
 "you suddenly vomit" was given two turns before actually vomiting, so you
 	could get that message, move a bit, then get "you can move again"
 	after the 2 turn freeze applied along with the actual vomit
-fix mention_walls reporting secret doors as solid walls
 corpses and other flammable items not subject to direct burning or fire-based
 	erosion which were thrown or dropped into lava remained intact
 if a potion on the floor survived a land mine explosion and got propelled at
 	the hero, it didn't behave like a potion if it hit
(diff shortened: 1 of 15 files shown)


FredrIQ:
This bug is huge, and works in what seems to be every notable version and variant around. Nice find!

barthouse:
This is ready to go I think.

barthouse:
Want to clean up branch history.

k21971:
have confirmed this bug exists in GruntHack and SporkHack (will patch these two soon), as FIQ said this bug probably exists across most all variants. Good job barthouse!

barthouse:
I closed this pull-request replacing it with pull request #54.  (I had done a rebase which made it show up a bit funny here on GitHub).  Re-opening it now since folks are still commenting on it.

PR #54, opened 2017-09-26
Dieroll fix

barthouse:
In uhitm.c, we keep a module level static value **dieroll** which is used in various spots in the module.

Because the general damage monster routine **hmon()** can get called from outside uhitm.c, we can call into **hmon()** without actually changing the value of **dieroll**. **Hmon**() in turn calls **hmon_hitmon**() which in turn uses **dierol**” for some decision making including calling **artifact_hit**() passing **dieroll**.

So, if you have a vorpal blade and you happen to chop a monsters head off (meaning that you have a **dieroll** of 1) you can then proceed to throw the vorpal blade at other monsters and assuming you hit, you will be guaranteed to chop off the monsters head since **dieroll** will not change when throwing weapons. (**dieroll** stored in uhitm.c, only changes on certain attacks).

Not that throwing vorpal blades makes a lot of sense …. But demonstrates the bug.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -2455,7 +2455,7 @@ E void FDECL(check_caitiff, (struct monst *));
 E int FDECL(find_roll_to_hit,
             (struct monst *, UCHAR_P, struct obj *, int *, int *));
 E boolean FDECL(attack, (struct monst *));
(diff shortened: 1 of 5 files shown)


paxed:
Committed.

PR #55, opened 2017-10-01
Fix flashing issue that is seen with STATUS_HILITES in Windows port.

barthouse:
Without this change, tty_status_update() will end up calling putstr() multiple times for the same status line which in turn will cause tty_putstr() to call cl_end() multiple times per status update.  On the Windows port, each call into cl_end() will cause us to immediately call into the console interface.  Since we end up calling cl_end() a couple dozen times for a status update, this causes a noticeable flash flash in the status area.

There are various ways to address this but chose this approach given the various trade offs with other approaches.


paxed:
Committed.

PR #56, opened 2017-10-02
Fix pet AI not following hero as intended

FredrIQ:
Done by cutting down (but not eliminating) use of mtrack:
* Not used for leashed pets
* Not used if <=5 tiles from hero
* Also, tin whistles now clear the mtrack record.


Xaleth:
+1

FredrIQ:
This was merged as a cherry-pick

PR #57, opened 2017-10-03
Improve the handling of rawprint messages

barthouse:
There are several improvements in this pull request.  In particular if during program initialization you encounter problems causing multiple raw_print messages, all the messages will be displayed in a single dialog instead of getting a series of individual message dialogs for each raw_print message.

Fixed an undetected problem with building NetHack where we were defining MSWIN_GRAPHICS.

Implemented a mechanism for dynamically growing a string.  Leveraged this mechanism to reduce code complexity of splash window code.


paxed:
Committed, with minor changes.

PR #59, opened 2017-10-07
Fix problem with defaults.nh not being found when debugging within VS IDE

barthouse:
When debugging using the Visual Studio IDE, by default the working directory is the project directory.
Because of this, both NetHack and NetHackW will fail to find defaults.nh.  This change simply changes
the default working directory to $BinDir allowing defaults.nh to be found along side the executable.


paxed:
Committed.

PR #61, opened 2017-10-07
Fix permanent inventory window going away after pick item(s) operations.

barthouse:
In the windows GUI when permanent inventory is being used, there is an annoying bug where any command that uses the inventory window to pick an item (or items) will cause the inventory window to become hidden requiring the user to use 'i' to redisplay the inventory.

Another side effect of this problem is that a jarring white empty space is created if the user happens to be using window backgrounds other then white (such as black).

This fix is in two parts.  First, we don't hide the permanent inventory window when destroying its as a popup.  Second, after a pick operation where we just used the permanent inventory window we restore the proper display state of the inventory window (pick type NONE and no prompt).


paxed:
Committed.

PR #63, opened 2017-10-09
Fix bug with handling of hilite color NO_COLOR

barthouse:
If a status hilite color is set to NO_COLOR, we will map this color to white (0xffffff) instead of using default foreground color of status window.  This can cause the unexpected result of drawing white text on white in the case that the user has set their status window background to white.

Made a change to the background brush of the main window to be black.  The main window background color is used when we are hiding/showing child windows.  This is a better choice given that the map window's background brush is black.  In the case that all window backgrounds have been set to black, this gives a pleasing consistent user experience without any flashes of white.


paxed:
Committed.

PR #65, opened 2017-10-09
Use a back buffer for all status rendering eliminating all flickering.

barthouse:
We are still getting some flickering of the status window.  There are two sources of the flickering.  The first is due to the setting of a background brush in the status window class.  This cause the window to be filled with the background color when BeginPaint() is called.  Because there is a gap between the fill and the eventual rendering of the status text, the desktop compositor can catch the front buffer in this intermediate state and thus causing a flicker.  The second is once again small gaps between when text is filled with the background color and the text is rendered.  This gap is smaller and is within the DrawText call itself.  This can only be eliminated through the use of back buffer rendering.

We now maintain a back buffer or all rendering.  We render the entire status window into a back buffer and this back buffer is copied to the front buffer in a single BitBlt() operation.  This eliminates the flicker as the desktop compositor will now only see a fully coherent front buffer.

### DIFF
--- a/win/win32/mhstatus.c
+++ b/win/win32/mhstatus.c
@@ -12,6 +12,52 @@
 
 extern COLORREF nhcolor_to_RGB(int c); /* from mhmap */
 
+typedef struct back_buffer {
+    HWND hWnd;
+    HDC hdc;
+    HBITMAP bm;
+    HBITMAP orig_bm;
+    int width;
+    int height;
+} back_buffer_t;
+
+void back_buffer_free(back_buffer_t * back_buffer)
+{
+    if (back_buffer->bm != NULL) {
+        SelectObject(back_buffer->hdc, back_buffer->orig_bm);
+        DeleteObject(back_buffer->bm);
+        back_buffer->bm = NULL;
+    }
+}
+
+void back_buffer_allocate(back_buffer_t * back_buffer, int width, int height)
+{
+    HDC hdc = GetDC(back_buffer->hWnd);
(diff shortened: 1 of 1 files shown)


alkom:
It may not even require back buffer. The flickering is most likely because of InvalidateRect in MSNH_MSG_UPDATE_STATUS handler that invalidates the entire window with erase background flag -> background is erased and repainted on every turn. You may want to try calling InvalidateRect(hWnd, NULL, FALSE) and then calling FillRect before DrawText in WM_PAINT. Alternatively, you can compare old and new values in MSNH_MSG_UPDATE_STATUS and call InvalidateRect only for areas that have changed (it is more work to calculate offsets though).

barthouse:
Thanks for the suggestions.

I think comparing values and calling InvalidRect only on those field positions would come really close to no flicker since it would significantly reduce how often we do any rendering.  However, we still have the fundamental problem that rendering could still get caught in the middle of a "composition" of the desktop and we occasionally would see the flicker as a result (albeit at a very low rate).  It would also add a bit of complexity and more opportunity to get something wrong.

The back buffer rendering is straight forward and just leverages the existing rendering logic with the addition of a BitBlt at the end.  The storage needed for the back buffer is relatively small.

FYI, I left the InvalidateRect() calls as is and just set hbrBackground to NULL.  Setting the background brush to NULL has the same effect as calling InvalidRect() with no erase.  (I probably should go and change the calls to InvalidateRct to pass FALSE for erase so that other see a consistent intent).

paxed:
Committed.

PR #66, opened 2017-10-09
Attempt to open defaults.nh from HACKDIR if not found for WIN32.

barthouse:


barthouse:
Given that this is only fixing a developer scenario, I'm closing the commit.

PR #67, opened 2017-10-09
Clear selection count when reverting back to permanent inventory view.

barthouse:
We need to clear selection counts after the permanent inventory window was used to select items.


paxed:
Committed.

PR #68, opened 2017-10-15
Changes to player selection dialog.

barthouse:
Significant changes to player selection dialog that attempt to match recently made changes to the Qt and X11 window ports.


paxed:
Committed

PR #69, opened 2017-10-21
WIN32 CONSOLE: Detect when the current console font has glyphs that are too wide

barthouse:
For Win32 console (NetHack.exe), added support to detect when the current console font has glyphs that are too wide and will cause rendering errors in the console.  If detected, we warn the user and change the code page to 437 and the font to Consolas.  At exit, if we had changed the font and code page then we will restore to the original font and code page.

NOTE: the warning is given using raw_print().  Currently, raw_print messages are lost when we transition to displaying game windows.  We need another pull request (#57) to be completed in order for this warning to show up properly.  This will need to be tested.


paxed:
Committed

PR #70, opened 2017-10-22
Add configuration for Qt 5

chasonr:
I called the directory Qt4, but the source code also compiles with Qt 5, with a little help from the preprocessor.

With Qt 5, at least on Fedora 26, the needed packages are Qt5Gui Qt5Widgets Qt5Multimedia, the MOC is moc-qt5 and the source must be compiled with -fPIC. The linker complains if you don't use -fPIC.

This change adjusts sys/unix/hints/linux-qt4 to provide the variables QTGUI and MOC, with an alternate configuration for Qt 5, and adjusts sys/unix/Makefile.src to accept these variables.

### DIFF
--- a/sys/unix/Makefile.src
+++ b/sys/unix/Makefile.src
@@ -273,7 +273,7 @@ WINX11LIB = -lXaw -lXmu -lXext -lXt -lX11
 WINQTLIB = -L$(QTDIR)/lib -lqt
 #
 # libraries for Qt 4
-WINQT4LIB = `pkg-config QtGui --libs`
+WINQT4LIB = `pkg-config $(QTGUI) --libs`
 #
 # libraries for KDE (with Qt)
 WINKDELIB = -lkdecore -lkdeui -lXext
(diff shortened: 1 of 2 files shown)


tux871:
Hello, chasonr. Tell me please, is it possible to compile nethack qt 3.6.0 for windows 7? I have a small experience in compiling programs, I only know how to compile small programs into microsoft visual studio. The latest version of qt in windows is qt 5. I found your repository nethack-3.6.0-patches, it seems like it can be compiled under qt 5 in windows. But what to do next? I found no instructions for windows in the source files of nethack. The Install.Qt file only says that "This code should also work with Qt / Windows, but support for that is not currently official."

chasonr:
The nethack-3.6.0-patches repository has a means of compiling the Qt 4 branch in Windows, against either Qt 4 or 5, using MinGW-w64. That project seems to have given way to MSYS2 (and anyway the current Qt 5 ships with its own copy of MinGW). I have never set it up to work with Visual Studio, but it should be possible.

tux871:
Did not understand which project gave way to msys2? Did not understand your phrase "against either Qt 4 or 5, using MinGW-w64", does this mean that I can compile only the qt 4 branch in windows? I have to use a branch: interfaces, right? What is the easiest way to compile? Without a difference, qt 4 or 5. I have little idea what to do in order to compile, what should I read to learn how to do this?

chasonr:
I'll open [an issue](https://github.com/chasonr/nethack-3.6.0-patches/issues/8) on the nethack-3.6.0-patches repository, since this conversation is really about that version and not the DevTeam's work.

chasonr:
I'm going to close this pull request and redo my fork, now that 3.6.1 is out. I hope to get a full set of configurations for Qt 5 on Linux, Windows and Mac OS X.

PR #76, opened 2017-12-26
Fix memory error by adding byte for null string terminator.

bobbydurrett:
Line 95 in mail.c introduced a bug by allocating one byte less than was needed.
The call to alloc was supposed to get enough bytes to hold a string but it
used strlen() to get the number of bytes. We should have allocated strlen()+1.
This didn't cause an error on Linux on x86-64 but it showed up on the
Raspberry Pi platform.

### DIFF
--- a/src/mail.c
+++ b/src/mail.c
@@ -92,7 +92,7 @@ getmailstatus()
 {
     char *emailbox;
     if ((emailbox = nh_getenv("MAIL")) != 0) {
-        mailbox = (char *) alloc((unsigned) strlen(emailbox));
+        mailbox = (char *) alloc((unsigned) strlen(emailbox) + 1);
         Strcpy(mailbox, emailbox);
     }
     if (!mailbox) {


bobbydurrett:
This is a fix to commit 1c8f312fb6bbe2f8096d9ebdaa63e006aa052b54. I fixed an issue with memory and email in the pull request for that commit but I allocated one byte too few. For some reason this did not cause a problem on the Linux x86-64 bit platform that I tested on but it causes a crash when you use the -D flag on Linux on the Raspberry Pi. I'm not sure how many people use that platform but it probably is good to fix this so we don't have some subtle memory error on other Linux platforms. The problem is that I am allocating one byte less than I need and then overwriting some other byte in memory when I copy a string into the allocated buffer. My previous pull request that this fixes was called "Fix crash with MAIL env variable"

davecohrs (commented):
This has been addressed in a different way as part of commit 859ef823d6.

bobbydurrett:
Great! I will close the pull request.

bobbydurrett:
Dev team resolved another way.

bobbydurrett:
Great. Glad that it got fixed.

Bobby

PR #77, opened 2018-01-07
Refactor getobj and add the ability to handle floor items/features to it

FredrIQ:
This pull request refactors getobj as to allow the selection of objects, or dungeon traps/features, from the floor. This is done by supplying a "," reply when it asks for a letter. Doing so will pick the floor item/feature (if only one choice is valid) or give a list of items + feature, if there are multiple choices. '.' can also be selected, but it is merely an alias (it doesn't display as a valid option or similar), to allow for SLASH'EM muscle memory. If menulist is set to anything other than FULL, getobj prompts will work as they did earlier and prompt individually for each object or feature on the floor.

This code is not based on code from either SLASH'EM or NetHack4, but was created from scratch for 3.6.1. It works by, instead of having a list of object classes as a parameter for getobj, and having getobj special case basically every single case, use a callback for figuring out whether or not a choice is valid. The callback is an int, takes a normal object, NULL (for bare hands check) or &zeroobj (for dungeon feature/trap, if allow_floor is enabled), and returns 2 (valid option, and encouraged by being listed), 1 (possible choice, but not listed among choices or '?'), 0 (invalid).

This allows one to do away with all the special cases that surrounded getobj, meaning it only has the actual code for doing the logic in it.

Screenshots: http://home.fiq.se/nh/getobj.html

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -633,7 +633,7 @@ E int NDECL(opentin);
 E int NDECL(unfaint);
 #endif
 E void NDECL(eatmupdate);
-E boolean FDECL(is_edible, (struct obj *));
+E int FDECL(is_edible, (struct obj *));
 E void NDECL(init_uhunger);
 E int NDECL(Hear_again);
 E void NDECL(reset_eat);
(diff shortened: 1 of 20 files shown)


FredrIQ:
0b638592a487103384faa79aacae9ed2134ee396 makes this PR obsolete.

PR #79, opened 2018-01-24
Add two famous engravings

spixi:
I think they perfectly fit here.


FredrIQ:
The cake is a lie is from the original Portal.

paulwinner:
Shows my age, I thought it was from Alice In Wonderland.

spixi:
> The cake is a lie is from the original Portal.

Sorry, I fixed this.

PR #82, opened 2018-03-13
There should be more stuff about ducks.

trevortomesh:
I added a bit about ducks because there weren't ducks. DUCKS.


trevortomesh:
This was stupid... sorry.

NESHomebrew:
I agree.

NESHomebrew:
About more ducks.

PR #85, opened 2018-04-26
Fix bug in really_close() which can cause standard input to be closed prematurely

barthouse:
Fix bug where user is unable to confirm exit.  If the user chooses to quit when prompted for player name to choose, then they will find that they can not confirm the exit by hitting the return key.  This is due to a bug in really_close() which mistakenly closes the file descriptor zero which is the input stream.  really_close() needs to test lftrack.init before it attempts to access and modify lftrack state.

### DIFF
--- a/src/files.c
+++ b/src/files.c
@@ -664,13 +664,17 @@ int lev, oflag;
 void
 really_close()
 {
-    int fd = lftrack.fd;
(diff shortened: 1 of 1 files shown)


barthouse:
This change was taken and made 3.6.1.

PR #86, opened 2018-04-27
Changed build to default to using latest SDK available on build machine

barthouse:
Modified build configuration to use latest SDK available by default.  This change will eliminate the need for us to hard code an SDK version into our configuration file and will eliminate the need for developers to set the SDK version when they do not have the matching SDK version installed.  Updated the Install.nt file removing the mention of having to set the SDK version.


barthouse:
This commit was taken and made 3.6.1.

PR #87, opened 2018-04-27
Specify both width and height when creating font for width testing.

barthouse:
When we are creating the console font for testing character widths, we were not specifying width.  Because of this, the created font's average width might be larger then what we expect and we might falsely detect that the font was inappropriate for playing Nethack.  Fix provides the width that we are expecting when creating the font.


nhmall:
This was fixed upstream in 
331abfe53096d68bcf5b4e8e1190848a7cab9f4d

PR #90, opened 2018-05-03
fix nonverbose CC on bsd make

lotheac:
Hi, attempting to build 3.6.1 on OpenBSD results in:

```
/tmp/NetHack/src 1 % CONFIG=simple-config GAMEDIR=lib/nethackdir-3.6.1 FILESDIR=/usr/ports/games/nethack/3.6/files make 
touch ../src/config.h-t
[CC] monst.c
[CC] objects.c
[CC] makedefs.c
Using $< in a non-suffix rule context is a GNUmake idiom (Makefile:268)
*** Error 2 in /tmp/NetHack/src (Makefile:643 '../util/makedefs')
```

This was introduced in 1ff4dfee854be09b14c434d49c1f5bf680554e7d. The simplest fix is to use $@ instead (printing "foo.o" instead of "foo.c").


nhmall:
In the current NetHack-3.7 work-in-progress Makefile.src, the problematic section only occurs if the build sets QUIETCC=1. As long as you don't do that on OpenBSD, I don't think there's an issue.

lotheac:
Thanks, you are correct: the current master branch builds fine on OpenBSD.

PR #91, opened 2018-05-06
The following change addresses the recent lag issues with the Win32 console port

barthouse:


barthouse:
NOTE: I posted this to allow dev team folks to start testing while I finalize testing.  Need to understand why we had to use a buffer height of 26.

barthouse:
Creating new pull request to reflect this work is done.

PR #93, opened 2018-05-06
Implemented a fix to the lag problems that are happening with the Win32 console port

barthouse:
The fix implements a console back buffer which significantly reduces the number of calls made to WriteConsoleOutputXXX and eliminates the lag users have been experiencing.


barthouse:

PR #95, opened 2018-05-07
Introduced nhassert() macro to be used for assertions

barthouse:
Each port can decide to implement the macro as they see fit.  In the windows port, the macro is implemented for debug builds only.  When an assertion fails we generate a debug break point, when the debugger is present, and call error() to report the assertion failure.

Changes also include cleanup to mhdlg.c to use new macro and clean up some previously checked in debug code.


barthouse:

PR #96, opened 2018-05-08
General cleanup for nttty.c

barthouse:


barthouse:

PR #97, opened 2018-05-08
Fix for bug H7132 - conflict between extra choice accelerator and menu accelerator

barthouse:
In nethackw, there can be conflicts between menu accelerators and an extra
choice accelerator.  For example, when engraving the using fingers options
conflicts with the unselect all menu accelerator.  The extra choice
accelerator should take precedence.


nhmall:
Merging upstream, thanks.

PR #103, opened 2018-05-24
Changes to build for  MS-DOS

chasonr:
I build the MS-DOS version using FreeDOS in a VirtualBox VM, because DJGPP is painfully slow on DosBox. The changes presented here work around some FreeDOS quirks when running setup.bat.

Also, getreturn_enabled in sys/share/pcsys.c needs to be Win32-only, or the compile will fail.

### DIFF
--- a/sys/msdos/setup.bat
+++ b/sys/msdos/setup.bat
@@ -31,17 +31,13 @@ if exist ..\..\dat\data~1.bas goto long1b
 goto err_long
 :long1a
 echo Changing some long-named distribution file names:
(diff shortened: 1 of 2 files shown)


chasonr:
A further point that my commit  comment did not address: the MS-DOS version uses posixregex.c, which has a non-8.3 name, and setup.bat needs to deal with that one also.

nhmall:
NetHack commit 269363467c2cb46d415b166c8dae4f9de2c7cee8 was meant to deal with the posixregex.c bit

nhmall:
I think all of these were incorporated.

PR #104, opened 2018-05-24
Build with Qt 5 on Linux, Windows and Mac OS X

chasonr:
This pull request provides needed changes to compile against Qt5 on Linux (my distribution is Fedora 28), Mac OS X and Win32.

For Mac, use the Qt 5 distribution from Homebrew.

For Win32, use the official Qt 5 installer. This comes with a MinGW distribution; the Makefile.GCC is tested against that distribution.


paxed:
Merged.

PR #119, opened 2018-07-22
Migrate MH_RACE flags to their own field in the permonst struct

NullCGT:
In monflag.h, there is a comment about possibly moving MH_HUMAN, MH_ELF, MH_DWARF, MH_GNOME, and MH_ORC to their own field, instead of having them mask the corresponding M2 bitflags. This pull request implements a race field. The changes are as follows:

1. There is now an "mrace" entry in the permonst struct in order to hold the monster race bitflags.
2. M2_HUMAN, M2_ELF, M2_DWARF, M2_GNOME, M2_ORC, M2_UNDEAD, M2_DEMON, M2_GIANT, and M2_WERE have been changed to mrace bitflags.
3. All monsters in monst.c have been retagged to support the new system, and artifacts now grant warning based on mrace instead of mflags2. All relevant references to one of the mentioned flags and mflags2 have been updated.

These changes have several distinct advantages:

1. Most importantly, the M2 bitflags now make more sense, since they are no longer a mix of races and monster features, and instead solely correspond to monster features.
2. Adding new races is now much easier, since it just involves adding another race bitflag and changing ROLE_RACEMASK.
3. There is now room for additional M2 bitflags, which were previously all filled.
4. These changes can be potentially expanded in the future to clean up some of the code in mondata.h. For example, an MH_FLAYER tag could be added in order to clean up the definiton for is_mind_flayer(ptr).
5. Having a separate monster race field allows for interesting changes to races that were not possible before without some sort of kludge, such as making all giants hostile to dwarves.


NullCGT:
This should not change gameplay in any way, just the way that races are handled internally.

NullCGT:
Ignore this, going to clean this up.

PR #126, opened 2018-08-18
Elf Iron patch

brian-donovan:


brian-donovan:
Rewrite of the Elf Iron patch by Pasi Kallinen.

PR #128, opened 2018-08-29
fix issue with wiz_identify and char unsigned

bobbydurrett:
I emailed the dev team about a problem I had with identify in wizard mode on the Raspberry Pi. If you press control-I twice in a row it is supposed to permanently identify the objects in your inventory. The first control-I shows the identified inventory but the second control-I does nothing and after you are done if you check the inventory the items you wanted to identify are still unidentified.

This issue is caused by the C compiler on the Raspberry Pi implementing the char type as unsigned. The wiz_identify code that was introduced in 3.6.0 assumes that char is signed. According to this Stackoverflow article there is no guarantee that char is signed in C:

https://stackoverflow.com/questions/2054939/is-char-signed-or-unsigned-by-default

Here are the lines in the 3.6.0 code that assume that char is signed:

line 2083 in invent.c:

        any.a_char = -1;

line 595 in cmd.c:

        if (display_inventory((char *) 0, TRUE) == -1)

On the Raspberry Pi with char being unsigned display_inventory((char *) 0, TRUE) returned 255 instead of -1.

This pull request just casts the output of display_inventory to a signed char and the 255 is converted back to -1 and the wiz_identify works. The dev team may choose to fix this another way. Maybe we could choose a different value besides -1 such as 127 which would be the same with signed or unsigned char types.

Bobby

### DIFF
--- a/include/wintype.h
+++ b/include/wintype.h
@@ -59,6 +59,9 @@ typedef struct mi {
 } menu_item;
 #define MENU_ITEM_P struct mi
 
+/* constant for wiz_identify char return value */
+#define WIZ_IDENT_VAL 127
+
 /* select_menu() "how" argument types */
 /* [MINV_PICKMASK in monst.h assumes these have values of 0, 1, 2] */
 #define PICK_NONE 0 /* user picks nothing (display only) */
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -625,7 +625,7 @@ wiz_identify(VOID_ARGS)
 {
     if (wizard) {
         iflags.override_ID = (int) cmd_from_func(wiz_identify);
-        if ((signed char) display_inventory((char *) 0, TRUE) == -1)
+        if (display_inventory((char *) 0, TRUE) == WIZ_IDENT_VAL)
             identify_pack(0, FALSE);
         iflags.override_ID = 0;
     } else
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -625,7 +625,7 @@ wiz_identify(VOID_ARGS)
 {
     if (wizard) {
         iflags.override_ID = (int) cmd_from_func(wiz_identify);
-        if (display_inventory((char *) 0, TRUE) == -1)
+        if ((signed char) display_inventory((char *) 0, TRUE) == -1)
             identify_pack(0, FALSE);
         iflags.override_ID = 0;
     } else
--- a/src/invent.c
+++ b/src/invent.c
@@ -2580,7 +2580,7 @@ long *out_cnt;
     if (wizard && iflags.override_ID) {
         char prompt[QBUFSZ];
 
-        any.a_char = -1;
+        any.a_char = WIZ_IDENT_VAL;
         /* wiz_identify stuffed the wiz_identify command character (^I)
            into iflags.override_ID for our use as an accelerator */
         Sprintf(prompt, "Debug Identify (%s to permanently identify)",


Rhialto:
Would assigning the value 255 to a signed char not be integer overflow, and hence undefined behaviour?
Based on that, using 127 or some value like that which is valid for both signed and unsigned chars would be the only solution.

```
6.3.1.3 [Signed and unsigned integers]

1   When a value with integer type is converted to another integer type other than _Bool, if
    the value can be represented by the new type, it is unchanged.

2   Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or
    subtracting one more than the maximum value that can be represented in the new type
    until the value is in the range of the new type.60)

Footnote 60) The rules describe arithmetic on the mathematical value, not the value of a given type of expression.

3   Otherwise, the new type is signed and the value cannot be represented in it; either the
    result is implementation-defined or an implementation-defined signal is raised.
```

from C11. That suggests at least that we want to avoid that.

bobbydurrett:
I think what is happening based on the C11 things listed above is that rule 2 is coming into play. On line 2083 in invent.c in 3.6.0 we assign -1 to an unsigned character. any.a_char = -1; The compiler adds 256 which is one more than the maximum number that an unsigned char can represent and you get 255. It probably makes sense to use some number in the range 0 to 127 instead of -1. My solution was just a one line fix that does not affect anything else but changing the number stored and checked against may be the better way to go. Doesn't matter to me how we fix it. It is just helpful to be able to use Wizard mode identify on the Raspberry Pi.

Rhialto:
Eh, no. I added a quote from the (draft) C11 standard (http://port70.net/~nsz/c/c11/n1570.html) which seems to describe this situation (maybe you wrote your reply before you saw that): it is implementation defined, and hence you cannot portably depend on it. Also, integer overflow on a signed value is undefined behavior (see the EXAMPLE in this paragraph):

```
3.4.3 [Terms, definitions, and symbols]

1   undefined behavior
    behavior, upon use of a nonportable or erroneous program construct or of erroneous data,
    for which this International Standard imposes no requirements

2   NOTE Possible undefined behavior ranges from ignoring the situation completely with unpredictable
    results, to behaving during translation or program execution in a documented manner characteristic of the
    environment (with or without the issuance of a diagnostic message), to terminating a translation or
    execution (with the issuance of a diagnostic message).

3   EXAMPLE        An example of undefined behavior is the behavior on integer overflow.
```
but I'm not sure if that only applies in this case; maybe and with grepping I don't find more specific references.

But I would conclude that `(signed char) 255` is at least implementation-defined, and should for that reason be avoided.

bobbydurrett:
I'm not sure about (signed char) 255 not being portable. But it makes sense to use something other than -1 as the value stored in a char.

bobbydurrett:
Interesting discussion about 2's complement here: 

https://stackoverflow.com/questions/12276957/are-there-any-non-twos-complement-implementations-of-c

I guess it is possible that on a Unisys system you might have a representation of -1 that doesn't equal 255. I'm not sure that there is a platform supported by nethack that doesn't use 2's complement.

bobbydurrett:
I changed my pull request to use 127 instead of  -1 as the value passed back to wiz_identify. I also got rid of the cast to signed char. Lastly, instead of hard coding 127 I put it in a #define with a friendlier name to make the code clearer.

nhmall:
Fixed in f1d0636ba081c1f8052eaa87e9ca4ff85ad59ac9. There was a typo reference to 124 in that commit.

PR #131, opened 2018-09-07
Add makefile to build DOS version with a cross-compiler.

jwt27:
This makefile is based on the original one for DOS. I made some modifications so the utilities are built for the host system, and only the game for the target system. This file should be easier to maintain since it doesn't require any workarounds for short file names or the crippled default shell in MSDOS.
Parallel build doesn't work, but I guess that's no big deal. It's already a million times faster than building on a slow DOS machine.


chasonr:
1) What are you using for a host system? You need to document that. Cross-compiling NetHack is tricky, because the intermediate programs (makedefs, dlb, lev_comp, dgn_comp, tilemap) are not made for it; the structures and the options that affect monster and object sets (only MAIL, currently) have to be the same for the host system as the target. I guess mingw32 would work, but I would expect a 64 bit Linux to fail.
2) I might put MSDOS_CROSS in pcconf.h, as it isn't relevant to anything other than DOS, and provide it as -DMSDOS_CROSS in Makefile.CROSS.
3) It looks like Makefile.CROSS is based on the existing Makefile.GCC. Makefile.GCC uses a number of hacks to keep the command lines short and to avoid some FreeDOS bugs. In particular, the final link phase could be done by directly passing the .o files in a single command.

jwt27:
1. I'm building on mingw64. So, the utilities and game executable share some binary structures, that must be the same on both build and target systems? I hadn't considered that, but I see how that may cause trouble. Nevertheless, this makefile does produce a working game executable. I did have to undefine MAIL due to a compilation error, which I didn't look into any further.
2. That's a much better idea. I implemented that just now.
3. I did notice that particular hack, and removed it. See line 450.

sys/msdos/setup.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+#    NetHack 3.6 setup.sh   $NHDT-Date: 1432512792 2015/05/25 00:13:12 $ $NHDT-Branch: master $:$NHDT-Revision: 1.9 $
+#    Copyright (c) NetHack PC Development Team 1990 - 2018
+#    NetHack may be freely redistributed.  See license for details.
+
+echo
+echo "   Copyright (c) NetHack PC Development Team 1990 - 2018"
```
Xaleth, line 8: Wrong copyright name.

PR #136, opened 2018-09-19
Fix spelling and capitalization issues in `dat` and `DEVEL`.

mogigoma:
These changes are all 'obviously wrong' to me. They don't occur in any stylistic or old-style dialect instances.

```
nicname         -> nickname
user-writeable  -> user-writable
legibile        -> legible
parenthese      -> parentheses
Satements       -> Statements
challanges      -> challenges
escae           -> escape
retrive         -> retrieve
flexs           -> flexes
entrace         -> entrance
```

This change was made in a heading and nearby, not every instance was changed in the codebase.

```
emacs -> Emacs
```

This change was made to make the usage consistent with the rest of the codebase, which uses the latter.

```
encumberance -> encumbrance
```

### DIFF
--- a/DEVEL/Developer.txt
+++ b/DEVEL/Developer.txt
@@ -62,7 +62,7 @@ NOTE: The following instructions require perl.  If you do not have perl on
 A. If you have never set up git on this machine before:
    (This assumes you will only be using git for NetHack.  If you are going to
    use it for other projects as well, think before you type.)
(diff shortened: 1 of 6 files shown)


nhmall:
pulled upstream

nhmall:
upstream changes will appear in 4 hours or so

PR #137, opened 2018-09-20
Correct Jon Wätte's name in many places to account for encoding.

mogigoma:
I noticed repeated references to `Jon W{tte` and variations of that in various places where credit was given (e.g., guidebook, copyrights). Grepping for `Jon W` showed me all of those references, in addition to two for `Jon Watte`. This seemed like an search-and-replace gone awry or a locale issue, but nothing in the Git history went back far enough. So I pulled down the source for 3.1.0 and found `Jon Watte` as expected, but also `Jon Wätte` (in LaTeX as `Jon W\"atte`).

This commit restores the name to its previous representation of `Jon Watte` in source code and other text files, and `Jon Wätte` in `Guidebook.tex`.

Note that the name `Janne Salmijärvi` is also represented outside of the ASCII (and Unicode) encoding in several places, but it isn't wacky the way Jon's name was.

(Disclaimer: I've never heard of any of these people, but I'm on a proofreading spree.)

### DIFF
--- a/dat/history
+++ b/dat/history
@@ -54,7 +54,7 @@ Mike Passaretti, and Olaf Seibert, developed NetHack 3.1 for the Amiga.
 Norm Meluch and Kevin Smolkowski, with help from Carl Schelin, Stephen
 Spackman, Steve VanDevender, and Paul Winner, ported NetHack 3.1 to the PC.
 
-Jon W{tte and Hao-yang Wang, with help from Ross Brown, Mike Engber, David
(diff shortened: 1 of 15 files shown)


nhmall:
As I recall at the time, the man himself insisted on Jon W{tte for the ASCII representation of his own name, and Izchak Miller upheld his decision, so this won't be changed in source ASCII references.

mogigoma:
Makes sense. Some are still erroneously `Jon Watte` in the ASCIIs, though. May want to update those to `W{tte`.

PR #138, opened 2018-09-20
Fix other spelling issues.

mogigoma:
Another slew of spelling corrections. These corrections come from going through the `doc` directory, but ignoring the `fixes*` files since those feel less 'timeless'. Propagated fixes for issues found in `doc` to rest of codebase where I felt it was appropriate. Tried to mitigate spacing changes that were caused.

This is the last of the 'obvious' and automate-able spelling fixes. Going to run some grammar checkers next to see if they can detect more subtle issues. Will try not to send lame fixes or ones that seem intentional (e.g., dialog choices).

### DIFF
--- a/doc/Guidebook.mn
+++ b/doc/Guidebook.mn
@@ -54,7 +54,7 @@
 .in -\\n(PYu				\" undo indent past label (etc)
 .sn \\n(pdu				\" tmac.n: inter-paragraph space
 ..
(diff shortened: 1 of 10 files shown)


nhmall:
Pulled upstream. It should arrive here in 4 hours or so.

nhmall:
5a1b8fdf967675d6db054e90a120a0dc70e70a1b

PR #139, opened 2018-09-20
Grammar and punctuation fixes.

mogigoma:


nhmall:
pushed upstream. Upstream change should arrive in 4 hours or so.
doc/tmac.n band was backed out to comply with distribution requirement.

PR #140, opened 2018-09-26
Fix "would flyif you weren't levitating" - missing space

copperwater:


nhmall:
pulled upstream; give it about 4 hours to show up here

PR #168, opened 2018-12-10
Autosearch implementation and game option

ghost:
* Default key binding for toggle: '%'
* 'multi' and 'save_cm' are used to queue a search command on walk.
* Autosearch will cost the player game turns which distinguishes it from intrinsic/extrinsic "Searching".


src/cmd.c
```diff
@@ -3210,6 +3212,8 @@ struct ext_func_tab extcmdlist[] = {
             doattributes, IFBURIED },
     { '@', "autopickup", "toggle the pickup option on/off",
             dotogglepickup, IFBURIED },
+    { '%', "autosearch", "toggle the pickup option on/off",
```
NullCGT: This was meant to be, "toggle the search option on/off," or something similar, right?

ghost: '%' was the hotkey used in testing the autosearch on/off. I'm open to recommendations or alternatives.

The idea is that a blinded character may then enable autosearch. This will cost actions on every step, but allow for visibility of the immediately surrounding squares (as search already does, but it is often too inconvenient to repeat).

Another use case is early game, when the dungeon appears to have no outlet, autosearch can quickly reveal the passage onward without the tedious to 'step-and-search'.

Perhaps some other cases I have not yet discovered ;)

copperwater: I think what NullCGT was getting at was
```suggestion
    { '%', "autosearch", "toggle the autosearch option on/off",
```
It shouldn't read "pickup".

ghost: oh my ;) Fixing

ghost:
Moving PR to NetHack-3.6

PR #174, opened 2019-01-06
Add hallucinatory trap names

copperwater:
This adds many funny, realistic, and nonsensical traps to the game, to be shown when the player is hallucinating. Started as a discussion on IRC, and it was evidently a pretty popular idea since a bunch of devteam members ended up tossing in names, so I thought I would write this up as a patch versus vanilla directly.

Architecturally, the biggest change is merging the what_trap macro and
the "defsyms[trap_to_defsym(ttyp)].explanation" pattern into a single
function "trapname", which returns the name of the trap, handling the
hallucination case. There is also a second parameter used for overriding
hallucination in the occasional cases where the actual trap name should
always be returned.

In addition, the what_trap and random_trap macros are now obsolete and
not used anywhere, so they are removed.


FredrIQ:
Missed opportunity to reference traps from other roguelikes (for example, Zot traps from DCSS).

copperwater:
If you want to provide additional trap names that I don't know about, add them in your comment here.

chsk-uk:
> If you want to provide additional trap names that I don't know about, add them in your comment here.

This is a pretty neat, and I'd like to propose adding (some of) the [Warning Signs for Tomorrow](https://lifeboat.com/ex/warning.signs.for.tomorrow) as additional hallucinatory traps. In particular, these seem like worthwhile additions:

* chaotic system
* macroscale quantum system
* stable strangelet
* nanoparticle trap
* diamondoid surface
* self-replicating device
* memetic hazard

PR #175, opened 2019-02-05
Update Makefile.ami

ethandicks:
Added line continuations for dependency list for include/hack.h


nhmall:
The pull request is a very useful way to do this going forward. I already applied this particular fix manually in response to your email earlier today, otherwise I would have used this pull request and therefore left it credited to you. The fix should be up on github in about 4 hours.

Issue #117, opened 2018-07-09
build script "linux" top-level folder NetHack


ghost:
#!/bin/bash
#if user root
cd sys/unix
sh setup.sh hints/linux
cd ../..
make all
make install
sh /root/nh/install/games/nethack

ghost:
on CentOS 7
yum install byacc-1.9.20130304-3.el7.x86_64
 yum install flex-2.5.37-3.el7.x86_64 -y

ghost:
append for executable to script.sh

 cp -rf /root/nh/install/games/nethack /usr/bin/

marceliwac:
I know this all comes from the `README` and appropriate other files `(DEVEL/..)` etc. but this is genuinely so helpful for people who just want to set it up on linux. I'd consider adding a script to handle this for the linux, optimally with added instructions for getting the `ncurses.h` library.

marceliwac:
I've found the following to work on Ubuntu 18.04.1 LTS

**1. Switch to root:**
```
sudo su root
```

**2. Create a script `install.sh` using the above comments (i.e. using `nano`):**
```
#!/bin/bash
#if user root
cd sys/unix
sh setup.sh hints/linux
cd ../..
make all
make install
sh /root/nh/install/games/nethack
cp -rf /root/nh/install/games/nethack /usr/bin/
```

**3. Install necessary libraries:**
```
sudo apt-get update &&
sudo apt-get install -y build-essential gcc bison flex libncurses5-dev libncursesw5-dev
```

**4. Make the file executable:**
```
sudo chmod 777 install.sh
```

**5. Run the install script as `root`:**
```
./install.sh
```

copperwater:
Since you don't particularly need to install nethack as root and it's more common for Ubuntu to be a single-user install these days, wouldn't it be simpler to use the default single-user build? I guess you would still need superuser access to install the libraries, but not nethack itself.

Here's an install.sh that I think will work (I don't have Ubuntu, but it should work on any linux flavor that has ~/.local/bin in PATH by default):
> #!/bin/bash
> sh sys/unix/setup.sh sys/unix/hints/linux
> make all
> make install
> mkdir -p ~/.local/bin
> ln -s ~/nh/install/games/lib/nethackdir/nethack ~/.local/bin/nethack

Then place this install.sh in nethack's top level folder, `chmod +x install.sh` to make it executable, and run it as non-root.

PR #176, opened 2019-02-12
Linux installation scripts

marceliwac:
Added the installation scripts for linux as discussed in #117 .

### DIFF
--- a/README
+++ b/README
@@ -122,6 +122,28 @@ Please read items (1), (2) and (3) BEFORE doing anything with your new code.
 
                         - - - - - - - - - - -
 
+Installation instructions
+
+Following are the convenience scripts that can be used to install 
+NetHack on Linux.
+
+1. Install necessary packages
+
+	sudo apt-get update &&
+	sudo apt-get install -y \
+	build-essential \
+	gcc \
+	bison \
+	flex \
+	libncurses5-dev \
+	libncursesw5-dev \
+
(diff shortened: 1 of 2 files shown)


marceliwac:
@copperwater Can you review this PR and let me know if you think anything should be changed please? It seems like you've contributed to this repo previously so you might have some insight.
Thanks!

copperwater:
Honestly, based on my experience with the devteam, I rather doubt they would show favoritism to Linux by providing a top-level install script for it (and note, this doesn't work on all linux flavors - it will only work on flavors where the default shell looks in ~/.local/bin, so someone could trip up on it not working if they assume it works on any Linux install). Either that, or the devteam wouldn't want to open up the possibility of Mac users and Windows users and VMS users, etc, all demanding their own top-level one-command installation script.

Now the script could be placed somewhere down in sys/ and have a `cd ..` attached to the beginning, but no place in sys/ really seems appropriate for it, since there's no Linux-specific folder. Maybe try to contact the devteam directly by email about it, but I'm not sure this is something they'd want to add support for.

Rhialto:
Adding to what @copperwater mentions, I see that the script uses apt-get, which is *buntu or Debian specific. Other distributions would have another command, and possibly different names for the packages to install. That should at least be mentioned in the README and briefly in the script, I'd think.

PR #203, opened 2019-07-07
phrasing clarification

G7Nation:
clarify message when a player displaces a monster


nhmall:

Issue #208, opened 2019-08-06
Feature suggestion - autosave (and quit) each N turns.


ostrosablin:
There's several uses for such feature:

* Public NetHack servers, particularly *Robin accounts. NetHack could be played multiplayer by playing in turns and then giving control of character to another person (usually, after playing N turns). But as you often can get immersed in game and forget about turn counter (even with status hiliting) - it would be great, if game could control this by itself.

* Artifically limiting game sessions. NetHack could be an excellent roguelike to play in a free minute, such as coffee break, but like most turn-based games, it suffers from "just one more turn" syndrome. If you play NetHack as a coffee break roguelike, it would be good to know that game would automatically save and quit, say, after 250 turns. Also, doing this from game itself is much safer, as forcibly terminating game with SIGHUP sometimes damage the saves (I tried timer-based approach and it doesn't work well).

I propose to add a new option with a parameter, such as:

```OPTIONS=autosave:500```

Probably, the most logical and simple way to implement this is as a countdown timer. When game is launched, autosave option value becomes a turn countdown until save-and-quit. Then, when turn countdown reaches 0 - game waits for a safe opportunity to save (that is, when game is not mid-turn and saving won't cancel an important prompt) and then automatically does save-and-quit. Such feature won't require any changes to save format and would be very useful.

ostrosablin:
Implemented the idea in Pull Request #213

nhmall:
Moved to [Feature Requests](https://github.com/NetHack/NetHack/wiki/Feature-Requests) page, with a link back to this Issue and its comments.

Closing the issue.

PR #213, opened 2019-08-13
add a new option: autosave

ostrosablin:
This patch adds a new compound option: autosave (as proposed in GitHub
issue #208).

The idea behind patch is to allow artificially limiting game session
duration, by automatically saving-and-quitting after playing for N turns
(where N is argument of the option). This could be useful for public
servers, where *Robin accounts are used as a form of multiplayer and
players play for N turns, and then give control to another person.
Another use case is to use this option to just keep yourself from
playing too long game sessions (e.g. during coffee-break, etc.).

Option is specified like this:

OPTIONS=autosave:500

Which would save-and-quit after playing for 500 turns.

Defaults to -1 (autosave disabled). 0 is invalid value, because game
would then quit right after starting. Any positive value is used as turn
countdown, and when it reaches 0, game is saved and quit automatically.

Signed-off-by: Vitaly Ostrosablin <tmp6154@yandex.ru>

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -576,6 +576,11 @@ typedef unsigned char uchar;
 
 #define USE_ISAAC64 /* Use cross-plattform, bundled RNG */
 
+/* AUTOSAVE enables a new 'autosave' compound option, which allows
+ * to specify max number of turns to play, before doing automatic
+ * save-and-quit, defaults to off. */
+/* #define AUTOSAVE */
+
 /* End of Section 4 */
 
 #ifdef TTY_TILES_ESCCODES
--- a/include/flag.h
+++ b/include/flag.h
@@ -217,6 +217,9 @@ struct instance_flags {
 #ifdef ALTMETA
     boolean altmeta;      /* Alt-c sends ESC c rather than M-c */
 #endif
+#ifdef AUTOSAVE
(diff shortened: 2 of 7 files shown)


tmp6154:
Not really sure, why Travis checks has failed for X11 and QT5. Judging from log, X11 has failed due to missing headers (X11/Xaw/Cardinals.h) and QT5 has failed because of missing tools (bdftopcf), so it seems to be a Travis issue. Anyway, patch doesn't change anything in the windowports, so it shouldn't break them.

ostrosablin:
Now Travis config is fixed and checks are passing.

During my testing, feature also works fine.

Note, that implementation doesn't give strict guarantee that you can't go over specified number of turns (e.g. if you use travel - it could use more turns than specified, but then game will quit after the command ends). This is intentional (to prevent interruption of pending commands), so that option will be safe and won't save the game in inconsistent state and/or with incompletely performed commands.

ostrosablin:
Updated the patch, now it's compiled conditionally, as long as AUTOSAVE define is set. Put the (commented out) define into experimental section of config.h.

This way patch won't cause any changes to the game, as long as it's not compiled in.

ostrosablin:
Updated to make it changeable from the game.

ostrosablin:
Rebased to 3.7 branch

PR #219, opened 2019-09-09
Align add_opvars formats with their parameters

chasonr:
This change fixes all add_opvars calls in lev_comp.y and its Yacc output, lev_yacc.c. Most of the changes consist of changing 'i' formats to 'l'.

A separate commit bears the Clang plugin that I used to find these mismatches. You may or may not want this in your tree.


chasonr:
The switch to Lua renders this change moot. Withdrawing.

PR #222, opened 2019-09-18
Refactor mongets to return the object it creates

copperwater:
Most cases where mongets is used discard the returned value (which used
to be the created object's spe); the only places that do use it are the
series of statements that give various human monsters armor and prevent
them from getting too much armor. These statements included hardcoded
constants representing the base AC of the armor, which would have caused
discrepancies if armor's base AC were ever changed.

With mongets now returning a pointer to the created object, it can just
be passed into ARM_BONUS instead, which covers both the base AC and the
enchantment. (It will also cover erosion, if anyone ever decides that
armor should rarely generate as pre-eroded).

The overall algorithm is not changed by this; human monsters should
receive armor with the same probabilities as before.

The impetus for this was to avoid ugly constructions such as the one
below:

```c
mongets(mtmp, LONG_SWORD);
struct obj* sword = m_carrying(mtmp, LONG_SWORD);
if (sword)
    /* do thing to sword */
```

No such constructions appear in vanilla NetHack, but I did notice that there are a lot of mksobj calls followed by mpickobj, which could probably be replaced with this version of mongets. This would also cut down on some other code duplication (such as ensuring angels' gear is erodeproof and non-cursed).

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -142,7 +142,7 @@ E void FDECL(Lift_covet_and_placebc, (int, const char *, int));
 #define unplacebc_and_covet_placebc() \
             Unplacebc_and_covet_placebc(__FUNCTION__, __LINE__)
 #define lift_covet_and_placebc(x) \
-            Lift_covet_and_placebc(x, __FUNCTION__, __LINE__) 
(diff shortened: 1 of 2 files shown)


paxed:
Committed.

PR #228, opened 2019-10-01
Port the autounlock feature from UnNetHack

copperwater:
This adds a boolean option, autounlock, defaulting to true. When this is
set to TRUE, messages stating that some door or container is locked are
automatically followed by a prompt asking if you would like to unlock
it, if you are carrying an unlocking tool (key, lock pick, or credit
card).

Architecturally, this extends the pick_lock function to take three
additional arguments (door coordinates or a box on the ground you are
autounlocking).

Because this adds a new field to struct flag, this is not a
save-compatible change. I have not adjusted EDITLEVEL or
VERSION_COMPATIBILITY, though.

The code that selects an unlocking tool will always look first for a
skeleton key, then a lock pick, then a credit card. Since curses, rust,
and other attributes don't really have an effect on the viability of the
unlocking device, it didn't seem to warrant making a more complex
function for that.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -1100,7 +1100,7 @@ E boolean FDECL(picking_at, (int, int));
 E void FDECL(breakchestlock, (struct obj *, BOOLEAN_P));
 E void NDECL(reset_pick);
 E void FDECL(maybe_reset_pick, (struct obj *));
-E int FDECL(pick_lock, (struct obj *));
+E int FDECL(pick_lock, (struct obj *, xchar, xchar, struct obj *));
 E int NDECL(doforce);
 E boolean FDECL(boxlock, (struct obj *, struct obj *));
 E boolean FDECL(doorlock, (struct obj *, int, int));
--- a/include/flag.h
+++ b/include/flag.h
@@ -20,6 +20,7 @@ struct flag {
     boolean autodig;    /* MRKR: Automatically dig */
     boolean autoquiver; /* Automatically fill quiver */
     boolean autoopen;   /* open doors by walking into them */
+    boolean autounlock; /* automatically apply unlocking tools */
     boolean beginner;
     boolean biff;      /* enable checking for mail */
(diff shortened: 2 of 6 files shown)


copperwater:
Another feature, not strictly related to this, is removing the "Lock/Unlock it?" prompt when an unlocking tool is applied towards a door (more confirmation doesn't seem necessary; the player already indicated they want to use the tool on the door). Since it's not directly related, I didn't alter that behavior here, but it's something to consider. I'd keep the confirmation prompt in for autounlock, but remove it for manual use.

PFGimenez:
I would argue that suggesting to use a credit card to unlock something is a bit of a spoiler and, as such, should not be included in the autounlock feature. Or maybe propose it only if the player already used his credit card as an unlocking tool earlier.

copperwater:
It doesn't seem worth it to me to specially track whether a player already used a credit card, and also add some inconsistent interface behavior for a spoiled player who hasn't yet used one (think a Tourist who bumps into a locked door in the first room), all for the sake of preserving the minor spoiler of "locks can be opened with a credit card" for a hypothetical unspoiled player.

Not including the card as a suggestion at all would also be annoying and inconsistent to spoiled players. I don't imagine most players would be happy with an autounlock that works with keys and lock picks but makes you have to apply credit cards manually.

bhaak:
There might be some cultural thing going on as well. I don't know how many people nowadays think about using a credit card to crack a door lock. It's likely no longer possible.

With iron safes in UnNetHack I only realized later that modern safes can't be cracked with a stethoscope. You either steal the whole safe to crack it somewhere else or get a drill.

I don't even know if stethoscopes ever were useful or if this was just a movie thing from really old movies.

k21971:
1) Yes, you can use a credit card to open certain types of locked doors. Even to this day.
2) Older safes could be and were 'cracked' using a stethoscope, you could listen for the clicks as you turned the dial.

PR #232, opened 2019-10-15
Curses symset

FredrIQ:
This patch makes the previous "cursesgraphics" symset into a fully fledged symset, as well as renaming it "curses". It also does some minor adjustments to the graphics set to complete a few missing cases where it's likely that the developer wanted more refinements, approximating IBMgraphics as much as possible while retaining the previous limitations in terms of character availability.

It also changes default symset handling under curses, to load the curses symset by default, unless "default" is configured explicitly (which will cause it to use the internal graphics symbols), either in the file, or via the O menu.

This PR also fixes a bug with DECgraphics handling (commit c557998), which should probably be merged even if the rest of this is dismissed.

### DIFF
--- a/dat/symbols
+++ b/dat/symbols
@@ -15,6 +15,11 @@
 # NetHack encodes the request to use the alternate font here by
 # having the high bit set (in hexadecimal, \x80 is combined with
 # a character code between \x60 and \x7f).
+#
+# curses is an approximation of IBMgraphics which relies on DEC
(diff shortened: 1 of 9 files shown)


FredrIQ:
Travis build error appears to be on their side after a quick look-over. Sorry about the confusion.

FredrIQ:
Merged in d0c4d27a.

PR #234, opened 2019-10-17
The Colours of Magic

Vivit-R:
Currently, if you want to change the colors of explosions or zapped rays, e.g., to make lightning a different color than frost, or to make magic missiles a different color than sleep rays, you have to edit global constants arrays defined in `decl.c` and `mapglyph.c`. This branch extends the configurable color palette in `color.h` to cover rays and explosions as well as object materials.


FredrIQ:
Customizeable coloring for things in general is something I've considered implementing by means of text based tilesets so you don't need to recompile the entire game for it. Kinda like what exists now for symbols in dat/symbols.

Perhaps make something like that?

ostrosablin:
Shouldn't poison and acid colors be swapped? I think going logically from dragon colors, yellow is for acid and green is for poison, not the other way around.

Vivit-R:
> Shouldn't poison and acid colors be swapped? I think going logically from dragon colors, yellow is for acid and green is for poison, not the other way around.

 I would think so too, but I kept the beam colorings the same as how they are defined in `decl.x`.

Vivit-R:
@nhcopier -- It was an accident that this branch was based on my sparkle PR. I've reverted the changes from that branch -- struct flags should back to normal.

copperwater:
nhcopier is a bot. You probably want @nhmall?

Also, great name for this pull request.

nhmall:
> > Shouldn't poison and acid colors be swapped? I think going logically from dragon colors, yellow is for acid and green is for poison, not the other way around.
> 
> I would think so too, but I kept the beam colorings the same as how they are defined in `decl.x`.

Doesn't  this  change HI_ZAP to a different color?

Vivit-R:
@nhmall I don't think so? I split the color for magic missile off from HI_ZAP so different colors could be used for the resistance sparkle and for the default magic missile color if desired

PR #235, opened 2019-10-17
Invert symset.fallback to symset.explicit and centralize handling.

FredrIQ:
The invert allows us to default to FALSE, which means we only need
to handle symset.explicit in a single function (read_sym_file).

This allows removing the per-case setting of it, and as a result fixes
an inconsistency where OPTIONS=DECgraphics/IBMgraphics failed to set it
correctly, causing curses to ignore the option.

Tested: result

* no symset or DEC/IBMgraphics; curses
* DECgraphics, no symset: DECgraphics
* default symset: default
* "default2" symset (invalid): curses
* DECgraphics symset: DECgraphics

Trying to load more than one symset (by using DECgraphics and symset:default, for example) will use the last entry. This also happens if the last entry is invalid and a previous one wasn't. I believe this is consistent with tty.


FredrIQ:
Apparently "explicit" is reserved on Windows according to Travis, so I ended up using explicitly instead. Sorry about that.

FredrIQ:
This pull request as a file (single-commit): http://home.fiq.se/symset_fix.patch

PR #236, opened 2019-10-17
Keep wintty.c's maxwin up-to-date.

heiner:
Otherwise the check in https://github.com/NetHack/NetHack/blob/2172c5e7ff40ea073e2749a0f7034db9e6a7f835/win/tty/wintty.c#L1423 doesn't seem to make sense.

An alternative would be to remove both the check and  `maxwin` altogether.

### DIFF
--- a/win/tty/wintty.c
+++ b/win/tty/wintty.c
@@ -1503,6 +1503,7 @@ int type;
         panic("No window slots!");
         return WIN_ERR;
     }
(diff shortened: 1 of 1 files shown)


nhmall:
https://github.com/NetHack/NetHack/commit/b7ca0b4cc6502a92e3cb5485d8e075e1fde227b8

PR #241, opened 2019-11-04
.gitignore: add .vimrc

hornlo:
I prefer a local .vimrc to incorporate the Vim settings suggested in DEVEL/code_style.txt rather than alter my global preferences -- which may not always match these suggestions.

This prevents inadvertently adding my (or others') preferences to the repo.


FredrIQ:
Would also be nice to ignore .dir-locals.el for similar reasons. I have it on gitignore for my own NetHack variant's repo for this reason.

PR #249, opened 2019-12-07
Quiet most of the -Wcomma warnings

MaddTheSane:
Xcode likes to enable `-Wcomma` warnings when upgrading projects. For the most part, just replacing the offending comma with a semicolon quiets the warning.
I'm not even going to touch those that are in a `for` declaration.


nhmall:

PR #250, opened 2019-12-07
ANSI-fying

MaddTheSane:
This moves NetHack away from K&R to ANSI C calling conventions.

This also replaces #10


nhmall:
Similar changes were done for NetHack-3.7 work-in-progress.

PR #251, opened 2019-12-07
Remove the mysterious force

copperwater:
The mysterious force is the single most universally hated feature in
NetHack.

Backsliding through Gehennom has proven to be so boring and tedious for
the player, while adding nothing of value to the game, that nearly every
variant removes it. (The sole exception is NetHack4, which makes very
few gameplay changes.)

To compensate for this, all horizontal teleportation within a level is
blocked when carrying the Amulet, rather than 1/3 of the time.

This makes alignments slightly less differentiated, since there is no
longer an advantage to playing chaotic to make the force less odious.
However, this behavior was already on a shaky basis flavor-wise (Moloch
and Gehennom itself have no bias against any alignment, and plenty of
lawful devils and demon lords reside there).

I would also note that in the absence of the mysterious force, there are
several other proposals and implementations of ways to make post-Amulet
Gehennom more interesting than simply backtracking through completed
levels, ranging from monster spawning behavior to a boss rush of demon
lords to the entire branch beginning to collapse.

### DIFF
--- a/src/do.c
+++ b/src/do.c
@@ -1265,48 +1265,6 @@ boolean at_stairs, falling, portal;
     if (new_ledger <= 0)
         done(ESCAPED); /* in fact < 0 is impossible */
 
-    /* If you have the amulet and are trying to get out of Gehennom,
-     * going up a set of stairs sometimes does some very strange things!
-     * Biased against law and towards chaos.  (The chance to be sent
-     * down multiple levels when attempting to go up are significantly
-     * less than the corresponding comment in older versions indicated
-     * due to overlooking the effect of the call to assign_rnd_lvl().)
-     *
-     * Odds for making it to the next level up, or of being sent down:
-     *  "up"    L      N      C
-     *   +1   75.0   75.0   75.0
-     *    0    6.25   8.33  12.5
-     *   -1   11.46  12.50  12.5
-     *   -2    5.21   4.17   0.0
-     *   -3    2.08   0.0    0.0
-     */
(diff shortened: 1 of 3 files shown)


jeremyn:
To maintain the current balance -- meaning, to give something back to chaotic characters -- what about making balrogs always peaceful to chaotics? This would mirror the situation with lawful characters and Archons, and making balrogs the opposite of Archons would be consistent with Tolkien lore.

copperwater:
Making balrogs always peaceful to chaotics would also mean making them peaceful to all elves, which is not very consistent with Tolkien. Also, balrogs aren't really in the same weight class as an Archon; the biggest threat they usually pose is generating with a bullwhip and disarming the player.

This may be a mostly moot point since vanilla has since nerfed the mysterious force in a way that makes it much more even for all alignments and though I hope they would still consider this pull request, they may not be willing to remove it entirely.

jeremyn:
I agree it's a micro-change that's not worth messing with.

But just to nerd out a bit: in Tolkien, balrogs were powerful "fallen angels" and so the natural opposites of Archons. Also NetHack elves are inconsistent since they are appear Tolkien-ish but are chaotic which Tolkien elves were not. Perhaps their alignment was inspired by Discworld elves which were evil or at least totally amoral. And I agree that balrogs are less powerful than Archons in terms of NetHack stats, but that's balanced by balrogs showing up earlier. You normally only deal with Archons after leaving Gehennom during the run up with the Amulet.

PR #252, opened 2019-12-08
Allow teleportation onto the Vibrating Square

copperwater:
It's a minor annoyance when you forget you can't do this in vanilla and
then get relocated somewhere random on the level. Since it's not a
harmful "trap", just allow the adventurer to teleport directly onto it.


copperwater:
Bouncing the pull request to make Travis run on it again, since the build on this repository was broken when I opened this.

PR #253, opened 2019-12-08
Make mkclass usually unable to select an out-of-difficulty monster

copperwater:
This constitutes the surprisingly small "compromise mkclass fix", in
which mkclass always picks an in-difficulty monster from the target
class, unless there are none eligible, in which case a monster will be
selected from the set of monsters with the next lowest difficulty within
that class. The fix consists only of removing the 50% chance that
mkclass would push forward into out-of-difficulty territory when it
didn't need to.

Specific perennial problems that this fixes:
- The infamous master and arch-liches in the Castle, even when
  lower-level liches are valid targets, and even though master and
  arch-liches are supposed to generate only in Gehennom.
- Large and giant mimics generating in shops on very low dungeon levels,
  spelling pretty certain death for anyone who stumbled into them due to
  their stickiness and heavy attacks.
- The random h in the Gnomish Mines being a mind flayer or master mind
  flayer, which also tends to mean doom for low-level characters who are
  unequipped to handle them.

Basically any place mkclass is called is a possible thing fixed by this,
though most such cases are probably latent (e.g. there are no insects
that would be out of difficulty by the time the player is encountering
the summon insects spell). However, if very high difficulty monsters are
ever added to a class that mkclass is called with, this will continue to
work to prevent them from being selected.

Note that if the player manages to genocide or extinct all the
in-difficulty members of a class, the lowest out-of-difficulty one(s)
will still be chosen.

One of the concerns brought up about mkclass fixes in this vein was that
they would substantially wreck the intended difficulty of certain
levels, particularly the Healer quest's dragons. I haven't fully
reviewed all possible special levels for these types of things, but in
the case of the Healer quest the dragons are the same as before. (The
baby dragons of lower difficulty have 0 frequency and can't be chosen by
mkclass, while all of the adult dragons have the same 20 difficulty and
so get selected uniformly. This fix also avoids the bug in which the
first out-of-difficulty monster, such as gray dragon, would always be
selected. Note that if higher-difficulty dragons were added to the
end of the dragon list, those would *not* appear in the Healer quest
under this system unless the player were already at a very high
experience level to put them in-difficulty.)

Other proposed mkclass fixes, such as the one where if all eligible
monsters from a class are out-of-difficulty then no monster generates or
a random monster not in the desired class generates, would however throw
the balance of difficulty on those levels out of whack.

This does not change mkclass_poly, which is explicitly designed to
ignore difficulty considerations (and is only used for player polyself
or monpolycontrol in wizard mode).


copperwater:
Bouncing the pull request to make Travis run on it again, since the build on this repository was broken when I opened this.

PR #254, opened 2019-12-08
Allow teleportation onto the vibrating square in wizard mode

NullCGT:
The vibrating square is classified as a trap, and as such cannot be teleported onto. Testing the vibrating square in wizard mode can be frustrating, since it cannot be teleported onto directly.


NullCGT:
Closed out, since this duplicates #254.

PR #255, opened 2019-12-10
moc-qt5 is called just moc on Ubuntu

rockola:
moc-qt5 is called just moc on Ubuntu (at least on a system with just qt5 and no previous install of qt4).
This patch sets MOC to moc-qt5 if it is found in PATH (old default behaviour), if not, to moc if it is found in PATH. If neither moc nor moc-qt5 is found, an error is signaled.

This patch does not address the multiple ways in which moc is called (e.g. as $(QTDIR)/bin/moc in Makefile.src and elsewhere). However, this patch is sufficient for building NetHack 3.7 (and 3.6 if backported) on Ubuntu 19.10.


rockola:
Actually, old default behaviour is to set MOC to moc-qt5 unconditionally.

> This patch sets MOC to moc-qt5 if it is found in PATH (old default behaviour)

nhmall:
This shouldn't be an issue with hints/linux.2020 on NetHack-3.7 branch

PR #259, opened 2019-12-14
Add paranoid trap, which prevents accidental triggering of harmful traps

copperwater:
Primarily, this ports UnNetHack's paranoid_trap to the vanilla 3.6+
paranoid options system via xNetHack. When this option is enabled, the
player must type out "yes" before moving onto a known trap that apparently
poses a danger to them. When not enabled, the game behaves as it did
before, and there is no warning for moving onto traps.

If the hero is hallucinating, there is no way to determine what the trap
actually is, so they will always be prompted to confirm.  If the hero is
stunned or confused or has prefixed their movement command with 'm',
they will not be prompted regardless of the trap type (overriding
the hallucination case).

The bulk of the code in this commit is a new function, immune_to_trap,
which is how the game decides whether a given monster or the hero is
safe from a certain type of trap. This helps avoid interface annoyances
in places like levitating over the holes in the Castle (a levitating
hero can never trigger a hole just by moving onto the space, so they're
"immune" to it). The function has three possible results: obviously
immune, obviously not immune, or immune but only by way of items or
intrinsics the hero might not have identified. Only the first case lets
you move onto that trap type freely. Since the game doesn't track your
knowledge of your own intrinsics, this is imperfect; a player with
innate fire resistance (and no other source of it) will still be
prompted before moving onto a fire trap.

I have not currently implemented logic allowing monsters to freely move
onto traps they have seen and are immune to, though this should
theoretically only require changing a couple lines in monmove.c.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -144,7 +144,7 @@ E void FDECL(Lift_covet_and_placebc, (int, const char *, int));
 #define unplacebc_and_covet_placebc() \
             Unplacebc_and_covet_placebc(__FUNCTION__, __LINE__)
 #define lift_covet_and_placebc(x) \
-            Lift_covet_and_placebc(x, __FUNCTION__, __LINE__) 
+            Lift_covet_and_placebc(x, __FUNCTION__, __LINE__)
 #endif
 E void FDECL(set_bc, (int));
 E void FDECL(move_bc, (int, int, XCHAR_P, XCHAR_P, XCHAR_P, XCHAR_P));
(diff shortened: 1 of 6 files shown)


Planet-Patrick:
If you're hallucinating, shouldn't you step onto the trap if you think it's harmless?

copperwater:
If you're hallucinating, the trap will appear as various hallucinatory traps, some of which are quite dangerous! The hero wouldn't see it as harmless.

Planet-Patrick:
@copperwater But it WOULD see it as harmless for some traps though right?

copperwater:
I think it's reasonable to handwave this as when a hallucinating hero sees any trap, they hallucinate it as something that's dangerous to them. I don't see the point in trying to implement "sometimes when hallucinating you should step onto a trap anyway" since that's less fun.

PR #262, opened 2019-12-15
Make the #terrain command work while impaired

copperwater:
This is an interface only command, acting just as an aid to remember
what the player has already seen, and removing its availability based on
whether the character is impaired is interface screw.

Note: This does not reveal true monster or object identities while
hallucinating; the hallucinatory glyphs will remain hallucinatory. (Even
though it shouldn't be possible to see monsters with #terrain anyway.)


pat-rankin:
The #terrain command shows the player what the hero remembers, and if the hero's mental faculties are impaired, it refuses to operate.  That's how it's designed to work.

PR #265, opened 2019-12-19
Consolidation pull request for many minor features

copperwater:
Since there are a lot of these, it seems better to make a single pull request instead of over 20 little ones. The commits in here are more or less independent of one another and self-contained.

All of these are from xNetHack, though some originated in other variants. A lot of them are small enhancements to the interface, mostly via messages.

If I have other small feature commits to submit later on, I plan on adding them to this pull request.


src/objects.c
```diff
@@ -675,7 +675,7 @@ TOOL("leash",               None, 1, 0, 0, 0, 65, 12, 20, LEATHER, HI_LEATHER),
 TOOL("stethoscope",         None, 1, 0, 0, 0, 25,  4, 75, IRON, HI_METAL),
 TOOL("tinning kit",         None, 1, 0, 0, 1, 15,100, 30, IRON, HI_METAL),
 TOOL("tin opener",          None, 1, 0, 0, 0, 35,  4, 30, IRON, HI_METAL),
-TOOL("can of grease",       None, 1, 0, 0, 1, 15, 15, 20, IRON, HI_METAL),
+TOOL("can of grease",       None, 1, 0, 0, 1, 15, 15, 20, PLASTIC, CLR_BRIGHT_BLUE),
```
bhaak, line 678: Well, maybe you're not old enough to have seen metallic cans of grease. :-)

https://images.ehive.com/accounts/5386/objects/images/r5l8pe_a36v_m.jpg

They certainly can rust. Plastic are somewhat anachronistic for NetHack but then again, NetHack is in itself anachronistic.

src/pray.c
```diff
@@ -1527,6 +1527,7 @@ dosacrifice()
                 /* Moloch's high altar */
                 if (u.ualign.record > -99)
                     u.ualign.record = -99;
+                pline("An invisible choir chants in Latin, and you are bathed in darkness...");
```
bhaak, line 1538: Where's the fat lady? :-D

copperwater: That one was implemented as a YAFM, which I was going to make a separate pull request to collect all of them in. Though, maybe this one should just be used, since it already has some (bitcoins for an ale for instance)

src/priest.c
```diff
@@ -580,8 +580,10 @@ register struct monst *priest;
             long pmoney = money_cnt(priest->minvent);
             if (pmoney > 0L) {
                 /* Note: two bits is actually 25 cents.  Hmm. */
-                pline("%s gives you %s for an ale.", Monnam(priest),
-                      (pmoney == 1L) ? "one bit" : "two bits");
+                pline("%s gives you %s %s%s for an ale.", Monnam(priest),
+                      (pmoney == 1L) ? "one" : "two",
+                      Hallucination ? "bitcoin" : "bit",
+                      (pmoney == 1L) ? "" : "s");
```
bhaak, line 586: No. Don't drag that clusterfuck into NetHack.
                                                         
We already have hallucinated currencies, how about:
`const char *bits = (Hallucination) ? currency(pmoney) : (pmoney == 1L) ? "bit" : "bits";`
`pline("%s gives you %s%s for an ale.", Monnam(priest), (pmoney == 1L) ? "one " : "two ", bits);`

src/read.c
```diff
@@ -1268,15 +1268,14 @@ struct obj *sobj; /* scroll, or fake spellbook object for scroll-like spell */
     case SPE_REMOVE_CURSE: {
         register struct obj *obj;
 
-        You_feel(!Hallucination
-                     ? (!confused ? "like someone is helping you."
-                                  : "like you need some help.")
-                     : (!confused ? "in touch with the Universal Oneness."
-                                  : "the power of the Force against you!"));
-
```
bhaak, line 1276: I can understand the wish to change something there. But I'd rather
have a special effect when cursed from which a new message would
logically follow instead  of just having the extended message only appear when
non-cursed.

include/extern.h
```diff
@@ -534,7 +534,7 @@ E void NDECL(update_mlstmv);
 E void NDECL(losedogs);
 E void FDECL(mon_arrive, (struct monst *, BOOLEAN_P));
 E void FDECL(mon_catchup_elapsed_time, (struct monst *, long));
-E void FDECL(keepdogs, (BOOLEAN_P));
+E void FDECL(keepdogs, (BOOLEAN_P, BOOLEAN_P));
```
bhaak, line 537: I like the general idea but not in this form.

As stated in the commit message, there are situations when you
shouldn't be able to notice that monsters followed you.

I also seriously dislike just adding boolean flags to functions (note that the current code sometimes also does that so this is more of a general remark than just criticism of this commit).

One boolean is often okay, but two or more make it unreadable. Either
break it out into different functions or change the booleans to enum
options.

src/pray.c
```diff
@@ -47,6 +47,12 @@ static const char *godvoices[] = {
     "booms out", "thunders", "rings out", "booms",
 };
 
+static const char *hgodvoices[] = {
+    "belches", "ring-a-lings out", "squeaks", "squeals", "screeches", "drones",
+    "buzzes", "pipes", "tinkles", "rumbulates", "moos", "brays", "barks",
+    "screams", "howls", "ululates", "quacks", "meows",
+};
```
bhaak, line 54: I'd argue that the voice of a god will overwrite your hallucination.
But also that I don't like getting stuff too silly.

src/read.c
copperwater: I'm totally OK with adding a new cursed effect to this scroll rather than just modifying its message. I don't think I ever came up with one, though, which is why it's not in this pull request.

include/extern.h
copperwater: When the booleans are totally independent, the only real options for reducing the number of arguments are to make a clunky enum that enumerates the different cases, or use bitflags, the added complication of which may be worth it for 3+ independent booleans but probably not 2.

src/zap.c
```diff
@@ -4626,7 +4626,7 @@ short exploding_wand_typ;
         case ZT_COLD:
             new_doormask = D_NODOOR;
             see_txt = "The door freezes and shatters!";
-            sense_txt = "feel cold.";
+            hear_txt = "a deep cracking sound.";
```
repeat83: can we broke door after that (cold damage) because use cold? If answer is "Yes" then how we understand that cracking sound is craking of cold from this text?

src/pray.c
copperwater: Well, we have hallu_gods already...

copperwater:
All these were incorporated except:

- messages for monsters following: @bhaak, since you agree with the general idea, we should try to hash out what would be better here. One question is whether you should always be told when a monster pursues you via a non-stairs method even if you can't see it, because that produces the "gets a grip on you" message. I still think two boolean arguments for a function is okay whereas three would be pushing into making it a bitmask territory.
- god voice funny verbs when hallucinating: I don't really agree with rejecting it on tonal grounds, but I understand the argument.
- removing the "helping you" message from cursed remove curse: Fine, this leaves it open for a possible new cursed remove curse effect in the future, and in the meantime the interpretation of "it removes the curse from itself" still applies.
- grease color and material: Okay, leave it as metal, but can it be a bright blue can of metal? Helps it be more recognizable and not look like a key or pick-axe.
- additional cases for newcham polymorph: Not sure why this one didn't get picked up, I don't see any comments on it at all.

nhmall:
This pull request has all been committed to NetHack-3.7, except for the following bands:
https://gist.github.com/nhmall/17646bf27be669dd151b82db9dd0e5f5

copperwater:
Closing this old pull request.

PR #269, opened 2019-12-22
Peaceful displacement: move peaceful monsters aside just like pets

copperwater:
This is a very popular feature which has been implemented in several variants so far. It allows you to displace peaceful monsters out of the way, following the code which already exists for pets.

Specifically, this is xNetHack's implementation; there are a few other changes in xNetHack attached to the base implementation that I think work well with it.
* Refactoring a bunch of identical code blocks to occur only in one place (didnt_move).
* Some monsters can't be displaced at all: priests, shopkeepers, the Oracle, and your quest leader.
* Peaceful and tame monsters alike will refuse to switch places with you if you are on a trap or otherwise unsafe terrain for it (such as liquid). This indirectly fixes a source of extreme aggravation where a levitating or water walking hero accidentally displaces their pet onto water, instantly drowning it. It also prevents you from trivially killing any peaceful by the same method.
* Monsters can't be displaced out of a trap they are trapped in, meaning that the logic for untaming your pet by displacing it out of a trap no longer exists.
* Sleeping monsters count as immobilized. Additionally, immobilized monsters can't be displaced at all; there used to be a 1/6 chance of displacing them anyway. The 1/6 chance for sessile monsters is retained.

Other notes:
* I noticed when debugging that attempting to move onto a tame, disguised, mimic's space displaces and decloaks it with no message at all; this seemed weird but that appears to be the existing behavior.

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -128,7 +128,7 @@
  * definition here is convenient.
  */
 #define is_safepet(mon)                                                   \
-    (mon && mon->mtame && canspotmon(mon) && flags.safe_dog && !Confusion \
+    (mon && (mon->mtame || mon->mpeaceful) && canspotmon(mon) && flags.safe_dog && !Confusion \
      && !Hallucination && !Stunned)
 
 /*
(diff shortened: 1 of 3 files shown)


Xaleth:
What if you want to fight the peaceful monster? :P

ostrosablin:
Then you can explicitly fight it with F command, rather than using movement keys. Actually, just like it's now with pets. You probably wouldn't want to fight a pet, but if you really want to fight it (and not displace) - you can.

paxed:
Merged.

PR #276, opened 2020-01-04
Use recent GCC for DOS cross-builds on Linux

dimkr:


nhmall:
The NetHack-3.7 branch cross-compile for msdos supports gcc 10 now.

PR #280, opened 2020-01-12
Display weights of objects in inventory

copperwater:
The bulk of this pull request is the commit adding the 'invweight' option, which allows the player to see the weight of any object they are carrying. See its commit message for full details.

I also put in the complementary feature in which showing your inventory will display your total carried weight compared to your current carrying capacity, and your slots used/total.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -2410,8 +2410,13 @@ xprname(struct obj *obj,
     boolean use_invlet = (flags.invlet_constant
                           && let != CONTAINED_SYM && let != HANDS_SYM);
     long savequan = 0;
+    long save_owt = 0;
 
     if (quan && obj) {
+        if (iflags.invweight) {
+            save_owt = obj->owt;
+            obj->owt = obj->owt * quan / obj->quan;
+        }
         savequan = obj->quan;
         obj->quan = quan;
     }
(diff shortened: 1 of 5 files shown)


src/invent.c
```diff
@@ -2686,6 +2695,14 @@ long *out_cnt;
                  xtra_choice, MENU_ITEMFLAGS_NONE);
         gotsomething = TRUE;
     }
+
+    /* Show weight total and item limit. */
+    char invheading[QBUFSZ];
+    int wcap = weight_cap();
+    Sprintf(invheading, "Inventory: %d/%d weight (%d/52 slots)",
+            inv_weight() + wcap, wcap, inv_cnt(TRUE));
```
nicklewis: It looks like including gold here will make this display `(53/52 slots)` for a full inventory. Since `$` is a dedicated inventory slot, is it better to just omit it (rather than increase to 53 slots)?

copperwater: I agree, this would be better off with inv_cnt(FALSE) since gold doesn't occupy a letter slot.

copperwater:
Pushed an update to this that makes it compatible with the new optlist system (basically taking the xNetHack implementation).

PR #281, opened 2020-01-12
Add the Grudge Patch: monsters with natural enmity will fight each other

copperwater:
This ultimately derives from Nephi's Grudge Patch back in the 3.4.3 era,
plus some updates from SpliceHack and xNetHack. Its primary purpose is
to make the dungeon a richer, more interesting place by providing more
opportunities for natural monster-to-monster interactions. All of the
code necessary to support monster battling was already in place; this
merely adds more monster pairs to mm_aggression.

Monsters that will fight each other are:
- Quest guardians and hostile non-guardians
- Angels and demons
- Elves and orcs
- Hobbits and Nazguls

Monsters with a one-sided grudge (where the defending monster may
counterattack, but won't use its own turns to retaliate aggressively)
are:
- Baby purple worms against shriekers (purple worms already had this)
- Ravens against floating eyes
- Spiders against insects (x and a)
- Bats against flying insects
- Cats against rats
- Woodchucks against the Oracle

Pets will never attack other pets, even if they would normally be
hostile to each other.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -1433,6 +1433,7 @@ E int FDECL(curr_mon_load, (struct monst *));
 E int FDECL(max_mon_load, (struct monst *));
 E int FDECL(can_carry, (struct monst *, struct obj *));
 E int FDECL(mfndpos, (struct monst *, coord *, long *, long));
+E long FDECL(mm_aggression, (struct monst *, struct monst *));
 E boolean FDECL(monnear, (struct monst *, int, int));
 E void NDECL(dmonsfree);
 E void FDECL(elemental_clog, (struct monst *));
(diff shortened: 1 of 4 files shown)


NullCGT:
If spiders grudge against the 'x' class, then they would attack xans, which have an high enough AC and speed that they would generally never be killed by a spider. This seems like a potential issue.

copperwater:
Yeah, since x is only grid bug and xan and you won't generally see grid bugs and spiders around at the same time, that grudge might be ineffective in practice. Same deal with the bat one; the only actual flying insect is the killer bee.

hamzaseriat (commented):
İys very good

ostrosablin:
Great patch, would be great to have it merged, since most variants do this.

PR #283, opened 2020-01-15
Add more YAFMs to the game

copperwater:
YAFM - Yet Another Funny Message. This is a consolidation of ones added to xNetHack, though as described in the commit messages some were created in other variants.

Most of these happen when you are hallucinating.

### DIFF
--- a/src/pray.c
+++ b/src/pray.c
@@ -1572,8 +1572,10 @@ dosacrifice(void)
                 u.uevent.ascended = 1;
                 adjalign(10);
                 /* you might be able to see the invisible choir! */
-                pline("%s choir sings, and you are bathed in radiance...",
-                      (See_invisible ? "A visible" : "An invisible"));
+                pline("%s sings, and you are bathed in radiance...",
+                      Hallucination ? "The fat lady"
+                          : (See_invisible && !Blind ? "A visible choir"
+                                                     : "An invisible choir"));
                 godvoice(altaralign, "Mortal, thou hast done well!");
                 display_nhwindow(WIN_MESSAGE, FALSE);
                 verbalize(
--- a/src/pray.c
+++ b/src/pray.c
@@ -1571,8 +1571,9 @@ dosacrifice(void)
             } else { /* super big win */
                 u.uevent.ascended = 1;
                 adjalign(10);
-                pline(
-               "An invisible choir sings, and you are bathed in radiance...");
+                /* you might be able to see the invisible choir! */
+                pline("%s choir sings, and you are bathed in radiance...",
+                      (See_invisible ? "A visible" : "An invisible"));
                 godvoice(altaralign, "Mortal, thou hast done well!");
                 display_nhwindow(WIN_MESSAGE, FALSE);
                 verbalize(
--- a/src/timeout.c
+++ b/src/timeout.c
@@ -747,7 +747,10 @@ nh_timeout(void)
                      * to this number must be thoroughly play tested.
                      */
                     if ((inv_weight() > -500)) {
-                        You("make a lot of noise!");
+                        if (Hallucination)
+                            You("forget the rules to Charades!");
+                        else
+                            You("make a lot of noise!");
                         wake_nearby();
                     }
                 }
@@ -1145,7 +1148,10 @@ lantern_message(struct obj* obj)
             pline("Batteries have not been invented yet.");
         break;
     case OBJ_FLOOR:
-        You_see("a lantern getting dim.");
+        if (Hallucination)
+            You("witness the force of entropy at work.");
+        else
+            You_see("a lantern getting dim.");
         break;
     case OBJ_MINVENT:
         pline("%s lantern is getting dim.", s_suffix(Monnam(obj->ocarry)));
--- a/src/allmain.c
+++ b/src/allmain.c
@@ -307,6 +307,8 @@ moveloop_core(void)
                         warnreveal();
                     mkot_trap_warn();
                     dosounds();
+                    if (Hallucination && !rn2(10000))
+                        You("vaguely feel a vague sense of vagueness.");
                     do_storms();
                     gethungry();
                     age_spells();
--- a/src/apply.c
+++ b/src/apply.c
@@ -3475,6 +3475,15 @@ do_break_wand(struct obj *obj)
     freeinv(obj);       /* hide it from destroy_item instead... */
     setnotworn(obj);    /* so we need to do this ourselves */
 
+    /* If you know the wand you're breaking is a wand of nothing,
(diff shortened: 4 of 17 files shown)


src/pray.c
```diff
@@ -1554,8 +1554,11 @@ dosacrifice()
             } else { /* super big win */
                 adjalign(10);
                 u.uachieve.ascended = 1;
-                pline(
-               "An invisible choir sings, and you are bathed in radiance...");
+                /* you might be able to see the invisible choir! */
+                pline("%s sings, and you are bathed in radiance...",
+                      Hallucination ? "The fat lady"
+                          : (See_invisible && !Blind ? "A visible choir"
```
Xaleth, line 1577: Will it make more sense for it to just say "A choir" instead of having the extra "visible" in there?

src/pray.c
```diff
@@ -1554,8 +1554,11 @@ dosacrifice()
             } else { /* super big win */
                 adjalign(10);
                 u.uachieve.ascended = 1;
-                pline(
-               "An invisible choir sings, and you are bathed in radiance...");
+                /* you might be able to see the invisible choir! */
+                pline("%s sings, and you are bathed in radiance...",
+                      Hallucination ? "The fat lady"
```
jeremyn, line 1576: "The fat lady sings" is arguably a form of fat shaming. Whether you personally agree or not, maybe it's not something to subject a player to in their moment of triumph.

Maybe a reference to psychedelic rock instead? Perhaps a low-key Grateful Dead reference?

copperwater: "It's not over till the fat lady sings" is a proverb, and meshes well with the usual message involving singing and it taking place at the end of the game. I don't consider this to be fat shaming or fatphobic because it's not directed at anyone, isn't casting shame on anyone, and isn't normalizing non-fat people.

src/pray.c
copperwater: I figured it would be less noticeable this way, and a subtle YAFM is a good YAFM in my book.

src/pray.c
jeremyn: I totally get that it's a proverb and think that you are operating with good intentions here, but I'm not really trying to argue that it is or isn't offensive, specifically. What I am arguing is that there are a lot of people who do think it's offensive, which you can confirm with some web searches, and that that position is reasonable, and that it would be irritating to such a person to subject them to what they consider to be an offensive joke right when they should be celebrating.

And like I said there are other jokes to make. It wouldn't even be the first Grateful Dead reference, see https://github.com/NetHack/NetHack/blob/NetHack-3.7/dat/epitaph.txt#L211.

copperwater: Sorry, no interest in changing it. Getting offended by a reference to a proverbial (not real) fat lady is an unreasonable response and might indicate that the reader needs to grow a thicker skin.

If one applies this logic elsewhere, one finds oneself unable to describe anyone as fat, in any fiction, ever, because someone somewhere might be too sensitive to read something with the barest hint of an implication that "fat" can be used as a distinguishing adjective on a person.

ostrosablin: I agree with @copperwater . Pretty much anyone could be offended with pretty much anything, even the most innocent things. Despite the world becomes a strange place, where people are forced to accept as norm to say "sorry" for something that never had no intention to insult anyone, we should try to maintain some sanity for sake of mankind and it's future.

I mean, some feminists are offended with the existence of men itself and they find males offending just because they exist at all. You wouldn't commit suicide just because your existence _might_ offend some feminists, would you? Some people think that "whitelist" and "blacklist" and "master-slave" (in technical context) supports white supremacy and slavery, despite (in technical speak) they're never meant to convey anything like that. And instead of solving real problems (slavery is still a thing, but of course, fighting against the words like "blacklist" is much easier than actually freeing enslaved people), childfree people are offended with the existense of children, despite they also were kids some time ago (what a hypocrisy!).

If we continue down the road of banning words just because they might be offending to somebody - we would eventually run out of vocabulary one day.

jeremyn: This doesn't need to be a big deal. No one is demanding an apology. This is a new change, not some 30 year old piece of NetHack history. The "joke" itself isn't that funny. Reasonable people can and do disagree on whether "the fat lady sings" is offensive/annoying/whatever. We don't need to try out edgy humor (edgy to other people, not necessarily to you) on people as they're crossing the finish line of a NetHack marathon.

We can come up with something else. I propose something like "You hear Jimi Hendrix playing, and are bathed in psychedelic stage lights...." I'm sure there are many, many other alternatives.

Xaleth: @jeremyn whatever you're smoking, I want it so I can forget ever reading this

copperwater:
Updated this PR to current and since I was rebasing and force pushing anyway, merged the fixup commit in where it belonged.

src/pray.c
Vivit-R: I'd say that "a choir sings" would be subtler than "a visible choir sings", since the latter is obvious wordplay (of a kind that I associate with _hallucinatory_ YAFMs) on the idiomatic phrase "invisible choir". Whether or not you're actually going to _notice_ depends on how many times you've ascended. If you've ascended dozens of times already, you're just skimming the victory text and are more likely to notice the absence of the word altogether than you are to notice the absence of two letters at the beginning of the word. But if this is your _first_ ascension, the phrase "visible choir" will stick out like a sore thumb, and you won't even know that it normally says "invisible"—especially since on your first ascension, you are likely to have the see invisible intrinsic anyway. In fact, a lot of players are going to have see invisible on _most_ of their ascensions, which means they will see the choir more often than they won't.

More nitpickily, calling something visible because you have see invisible is inconsistent with the way the rest of the game treats seen invisible monsters. Having see invisible doesn't make invisible monsters visible, it just lets you see them _even though_ they are invisible.

Planet-Patrick: Perhaps "transparent" rather than "visible"?

src/timeout.c
```diff
@@ -1145,7 +1148,10 @@ lantern_message(struct obj* obj)
             pline("Batteries have not been invented yet.");
         break;
     case OBJ_FLOOR:
-        You_see("a lantern getting dim.");
+        if (Hallucination)
+            You("witness the force of entropy at work.");
```
Planet-Patrick, line 1152: Entropy isn't really a force or work; would be more correct just to say that "you witness the entropy increase", though none of these are particularly lantern-related. Perhaps the player can muse that we all must die eventually due to entropy?

Vivit-R: That just makes it better as a hallucination message, doesn't it? Hallucination messages _shouldn't_ make too much sense.

PR #284, opened 2020-01-16
Consolidation pull request for more minor features

copperwater:
As with #265, the commits in here are mostly self-contained and independent of each other, with a few exceptions (small refactors needed to make the following commit work, like 07f1e54). I'm making a new pull request because it seems like #265 has already been reviewed and partially incorporated.

As before, the implementations of all of these are from xNetHack, though many are from other variants such as SpliceHack and SliceHack.


src/zap.c
```diff
@@ -5320,6 +5320,13 @@ makewish()
         buf[0] = '\0'; /* for EDIT_GETLIN */
         goto retry;
     }
+    if (buf[0] == '\0') {
+        if (yn("Really forfeit this wish?") == 'y') {
+            Strcpy(buf, "nothing");
+        }
+        else
+            goto retry;
```
Xaleth, line 5427: What if you truly want to wish for a random object? Can you just type "random" "random item" "anything"

copperwater: There is currently no way to wish for a random object. Not that it would be hard to add a special case for the string "random", but I really don't expect anyone to actually want a random object when they use a wish.

paulwinner:
We are pretty well known for anticipating what players will never actually
do!

PR #286, opened 2020-01-18
Refactor and simplify the rndmonst function

copperwater:
This changes rndmonst() to use a weighted reservoir sampling algorithm, rather than a unwieldy pre-computed cache of eligible monsters that needs to be updated every time the hero goes up or down a floor, gains or loses a level, or extincts or genocides a monster. It removes the rndmonst_cache global (and thus is not save compatible) and the reset_rndmonst function.

This algorithm still requires only a linear pass over the list of monsters (the same as what rndmonst() required even when not updating the cache). 

Many of NetHack's loops could benefit from being rewritten using [reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling), but rndmonst was one of the biggest candidates due to being able to get rid of the cache.

This is a code-only change and does not change the distribution of monsters selected by rndmonst() at all.

c1c90bb is the same commit in the consolidated pull request.

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -979,12 +979,6 @@ struct instance_globals {
     /* lock.c */
     struct xlock_s xlock;
 
-    /* makemon.c */
-    struct {
-        int choice_count;
-        char mchoices[SPECIAL_PM]; /* value range is 0..127 */
-    } rndmonst_state;
-
     /* mhitm.c */
     boolean vis;
     boolean far_noise;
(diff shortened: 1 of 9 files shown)


Xaleth:
What‘s the end user benefit? Less RAM usage?

copperwater:
It's a refactor. It doesn't have an end user benefit, because there aren't any changes to the gameplay or the interface. Players won't see a difference - the main reason for doing this is to clean up the code, simplify it, and make it less brittle.

I suppose it does use a few hundred bytes less RAM by not needing the cache anymore, and it saves processor cycles by not needing to update the cache, but those are insignificant.

ostrosablin:
Simpler code is definitely better. From programmer perspective (either for development or code diving) I definitely find newer implementation more readable and straightforward.

Because many end users do code diving - it could be definitely considered an end user benefit. And tangled logic tends to cause nasty bugs, so you could say end users would benefit by getting less bugged game in future releases.

Performance/memory use advantage could be insignificant for modern platforms (it's a really minor improvement, given modern hardware power), but still could make a noticeable difference when compiled for outdated platforms, e.g. DOS, which are still supported.

PR #287, opened 2020-01-20
Fix typo

Xaleth:


nhmall:
For purposes of discussion, the original text states:
"Few people achieve this; most never do.  Those who have go down in history as heroes among heroes - and then they find ways of making the game even harder." 

That proposed "fix" leaves a seemingly incomplete sentence behind:
"Few people achieve this; most never do.  Those who have **gone** down in history as heroes among heroes - and then they find ways of making the game even harder." 

A possible clarification of  the original intent of the sentence, might be best achieved with the insertion of the words "done so" after the word "have" as follows:
"Few people achieve this; most never do.  _Those who have done so go down in history as heroes among heroes - and then they find ways of making the game even harder_."

I think it was meant to describe what happens to "those who have [done so]"; they "go down in history...", not "those who have gone down in history..."

Xaleth:
You betcha.

PR #289, opened 2020-01-24
Added guard for definition of WIN_STATUS for Linux and GCC 10

tachoknight:
Compiling with GCC 10 causes a duplicate definition error during linking with the `WIN_STATUS` definition in `decl.c`.

### DIFF
--- a/src/windows.c
+++ b/src/windows.c
@@ -860,7 +860,13 @@ const char *status_fieldnm[MAXBLSTATS];
 const char *status_fieldfmt[MAXBLSTATS];
 char *status_vals[MAXBLSTATS];
 boolean status_activefields[MAXBLSTATS];
+#if __linux__
+# if __GNUC__ < 10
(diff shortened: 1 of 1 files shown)


jer-gentoo:
Neither GCC 10 nor Linux have anything to do with this. As laid out in https://github.com/NetHack/NetHack/pull/291, the problem is with the `-fno-common` flag which has existed in many versions of GCC prior to 10 but will be newly enabled by default in that version of the compiler.

tachoknight:
I knew about the `-fno-common` flag, but given one of NetHack's claims-to-fame is that it compiles and runs everywhere, I wasn't sure what unintended consequences simply adding the flag would have, so I opted for a more conservative approach. The Dev Team is free to do what they will regarding this.

nhmall:
This should be resolved in https://github.com/NetHack/NetHack/commit/61d0d21f472e813b3e9872d294b5e33dd8999b93

PR #290, opened 2020-01-26
Extensions to the MS-DOS VESA mode

chasonr:
The first four commits, through "More bits for linear frame buffer," are incremental changes, adding working support for 8 bit color modes and a linear frame buffer.

Subsequent changes overhaul the VESA BIOS support. Tile sets are no longer limited to 16x16 in 8 colors; any tile set in BMP or GIF format is supported.


paxed:
Merged

PR #291, opened 2020-01-27
Fix compiling with -fno-common

jer-gentoo:
GCC 10 will enable -fno-common by default[0], which causes the linker to
fail like this [1]:

```
ld: windows.o:(.bss+0x0): multiple definition of `WIN_STATUS';
decl.o:(.data+0x40): first defined here
collect2: error: ld returned 1 exit status
```

So considering that include/decl.h has:

```
E NEARDATA winid WIN_STATUS;
```
which is correct, and that just one of the .c files should use `extern`, and
that src/decl.c aligns nicely with include/decl.h, keep that definition
and remove those invalid defnitions from other .c files.

Contrast this with [2] https://github.com/NetHack/NetHack/pull/289 which
tries to second-guess the problem by checking for GCC 10 / Linux, both
of which are incorrect assumptions.

[0] https://gcc.gnu.org/gcc-10/porting_to.html#common
[1] https://bugs.gentoo.org/706320
[2] https://github.com/NetHack/NetHack/pull/289

### DIFF
--- a/src/windows.c
+++ b/src/windows.c
@@ -860,7 +860,6 @@ const char *status_fieldnm[MAXBLSTATS];
 const char *status_fieldfmt[MAXBLSTATS];
 char *status_vals[MAXBLSTATS];
 boolean status_activefields[MAXBLSTATS];
-NEARDATA winid WIN_STATUS;
 
 void
 genl_status_init()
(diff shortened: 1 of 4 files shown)


nhmall:
This should be resolved in https://github.com/NetHack/NetHack/commit/61d0d21f472e813b3e9872d294b5e33dd8999b93

PR #294, opened 2020-02-01
Update description of OPTIONS=video

chasonr:
A recent change to the Guidebook says that video_width and video_height work with video:vga. That setting in fact forces the 640x480x16 mode, in which video_width and video_height do not operate. The desired setting is video:vesa. Update the description of OPTIONS=video, video_width and video_height to reflect this.

Also, fix some syntax errors in Guidebook.tex encountered while trying to compile it.

Attempting to update Guidebook.txt, using the Makefile, produced many more differences than expected. I am leaving that file unchanged.

### DIFF
--- a/doc/Guidebook.mn
+++ b/doc/Guidebook.mn
@@ -3987,14 +3987,19 @@ if needed.
 Cannot be set with the \(oqO\(cq command.
 .lp video
 Set the video mode used (PC NetHack only).
-Values are \(lqautodetect\(rq, \(lqdefault\(rq, or \(lqvga\(rq.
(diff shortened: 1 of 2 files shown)


nhmall:
Re: "_Attempting to update Guidebook.txt, using the Makefile, produced many more differences than expected. I am leaving that file unchanged._"

There's some automation run to generate an updated Guidebook.txt when the Guidebook.mn has changed, so no need to update Guidebook.txt in pull requests.

Thanks for going through the tedium of updating both Guidebook.mn and Guidebook.tex.

nhmall (approved):
Will merge

PR #300, opened 2020-02-12
NetHack  htmldump

NHTangles:
This is my current work on HTML dump logs, for your consideration.  This code is live on the hardfought servers. (sample: https://au.hardfought.org/userdata/T/Tangles/nethack/dumplog/1580642779.nh.html)
Please let me know if I can do anything to help you get it merged.
Thanks
Tangles.


NHTangles:
This feature adds a lot of extra code to windows.c that doesn't really belong there.  I was considering moving all the dump code to a new file (dump.c?), but this would be difficult to maintain separately if it does not get merged (since it would also move a lot of existing code).  I'd definitely be prepared to do that work and PR it if this one gets merged.

PR #308, opened 2020-02-29
Lua enhancements

copperwater:
This adds several enhancements to the special level parsing code:

*  Lua scripts can now call d() to get dice results, or percent() to get a percent threshold, avoiding the necessity of using lots of math.random() calls.
*   Special level functions such as terrain() or replace_terrain() cannot overwrite stairs or ladders.
*    Implement selection.gradient(), re-enabling the selection gradient functionality present in 3.6.


copperwater:
New additions as of today:
* When a lua error is encountered, print it in string form rather than error code
* If not specifying a lit state in des.terrain or des.replace_terrain, the preexisting lit state will not be changed.

copperwater:
Individual commits from this PR have been merged, others have not. The ones remaining unmerged are:
3817edc
b620ce7
fc9db0e
d3ef1ba

Closing this PR now because the others aren't particularly related to each other, will send them individually instead.

PR #324, opened 2020-04-15
Allow specifying a montype on figurines in lua files

copperwater:
This was already allowed for the other montype-compatible objects like
statues, corpses, eggs and tins. I don't see a reason why figurines
shouldn't be part of this group; perhaps it was an oversight.


paxed:
Merged

PR #328, opened 2020-04-16
Fix rnd(0) in create_gas_cloud, and blasts of fire that hit bodies of water create clouds of steam.

NullCGT:
This is a combination bug and feature pull request:

- If create_gas_cloud() is called with a damage of zero, the game will throw a panic for calling rnd(0). This pull request fixes this.
- If a bolt of fire hits a fountain or body of water, it produces a small area of clouds that fades after a few turns.

I originally wrote this feature for SpliceHack, but upon further reflection I think that it fits in well to vanilla. The messages "you hear hissing gas" and "steam billows from the fountain" imply the creation of steam, which can be easily represented using create_gas_cloud. I feel that giving the steam a visual representation increases the sense of interacting with the environment.

Fountains produce a larger area of clouds, since the message "steam billows from the fountain" seems to imply a greater amount of steam than "you hear hissing gas."

At present, this has no gameplay effects, but this feature could be extended to allow for scalding steam or blinding clouds of fog.

### DIFF
--- a/src/region.c
+++ b/src/region.c
@@ -1012,6 +1012,8 @@ genericptr_t p2;
 
     reg = (NhRegion *) p1;
     dam = reg->arg.a_int;
+    if (!dam)
+        return FALSE;
     if (p2 == (genericptr_t) 0) { /* This means *YOU* Bozo! */
         if (m_poisongas_ok(&g.youmonst) == M_POISONGAS_OK)
             return FALSE;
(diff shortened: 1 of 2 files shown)


copperwater:
Noticing some issues with the steam clouds. When a moat is zapped over, no moat squares render the cloud. Pools and fountains do it correctly. Try this on Medusa's Island.

The clouds also don't block vision across them (while standing outside the cloud from one side, you can see things happening on the other side of the cloud whereas you can't see things inside the cloud), but this is for some reason also the case with the regular poison gas clouds, which I would call a bug in its own right.

PR #344, opened 2020-05-14
Enable themed rooms to be constrained by level difficulty

copperwater:
The system of themed rooms currently makes it so that any themed room
can potentially generate anywhere a themed room can be placed. This is
problematic in the long run, since it makes it difficult to design new
rooms that are an appropriate amount of challenge at all levels of the
dungeon. (A few themed rooms already have this problem: a hero starting
out on level 1 probably won't live very long when the neighboring room
is full of giant spiders, or an arch-lich has generated in a mausoleum
nearby).

This pull request adds optional "mindiff" and "maxdiff" properties for
themerooms defined as tables and exposes level_difficulty() to Lua. A
themeroom whose mindiff exceeds the current level difficulty, or whose
maxdiff is lower than the current level difficulty, is prevented from
being selected.

I've also applied minimum difficulty cutoffs to two themed rooms that
probably pose too much of a challenge on the uppermost floors. Note
that this does NOT fix the "arch-lich in Mausoleum" problem, which I've
confirmed in wizard mode actually does happen - the solution I would 
recommend for this is pull request #253.


copperwater:
Something also to consider with respect to the spider nest room is the possibility of using the nh.level_difficulty() function to do an interesting mix of monsters, combining cave spiders with giant ones as the difficulty increases, like so:
```lua
for x=0,rm.width do
   for y=0, rm.height do
      if (percent(30)) then
         -- Gradually increase the chance of giant spiders
         if (nh.level_difficulty() - 6 < nh.rn2(10)) then
            des.monster("cave spider", x, y)
         else
            des.trap("web", x, y) -- makes a giant spider
         end
      end
   end
end
```

PR #347, opened 2020-05-18
Allow themed room subrooms to be filled

copperwater:
I noticed that any subrooms created within a themed room were bare - they never had any monsters, objects, traps, or anything really, regardless of whether filled = 1 was set on them. As a result, they're
pretty boring, and I wanted to fix this.

Then I noticed that special rooms incorporated into themed rooms (or just themed rooms defined with a type that wasn't "ordinary" or "themed") were being left empty as well. One thing led to another and I wound up refactoring some parts of the level creation code and fixing some bugs along the way. 

The broad strokes:
* Non-special subrooms of themerooms can now be filled (if filled=1), and get filled using the same routine as is used to populate a regular ordinary room.
* Special-type themerooms and subrooms of themerooms can now be filled.
* The confusing "prefilled" parameter is dropped; all des.room and des.region accept a "filled" table option. "filled=2" can be specified to produce a special room that sets level flags but isn't supposed to be stocked with anything (used mostly on Quests).
* Fixed bugs where monsters arriving on Ludios and wizard-1 would not be constrained properly.
* Identified the cause of occasional statues embedded in walls on medusa-2 and medusa-4.
* makelevel is restructured a bit; all special rooms on every level are left unstocked when they're first created, then stocked at the end of makelevel.


copperwater:
Merged.

PR #350, opened 2020-05-28
Reverting to original phrasing for pet displacement

G7Nation:
Comments:
- "displaced" was the original phrasing; no reason to change it
- more flexible for variants and future nethack versions:
  - hypothetical: perhaps a pet does not want to swap places into
    lava, but rather, be displaced to a spot of its own choosing
  - real-world: in slashem, the "displacer beast" exists because
    of the original wording
- regardless, past tense is appropriate

### DIFF
--- a/src/hack.c
+++ b/src/hack.c
@@ -1826,7 +1826,7 @@ domove_core()
             newsym(x, y);
             newsym(u.ux0, u.uy0);
 
-            You("%s %s.", mtmp->mtame ? "swap places with" : "frighten",
+            You("%s %s.", mtmp->mtame ? "displaced" : "frightened",
                 pnambuf);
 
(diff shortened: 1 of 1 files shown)


pat-rankin:
The wording was changed in preparation for having displacement work for monsters (hero sees displaced monster at wrong location).  That didn't happen--or perhaps hasn't happened yet--but is still a reason to not say "displace" when swapping occurs.

The displacer beast in slash'em came out of the AD&D Monster Manual, where is a creature that seems to be at a different location rather than one that can swap places with anything.  Slash'EM did not implement monster displacement and just glommed on the existing pet code.  NetHack has now added displacer beast but still uses pet swapping code, plus the 'bargethrough' monster capability.

PR #362, opened 2020-07-04
Avoid revealing mimics with mind flayer lock-on

entrez:
Using `cansee` for feedback on the mind flayer psychic attack means that the message identifies the target by name, even if it's a cloaked mimic or other hiding monster, as long as you can see the square it's on. Using `canseemon` instead will skip printing the message if the target is concealed in some way.


entrez:
Turns out I didn't understand how `canseemon` works, so this didn't do what it was supposed to and also had an unintended side effect of hiding feedback when invisible monsters are targeted.

PR #365, opened 2020-07-05
Use <lvl>d8 + 1 instead of <lvl>d8 for mon->mhpmax

entrez:
Monster sanity checking now raises an error when monster max hp is less than `mon->m_lev + 1` (as of 7d7b98f), but `newmonhp` still just rolls \<level\>d8 to determine a monster's starting (max) hp. This means monsters can start with hp equal to their level, if every die rolled is a natural 1 (e.g. 1d8 can end up a 1, 5d8 can end up a 5, etc); to avoid tripping this new sanity check, this changes the formula for determining a monster's max hp to \<lvl\>d8 + 1.

### DIFF
--- a/src/makemon.c
+++ b/src/makemon.c
@@ -1020,7 +1020,7 @@ int mndx;
     } else if (!mon->m_lev) {
         mon->mhpmax = mon->mhp = rnd(4);
     } else {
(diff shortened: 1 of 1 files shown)


entrez:
Rebased to 68638d7 & force-pushed so that Travis build will pass.

PR #371, opened 2020-07-11
make paranoid_confirmation:pray result in yes/no

### LINKED ISSUE #303: Paranoid pray prompt not paranoid enough
I would like to have the option to make `paranoid_confirmation:pray` require the player to type `yes` rather than just `y`. Out of all the paranoid prompts, why is prayer the only one that doesn't require an explicit `yes`?

In a recent game, I was repeatedly sacrificing corpses and praying for favors, something like
`#offer y #offer y #offer y ... #pray y #offer y #offer y ...`
Then, by mistake I prayed twice in a row and lost over 30 points of intrinsic protection. Since sacrifice asks for confirmation as well it's very easy to accidentally pray instead.

The quick and easy solution would of course be
```
diff --git a/src/pray.c b/src/pray.c
index b098311d4..9c8c41188 100644
--- a/src/pray.c
+++ b/src/pray.c
@@ -1846,7 +1846,7 @@ int
 dopray()
 {
     /* Confirm accidental slips of Alt-P */
-    if (ParanoidPray && yn("Are you sure you want to pray?") != 'y')
+    if (!paranoid_query(ParanoidPray, "Are you sure you want to pray?"))
         return 0;
 
     u.uconduct.gnostic++;
```
However that would likely annoy most people who are used to just hitting `y`.

paranoidtimes:
This change removed the default paranoid_confirmation:pray and the special logic that goes into paranoid_pray. Instead the default is paranoid_confirmation:none, and #pray will result in being prompted y/n (as was the previous default behaviour), if a player sets OPTIONS=paranoid_confirmation:pray they will then be prompted for 'yes' or 'no' if they attempt to pray. Bringing the pray difference in line with the other paranoid options, and preserving the y/n behaviour for those who don't set it.
fixes #303


paranoidtimes:
Not sure that I have access to link it myself, but this is in response to https://github.com/NetHack/NetHack/issues/303

pat-rankin:
The confirmation prompt for \#pray can now be changed to require 'yes\<return\>' instead just 'y' by also setting paranoid_confirmation:confirm.  That will have a side-effect of requiring 'no\<return\>' instead of just 'n' to decline (for all paranoid choices which are currently enabled).

Implemented by commit b1d30fbce94cdf26fb42ddebf9b42b097684e335 for issue #303, with several followup commits including 6cef49ed540ac8cd11f667dabe163423264a3add.

PR #374, opened 2020-07-13
Check bones data directly for deja vu messages

### LINKED ISSUE #322: Deja vu messages appear on non-bones floors with randomly named ghosts.
The change to amnesia made it so that you no longer forget maps, so deja vu messages can no longer appear in this case. Instead, these messages now appear when you enter a level containing a ghost whose name matches your own. It seems like this was intended to be seen on the player's own bones floors, but it also happens on levels with randomly named ghosts that choose the player character's name from the high score list (such as Moloch's Sanctum which cannot have bones, and you may not have 'been here before').

The message also would not appear on a bones floor without a ghost, such as if the character was petrified or turned into a wraith.

entrez:
After modifications to the amnesia mechanic, \`deja vu' messages are now displayed upon entering a level containing bones of a previous character of the current player. This test is done simply by checking for a ghost on the level that shares a name with the current character.

However, since ghosts generated in other circumstances (for example, in the Valley of the Dead and other special levels) can have names pulled randomly from the high score list, etc, this message can be displayed on non-bones levels where a ghost has been generated with the character's name. Additionally, when a bones pile doesn't include a ghost (such as when the character in question was slimed, killed by a wraith, etc), the \`deja vu' message will not be displayed when it should be. This is all described in in NetHack/NetHack#322.

This pull request would change the method of testing for `familiarity' by adding a function to iterate through any bones data for the current level, searching for a match with the hero's name.

Would close NetHack/NetHack#322.


copperwater:
This will leave the find_ghost_with_name function with no callers. This function was only added for the purpose of identifying a level that should print the deja vu message, so this pull request should delete it.

davidbau:
## seed0005 findings via `^dbg_dm_iter` instrumentation

Added per-iteration logging inside `dog_move`'s backtrack-avoidance loop to emit pet position, mfndpos count `k`, and `mtrack[0..3]` contents.  Diffed C and JS for mid=49 across turns 170-179.

### Turns 170-178: IDENTICAL

Every iteration across 9 turns matches perfectly between C and JS — same positions visited, same mtrack contents, same k.

### Turn 179: SUDDEN DIVERGENCE

| | C | JS |
|---|---|---|
| k | 5 | 8 |
| mt0 | 13,5 | 12,5 |
| positions | (12,5) (12,7) (13,5) (13,6) (13,7) | (12,5) (12,6) (12,7) (13,5) (13,7) (14,5) (14,6) (14,7) |

JS has 3 positions C doesn't include (`12,6`, `14,5..14,7`), and C has one JS doesn't (`13,6`).  mtrack[0] differs.

### What happens between turn 178 and 179

At the turn boundary, hero performs `.` (rest) at step 167.  Right before the pet's dog_move at turn 179, a monster spawns — C's `^makemon[322#96@15,5]`, JS's `^makemon[322#96@71,2]`.  Both sides spawn the same mid=96 eye of newt, but at completely different positions.

**RNG up to this point matches (seq 8116).**  If PRNG sequences agree, the random-placement roll inside makemon must agree.  So the spawn position must diverge via non-PRNG state.

### Hypothesis

`makemon()` at this call site picks a position using `enexto()` or similar, which scans valid tiles.  If the map/monster state at scan time differs between C and JS (e.g. different objects blocking tiles, different monsters at positions), the "first valid tile" picked differs even with identical PRNG state.

The 1-tile mtrack diff at turn 179 (`13,5` vs `12,5`) also suggests the pet's previous move at turn 178 went to a different tile — but turn 178 iteration matched, so the CHOICE within that iteration differed silently (rn2-dependent selection).  However RNG matched through this, so this should not have happened.  **So the pet must have moved to the same tile but mtrack's pre-shift state differed from an earlier turn**.

Actually more likely: mfndpos diverges silently earlier when the pet's current `mx/my` or neighboring tile state differs from C.  This ripples into different mtrack shifts and different k counts.

### Next diagnostic

Instrument `mfndpos` to emit its filter flags + returned positions.  If mfndpos inputs match but outputs differ, the problem is in mfndpos' internal filter logic.  If mfndpos inputs differ (i.e. the pet's own position or the tile state), trace further upstream.

### Unrelated supporting fix

Committed 510403e7f (bd84df332 after rebase): makemon event emission now uses `(mndx, mtmp.mx, mtmp.my)` matching C, not `(mnum, requested_x, requested_y)`.  This was unearthing false event-format divergences (seed0800 after shapeshift).  Plus new patch `032-umovement-tracing.patch` emitting `^u_calc_moveamt` for hero movement-points debugging.

Agent: agent:maud session:59a80440-c18b-4044-8f9e-fd5fe34bd70e

PR #377, opened 2020-08-07
Fix aborted worm placement failing sanity check

entrez:
`initworm(worm.c)` sets `wx` and `wy` of `wheads[worm->wormno]` to `worm->mx` and `worm->my`. Because `place_worm_tail_randomly` reverses all the segments of the worm in the process of placing them, however, this segment with a `wx` and `wy` matching the worm ends up being the final tail segment rather than the head.

When `place_worm_tail_randomly` is forced to abort early, it calls `toss_wsegs`, which removes the `g.level.monsters` entry for each segment's `wx` and `wy` (based on the assumption that a segment having a `wx` and `wy` means it has been placed on the map). Since the final tail piece has a `wx` and `wy` that matches the worm itself, the worm ends up being removed from the map, and therefore causes `mon_sanity_check(mon.c)` to fail -- the worm is still in the list of monsters, but nothing exists on the map where it's supposed to be.

This problem can be avoided by preemptively setting the `wx` and `wy` of `wheads[worm->wormno]` to `0` in `place_worm_tail_randomly` before starting to reverse the worm's segments/place the worm.

### DIFF
--- a/src/worm.c
+++ b/src/worm.c
@@ -728,6 +728,7 @@ xchar x, y;
         return;
     }
 
+    wheads[wnum]->wx = 0, wheads[wnum]->wy = 0;
     wheads[wnum] = new_tail = curr;
     curr = curr->nseg;
     new_tail->nseg = (struct wseg *) 0;


entrez:
To be more specific about the issue that this would fix, here's the actual sequence of events laid out.

1. The new worm is placed into `g.level.monsters[x][y]`:
https://github.com/NetHack/NetHack/blob/46f19f89acc2bc75a9286654bf4caa686e31c41c/src/makemon.c#L1265
2. The last segment of the worm, `wheads[worm->wormno]`,  is set to match the worm's position on the map:
https://github.com/NetHack/NetHack/blob/46f19f89acc2bc75a9286654bf4caa686e31c41c/src/worm.c#L121-L131
3. `place_worm_tail_randomly` uses `wtails[worm->wormno]` as the head, and delegates `wheads[worm->wormno]` to be the (eventual) final tail segment:
https://github.com/NetHack/NetHack/blob/46f19f89acc2bc75a9286654bf4caa686e31c41c/src/worm.c#L722-L735
4. On failure to place a segment, `place_worm_tail_randomly` calls `toss_wsegs` to delete all the unplaced segments:
https://github.com/NetHack/NetHack/blob/46f19f89acc2bc75a9286654bf4caa686e31c41c/src/worm.c#L755-L757
5. `toss_wsegs` removes the monster at `g.level.monsters[seg->wx][seg->wy]`, which winds up removing the worm when it hits the final unplaced segment:
https://github.com/NetHack/NetHack/blob/46f19f89acc2bc75a9286654bf4caa686e31c41c/src/worm.c#L152-L160

PR #385, opened 2020-08-30
libnethack

apowers313:
This PR has two new builds:
* `libnethack.a` -- a unix-ish library that can be linked against
* `nethack.js` -- a WebAssembly build of NetHack that can be run in the browser or node.js. This is also prepared to be published as a public [npm](https://www.npmjs.com/) module under `sys/lib/npm-package`.

See `sys/lib/README.md` for more details details about how to build and the library APIs.


apowers313:
Published here: https://www.npmjs.com/package/@neth4ck/neth4ck

I created a new `neth4ck` organization on NPM. If anyone wants to be added to the organization to control packages let me know and I will happily add you.

(note: npm doesn't allow organizations or packages with `hack` in the name, thus `h4ck`)

moon-chilled:
IMO `WASM` is better than `__EMSCRIPTEN__`.  Emscripten needn't be the only c compiler targeting webassembly.

moon-chilled:
`WASM` is better than `__EMSCRIPTEN__`.  Emscripten needn't be the only c compiler targeting webassembly.

apowers313:
@moon-chilled Not sure it matters much in the context of this PR since we can treat alternate WebAssembly approaches similar to other toolchains in the Unix build. That said I'm having a hard time parsing your comment and I'd like to learn more. My understanding is that "WASM" is just short for "WebAssembly" and the alternative to creating WASM with [Emscripten](https://emscripten.org/) is [a process that wouldn't scale to building NetHack](https://surma.dev/things/c-to-webassembly/).

Is there a trick or a toolchain that I'm unaware of?

nhmall:
I've resolved the new conflicts upstream and merged the changes (I think). They will be available later this evening.
The PR had been sitting dormant for 22 days, but was updated while merging earlier today.  The updates conflicted
because the PR was modified from original upon merge:

1. instead of separate new Makefiles, I adapted it to use the existing sys/unix/Makefiles and the new cross-compiler insertion and overrides, similar to how other recent cross-compiles are doing it.
2. In the updates that will be merged in a few hours, sys/lib/libnethackmain.c is included now, since more extensive code edits have been added in the changes today.

nhmall:
@apowers313, I think @moon-chilled was just referring to the choice of conditional code moniker (#ifdef X vs #ifdef Y) used in the sources, not another alternative toolchain.

nhmall:
Assuming that you have already installed emscripten on your macOS or Linux system,

cd sys/unix ; ./setup.sh hints/macOS.2020 ; cd ../..
  or
cd sys/unix ; ./setup.sh hints/linux.2020 ; cd ../..

make CROSS_TO_WASM=1        (results in targets/wasm)
make WANT_LIBNH=1                (resulting libnethack.a in src)

@apowers313 : thanks for starting this. We'll see where it might eventually go.

Issue #329, opened 2020-04-17
Pets become stuck within a shop


k21971:
Have had several reports from multiple players of their pets freezing/becoming stuck either inside a shop or at the threshold. Most reports state that a mimic is in the shop, and the pet will be frozen a couple tile spaces away from it. The pet can still do things (picking up/dropping objects), it just won't move from the spot it gets stuck in.

I have a few saves from our 370-hdf build I can share if necessary.

Also, please see this reddit thread - https://www.reddit.com/r/nethack/comments/g13uy3/pets_freezing_in_place/

ostrosablin:
Happened to me today as well (in my case, two dogs at once got stuck).

Not sure if it's related, but there were two mimics disguised as strange objects.

Dogs were undisplaceable and stuck on single square, picking up and dropping objects under them. But as I've hit and awakened both mimics - dogs suddenly un-stucked and began to follow me as usual.

I also suppose it might be related to c8fb419a2c1e71432b876c107e58ec6084aaa979. It's pretty consistent with newly added `return 3;` lines. Dog appears to be stuck in state, when it didn't move and probably can't do anything. It can't attack mimics because they're disguised (but it feels like dogs are aware about their presence) and it can't move for some reason. So I guess, they just stay in one spot, until mimics awaken.

tmp6154:
This if clause in m_move is shop-related and it now triggers due `return 3;` statements from above commit. Not sure, but might be somehow related.

https://github.com/NetHack/NetHack/blob/c8fb419a2c1e71432b876c107e58ec6084aaa979/src/monmove.c#L1534

copperwater:
I've done some sleuthing on a file that was displaying this pet behavior. I think the culprit is in dog_move in dogmove.c: 
```c
        /* Identify the best target in a straight line from the pet;
         * if there is such a target, we'll let the pet attempt an
         * attack.
         */
        mtarg = best_target(mtmp);

        /* Hungry pets are unlikely to use breath/spit attacks */
        if (mtarg && (!hungry || !rn2(5))) {
            int mstatus;

            if (mtarg == &g.youmonst) {
                if (mattacku(mtmp))
                    return 2;
            } else {
                mstatus = mattackm(mtmp, mtarg);

                /* Shouldn't happen, really */
                if (mstatus & MM_AGR_DIED)
                    return 2;

                /* Allow the targeted nasty to strike back - if
                 * the targeted beast doesn't have a ranged attack,
                 * nothing will happen.
                 */
[...]

PR #390, opened 2020-09-18
Fix the "stuck pets" bug (github issue #329)

copperwater:
This commit is intended to fix the bug where a pet will get fixated on
an unmoving monster and stop moving itself. I described the cause in the
github issue; the gist is that the pet AI chooses the unmoving monster
as its ranged target, doesn't do anything when it calls mattackm
(because it doesn't have ranged attacks), then returns a value
indicating it didn't move and can't take further actions.

I initially implemented a fix that refactored mattackm to distinguish
between "attacker missed" and "attacker did nothing", which the pet AI
could then use to determine whether the pet could continue doing things.
But then I realized that if mattackm is called with non-adjacent
monsters, a return of MM_MISS more or less unambiguously indicates that
the attacker did nothing (because the ranged functions it calls like
breamm don't actually check to see whether the target was hit, just
whether the monster initiated the attack.) So, this only really needed
to check whether mattackm returned with MM_MISS.

I also found a probable bug in mattackm, in that the thrwmm call isn't
treated the same as breamm or spitmm. In the latter two, mattackm
returns MM_HIT even though it doesn't check whether the ranged attack
actually hit its target. But there was no logic doing the same for
thrwmm, so this commit also adds that. (Otherwise, a pet could possibly
use a ranged weapon attack and then get to keep moving on its turn.)

### DIFF
--- a/src/dogmove.c
+++ b/src/dogmove.c
@@ -1156,11 +1156,17 @@ int after; /* this is extra fast monster movement */
 
         /* Hungry pets are unlikely to use breath/spit attacks */
         if (mtarg && (!hungry || !rn2(5))) {
-            int mstatus;
+            int mstatus = MM_MISS;
 
             if (mtarg == &g.youmonst) {
                 if (mattacku(mtmp))
                     return 2;
+                /* Treat this as the pet having initiated an attack even if it
+                 * didn't, so it will lose its move. This isn't entirely fair,
+                 * but mattacku doesn't distinguish between "did not attack" and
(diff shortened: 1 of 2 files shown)


ostrosablin:
I've been waiting for this patch for so long. A major game breaker for now, which (for me) spoils all the fun from playing 3.7.0 beta, since pets don't feel right and can get stuck at any moment.

Update: Tested for a while a build with this patch included. Works like a charm. Pets no longer get stuck and don't see any oddities. Seems like a perfect fix.

copperwater:
Merged.

PR #404, opened 2020-10-21
Prevent sokoban levels being flipped

amateurhourrrr:
Sokoban levels in particular being flipped is a cause of a frustration,
particularly for long-time players whose muscle memory gets confused.

Note that the player confusion is different from if there were new levels
instead (my preferred solution) that must be done.

To a lesser extent it affects new players attempting to follow a guide,
though guides can be independently updated.

### DIFF
--- a/dat/soko1-1.lua
+++ b/dat/soko1-1.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 des.map([[
 --------------------------
 |........................|
--- a/dat/soko1-2.lua
+++ b/dat/soko1-2.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 
 des.map([[
   ------------------------
--- a/dat/soko2-1.lua
+++ b/dat/soko2-1.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 
 des.map([[
 --------------------
--- a/dat/soko2-2.lua
+++ b/dat/soko2-2.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 
 des.map([[
   --------          
--- a/dat/soko3-1.lua
+++ b/dat/soko3-1.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 
 des.map([[
 -----------       -----------
--- a/dat/soko3-2.lua
+++ b/dat/soko3-2.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "premapped", "solidify", "noflip");
 
 des.map([[
  ----          -----------
--- a/dat/soko4-1.lua
+++ b/dat/soko4-1.lua
@@ -35,7 +35,7 @@
 --## Bottom (first) level of Sokoban ###
 des.level_init({ style = "solidfill", fg = " " });
 
-des.level_flags("mazelevel", "noteleport", "hardfloor", "premapped", "solidify");
+des.level_flags("mazelevel", "noteleport", "hardfloor", "premapped", "solidify", "noflip");
 
 des.map([[
 ------  ----- 
--- a/dat/soko4-2.lua
+++ b/dat/soko4-2.lua
@@ -4,7 +4,7 @@
 --
 des.level_init({ style = "solidfill", fg = " " });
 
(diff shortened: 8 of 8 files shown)


ostrosablin:
Agreed. In my last game, I too, reached sokoban and got a flipped 2a level, which on it's own a single most difficult sokoban level in NetHack, and it's just been made more annoying by flipping, messing with my memory.

Flipping level causes annoyance for long-time players while adding dubious value of making sokoban more difficult (not challenging, or more interesting - just plain more difficult and annoying, because it's same old levels, just flipped, causing frustration, because it feels confusing).

I propose either to revert flipping and keep with old behavior, or, like @gebulmer suggested - solve level diversity by adding true new levels (e.g. like UnNetHack did).

Personally, my preferred solution is just to undo flipping. I think, many NetHack players (me included) feel like a truly challenging and diverse Sokoban puzzles feels out of place in NetHack, because when you can get stumped for hours with a difficult level, it's just frustrating and intrusive, because it just draws you out from exploring dungeons and NetHack's gameplay. So, in my opinion, 8 default levels are perfectly fine (and can be done with muscle memory). They allow to view Sokoban as a refill branch, which allows to collect some food, rings, wands and ultimately, the Sokoban prize. And it anyway presents a challenge, even if you remember how to solve each level (which can be unfortunate monster generation,  forcing you to break boulders and incur luck penalty upon you, strong monsters that sometimes generate in Sokoban, which might provide a challenge, and even might crowd and kill you outright, if you're not prepared, this is especially true for a treasure zoo on the last level).

bhaak:
Level flipping was already longer in UnNetHack and over time, I tended to reduce level flipping for regular levels (a flipped mine town doesn't do much for level variation but hampers levels being recognisable or memorable).

But Sokoban levels being flipped, I still consider a good thing. But this is in UnNetHack, which has only 3 levels, additional Sokoban levels, and no luck penalties.

I'm not sure if muscle memory is a good argument here. Players didn't need to adjust their muscle memory for over a decade but if you want change, that's something that happens. With new Sokoban levels, this would also destroy their muscle memory for those new levels.

amateurhourrrr:
There's a difference between having no muscle memory for a new level (fine, can learn and pick it up) and getting confused and frustrated by the level you 'know' being different (not fine, have to unlearn in order to progress)

This isn't a hill I'm particularly wanting to die on, but that's my personal reason for not liking it

Others have expressed similar in IRC, so I figured I'd make the PR and see if it leads anywhere

copperwater:
Paraphrasing my sentiment from IRC - I agree with removing the flipping, largely on the basis that the vast majority of players who have expressed an opinion about flipped sokoban have shown dislike for it. (Still waiting on bhaak to produce evidence of a player other than himself that says they like flipped sokoban for some positive reason - other than hurting their brain or giving them something to hate on 3.7 for.)

I don't think the muscle memory based argument for replacing sokoban flipping with new levels is a particularly strong one, because then players who prefer just coasting through sokoban on muscle memory get dissatisfied. Still, I'd like to see that happen, because I'd like to see flipped sokoban removed for its unpopularity, and adding new levels is the cleanest and easiest way to escape from being stuck in 3.4.3-era sokoban forever. (What I see as the proper solution to satisfy both Sokoban-lovers and Sokoban-haters who are in it only for the prize is to provide a parallel non-puzzle branch, but this would require some overhauling to make it work...)

k21971:
Perhaps make it an option the player can toggle on/off via their config - soko_flip TRUE or FALSE, has to be set at game start. bhaak may be correct that some players enjoy the flipped sokoban levels... but I have yet to hear any positive mention of it. So, to appease both those that like it and (the vast majority) of those that do not, make it an option.

jeremyn:
I haven't tried 3.7 or encountered flipped Sokoban levels but it sounds pretty unfun to me. I don't have good muscle memory for any of the levels, just half-remembered solutions from previous plays, but most of them aside from 2a are easy enough to solve with a few moments thought, which is all right. Flipping just artificially increases the difficulty, but there are other ways to do that, like adding more boulder mimics, and I don't think that would be fun either.

If we're spitballing ideas here, I think it'd improve Sokoban to add a "lever" (special square) to each level that automatically fills in the pits if the player wants to skip solving the puzzle. Penalties, maybe randomly assigned, could include: a large luck or alignment hit; replacing the pits with slightly-harder-than-expected monsters; deleting remaining food/rings/wands on the level or from inventory; cursing player inventory. Put the lever at the end of a short hallway and before it add an engraving with a warning.

Adding extra, optional levels, maybe in an upstair from the current treasure zoo, could be fun. If each level has a bunch of food and one or two goodies like rings or wands, solving each could be its own reward. Apologies for the self-promotion, but if we want more levels, years ago I wrote a game called [Roguelike Sokoban](https://github.com/jeremyn/roguelike-sokoban) which plays like NetHack and where I've adapted many public domain XSokoban levels, visible in my repo [here](https://github.com/jeremyn/roguelike-sokoban/tree/master/levels). I'd be happy to help adapt them to NetHack if there's any interest.

jeremyn:
I haven't tried 3.7 or encountered flipped Sokoban levels but it sounds pretty unfun to me. I don't have good muscle memory for any of the levels, just half-remembered solutions from previous plays, but most of them aside from 2a are easy enough to solve with a few moments thought, which is all right. Flipping just artificially increases the difficulty, but there are other ways to do that, like adding more boulder mimics, and I don't think that would be fun either.

If we're spitballing ideas here, I think it'd improve Sokoban to add a "lever" (special square) to each level that automatically fills in the pits if the player wants to skip solving the puzzle. Penalties, maybe randomly assigned, could include: a large luck or alignment hit; replacing the pits with slightly-harder-than-expected monsters; deleting remaining food/rings/wands on the level or from inventory; cursing player inventory. Put the lever at the end of a short hallway and before it add an engraving with a warning.

Adding extra, optional levels, maybe in an upstair from the current treasure zoo, could be fun. If each level has a bunch of food and one or two goodies like rings or wands, solving each could be its own reward. If we want more levels, years ago I wrote a game called [Roguelike Sokoban](https://github.com/jeremyn/roguelike-sokoban) which plays like NetHack and where I've adapted many public domain [XSokoban](http://www.cs.cornell.edu/andru/xsokoban.html) levels, visible in my repo [here](https://github.com/jeremyn/roguelike-sokoban/tree/master/levels). I'd be happy to help adapt them to NetHack if there's any interest. You can also use it to try out new Sokoban levels really easily, since you just type the new level map into a text file and then pass the text file as an argument to the main program.

RojjaCebolla:
Yeah, soko flipping doesn't make the level more challenging or more interesting, it's just frustration while you get the hang of reversing your usual inputs.  New levels would be approximately 400% more interesting than the minor interface screw produced by transforming existing, memorized levels.

PR #407, opened 2020-10-29
`Readable' Hawaiian shirt designs

entrez:
Functionally similar to reading a t-shirt or apron, but rather than actual text printed on the shirt being read and displayed, the design of the Hawaiian shirt is instead described.

``The design features hula dancers on a yellow background''

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -15,6 +15,7 @@
 static boolean learnscrolltyp(short);
 static void cap_spe(struct obj *);
 static char *erode_obj_text(struct obj *, char *);
+static char *hawaiian_design(struct obj *, char *);
 static int read_ok(struct obj *);
 static void stripspe(struct obj *);
 static void p_glow1(struct obj *);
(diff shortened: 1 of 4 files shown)


src/read.c
```diff
@@ -157,6 +157,35 @@ char *buf;
...
+    static const char *aloha_styles[] = {
+        "flamingos", "parrots", "toucans", "birds of paradise", /* birds */
+        "angelfish", "tropical fish", /* fish */
+        "plumerias", "orchids", "hibiscus flowers", "palm leaves", "palm trees", /* plants */
+        "hula dancers", "sailboats", "surfboards", "ukuleles", /* other */
+    };
+
+    static const char *aloha_bgs[] = {
+        "purple", "yellow", "red", "blue", "white", /* solid colors */
+        "abstract", "geometric", "patterened", "checkered", "naturalistic", /* adjectives */
```
actual-nh: "patterned"?

entrez: Thanks for pointing this out!

actual-nh: Quite welcome. (As it happens, I was born in Hawaii...)

src/read.c
```diff
@@ -156,6 +156,34 @@ char *buf;
...
 }
 
+char *
+hawaiian_design(shirt, buf)
+struct obj *shirt;
+char *buf;
+{
+    unsigned style, bg;
+    static const char *hawaiian_styles[] = {
+        "flamingos", "parrots", "toucans", "birds of paradise", /* birds */
```
jeremyn: Any interest in making the specific elements more "NetHack-y", like the regular T-shirt messages? Krakens instead of sea turtles, harps instead of ukuleles, silver (dragon) background instead of purple, that sort of thing?

entrez: I do like this idea, which is why "giant eels" is already in there, but I want to make sure the motifs that are included have a sort of tropical feel that makes the description seem like something you might believably see on a Hawaiian shirt.  Piranhas would probably fit well too, come to think of it.

That effect could also be achieved by adding more verbiage to the monster names, e.g. something like "hula-dancing water nymphs", but the description length is a concern since the whole thing is printed in the end-of-game inventory disclosure list. With this patch applied, a Hawaiian shirt already has a much longer description than most other items, and I'm reluctant to use motifs or backgrounds that will make it even longer. To be honest, I have been debating removing the backgrounds entirely for that reason (to reduce the length), even though I like the way they add to the perceived loudness of the shirt.

entrez: I included some more NetHack monsters in the commit I just added, which was mostly intended to address the other problem I was talking about (wordiness in end-of-game disclosure list).

copperwater:
I would re-add the "sea turtles" and "tropical fish" monsters even though they don't exist in NetHack, because they're very common Hawaiian shirt motifs. (Not angelfish, those are mostly tropical fish.)

entrez:
I didn't want the list to become too biased towards generating sea creature motifs: if all the current sea creatures were kept on there, plus "sea turtle" and "tropical fish", the chance of a particular Hawaiian shirt having a sea creature of some sort on it would be ~44%. That seems pretty high to me.

But maybe it would be worth it to ditch "electric eel" (which is almost the same as "giant eel") and/or "piranha" (which brings jungle to mind more than tropical isle) to reintroduce those motifs.

jeremyn:
I've changed my mind from my earlier comment and think that going easy on NetHack references is for the best. What makes the shirt funny, both here and on Twoflower in the Discworld series where it originates, is that it's so wildly out of place for the setting. Making it heavily NetHack-oriented goes against that. So I agree with re-adding sea turtle etc.

You could add a minor Discworld shout-out to Twoflower's shirt's description in *The Colo[u]r of Magic* which is mentioned when he's introduced:

> [He] was small and skinny, and dressed very oddly in a pair of knee length britches and a shirt in such a violent and vivid conflict of colors that Weasel's fastidious eye was offended even in the half-light.

For example you could return a fixed description of "The design/It has such a violent and vivid conflict of colors that your eye is offended even in the half-light." say, 5% of the time.

Also, and I'm sure you've considered this but just to say, if you replaced the random composition approach with a hardcoded list of ten or twenty descriptions, you could control for length and also maximize humor, such as it can be. But your random composition approach is good too.

PR #409, opened 2020-11-10
Spell cancelling

AndrioCelos:
This branch reorganises `spell.c` so that it is now possible to cancel casting most spells at the direction or target prompt.

- `tryspell` checks conditions that would prevent casting, and handles Amulet energy drain.
- `targetspell` prompts for a direction or target (for skilled fireball or cone of cold).
- `spelleffects` now handles the actual spell effects, and assumes the spell is successful.
- `docastspecial` is now used to cast a specific spell via a command other than `Z` instead of `spelleffects`. It can skip the direction/target prompt (for `^T`).

However, there are some considerations regarding these changes:

- Wands aren't changed. If you cancel at the direction prompt when zapping a wand, it does nothing but still uses a charge.
- The message 'Never mind' is normally shown if you cancel at the direction prompt.
- The roll for spell success is still done before the direction prompt, so you can cancel after this roll.
- If you cancel after Amulet energy drain has occurred, it still uses a move. The message is replaced with 'Nothing happens' in this case.
- The jumping spell hasn't been changed. With teleport away, you can cancel at the direction prompt, but not at the prompt for where to teleport to if you have teleport control (doing so still casts the spell, but you don't actually teleport).
- For skilled fireball or cone of cold, picking an invalid target or casting the spell underwater or on the Plane of Water is considered the same as cancelling it. Some messages may need to be changed.

This branch also fixes a minor bug whereby unskilled fireball and cone of cold were subject to half physical damage when cast at yourself, and changes the message 'You don't have enough energy to cast that spell' to 'You no longer have enough energy to cast that spell' if you had enough energy prior to the attempt but didn't after Amulet energy drain.


FredrIQ:
What about making the Amulet energy drain consistent (say, 1.5x) so you can cancel spells properly while dealing with the amulet drain smoothly?

PR #419, opened 2020-12-10
Print a message for the juiblex change

RojjaCebolla:
To help clue people in who are using the old strategy.

### DIFF
--- a/src/dig.c
+++ b/src/dig.c
@@ -1420,6 +1420,7 @@ zap_dig()
                 You("pierce %s %s wall!", s_suffix(mon_nam(mtmp)),
                     mbodypart(mtmp, STOMACH));
             if (unique_corpstat(mtmp->data))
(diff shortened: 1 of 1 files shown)


copperwater:
I think people who are using the old strategy will tend to be clued in when they dig out of Juiblex and hit him and he doesn't die instantly. One could also expect that a "half-shredded" monster is one hit point away from dying anyway.

Also, he's a giant mass of mostly goo; goo doesn't seem very shreddable to me.

jeremyn:
Another problem is that if you zap Juiblex twice you're going to get "half-shredded" both times, when maybe the second time should be "quarter-shredded" (don't do that though, that's silly).

There should be some message though, to show that you've hurt it, and not just tickled it so it vomits you out. Perhaps "You've blasted a large hole out of Juiblex!"

PR #420, opened 2020-12-11
Add liquid flow to land mine explosions

entrez:
Land mine explosions currently do not call `liquid_flow(dig.c)`, and as a result the pit created by an exploding land mine will never fill with adjacent water or lava, as pits created by other sources -- digging, breaking a wand, and earthquake -- can do.

This commit adds the appropriate calls to `liquid_flow` and `fillholetyp` to `blow_up_landmine` so that land mine explosions may fill with water like other pits do.

The call to `losehp` in `dotrap` had to be moved from after to before `blow_up_landmine`, since waiting to call `losehp` when the pit can fill with water could lead to silly messages (\`\`That was a close one. You die...'').  After this change, a land mine that killed a character would be retained unexploded in a bones file, because death would occur before the call to `blow_up_landmine`.  To avoid this issue, the land mine is converted to a pit before calling `losehp`; `blow_up_landmine` does not check whether the target trap is in fact a landmine so works as usual even if the trap is converted to a pit, and will delete the pit in cases where it should not exist.

(I closed #414 and rebased the branch after recent changes to trap.c introduced merge conflicts that needed to be resolved manually, so this is effectively the same PR as that one)

### DIFF
--- a/src/trap.c
+++ b/src/trap.c
@@ -2147,11 +2147,15 @@ unsigned trflags;
             set_wounded_legs(RIGHT_SIDE, rn1(35, 41));
             exercise(A_DEX, FALSE);
         }
+        /* add a pit before calling losehp so bones won't keep the landmine;
+         * blow_up_landmine will remove the pit afterwards if inappropriate */
+        trap->ttyp = PIT;
+        trap->madeby_u = FALSE;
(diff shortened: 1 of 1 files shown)


paxed:
Added in 2df4f08c0ba7a247e0d09efb19f592f69c93631c

PR #428, opened 2020-12-24
Fix some dismount loopholes

entrez:
Dismounting doesn't currently test whether the destination spot could be moved to normally, so can be used to squeeze diagonally into a space that would normally be blocked because "you are carrying too much to get through". This can be abused in Sokoban to pass through gaps that are normally inaccessible, and even in normal levels can leave the hero in a spot from which they can't escape without dropping items or using a pick-axe. This patch would block dismounting through diagonals which are normally impassible by adding a successful result from `test_move` as a condition of choosing a dismount location.

Additionally, dismounting will opt to place the hero onto a boulder if no other appropriate spots are available. This too can be abused in Sokoban, so these changes would also add the standard luck penalty for using this behavior to pass over boulders on an unsolved Sokoban level.

### DIFF
--- a/src/steed.c
+++ b/src/steed.c
@@ -452,7 +452,8 @@ int forceit;
                 if (!isok(x, y) || (x == u.ux && y == u.uy))
                     continue;
 
-                if (accessible(x, y) && !MON_AT(x, y)) {
+                if (accessible(x, y) && !MON_AT(x, y)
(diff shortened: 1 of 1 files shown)


entrez:
Merged via a29b76e1bf9f8df8d4153067c3a3e89c81115bfa and e7e418c1ec723d6daa8e8136c7a58dfd9d941298

Issue #368, opened 2020-07-09
Remove gender preferences for Foocubi


todb-r7:
They're demons, so they probably don't have the sexual hangups that many mortals of the Nethack period might profess.

Also, they're demonically beautiful and supernaturally skilled at what they do, so they should have just as much chance of seducing people of any gender, even if they're not the preferred gender of that character.

Oh, and sexuality is a spectrum and assuming hetero-only interactions ain't woke.

Rhialto:
On Wed 08 Jul 2020 at 17:02:19 -0700, Tod Beardsley wrote:
> They're demons, so they probably don't have the sexual hangups that many mortals of the Nethack period might profess.
> 
> Also, they're demonically beautiful and supernaturally skilled at what they do, so they should have just as much chance of seducing people of any gender, even if they're not the preferred gender of that character.
> 
> Oh, and sexuality is a spectrum and assuming hetero-only interactions ain't woke.

I already proposed this around the times of NetHack 3.0 or so (when I
made the (second?) Amiga port), but back then it was rejected...

todb-r7:
Yeah, but maybe the time is right now!

k21971:
The only issue with this request, is that a foocubi sole purpose was to seduce and *reproduce*, creating half-human/half-demonic offspring (Cambion). Nothing against female/female or male/male hook-ups, but no matter how hard they try, those combos won't produce any offspring. But hey don't take my word for it - look it up. The lore goes back centuries.

todb-r7:
As a Sandman/Lucifer fan and general occultism fanboy, I am embarrassed to admit that I've never heard this explanation of why foocubi exist! That said, I turn to the blasted and eldritch pages of the _Wikipaedia Wikifactorum_, [here](https://en.wikipedia.org/wiki/Cambion), to quote:

> Because of this inability to create or nurture life, the method of the creation of a cambion is necessarily protracted. A succubus will have sex with a human male and so acquire a sample of his sperm. This she will then pass on to an incubus. The incubus will, in his turn, transfer the sperm to a human female and thus impregnate her.

This whole incubation of Cambion stock not only doesn't require direct genetic transfer, but specifically excludes it on the part of the foocubi. Thus, same-sex relations can (and should!) still be a part of this process.

copperwater:
[...]

PR #430, opened 2020-12-27
Art Contribution: Differentiating gendered monster tiles.

NullCGT:
This pull request is a response to 0c3b9642e4207ed3c0b756585b83477f59116305, in which nhmall expressed interest in contributions that would make gendered tiles visually distinguishable from one another. Since I've spent way too many hours editing NetHack's default tileset and the thought of trying to merge this commit into my variant gives me an absurdly massive headache, I thought I would have a go at it!

Making tiles of different genders distinct in NetHack presents an interesting problem. While it would be fun to create highly distinct tiles for every gender, doing so would reduce the accessibility of the game, since players would have to remember many more tiles, and might end up confusing one monster for another. Visual clarity is key.

Therefore, I had the following goals when creating this pull request:
1. If there is an interesting way to differentiate tiles by gender, do so.
2. Any sort of differentiation should be minor enough that a user can still tell what a monster is at a glance. Essentially, visual clarity comes before differentiation by gender.
3. Try to use a "TDTTOE method" of differentiating tiles. For example, female cats are more colorful than males, because generally male cats have only two colors of fur. Basically, I spent a lot of time on wikipedia researching sex characteristics of different species.
4. Try not to fall into "female = longer hair / eyelashes." While this feature will unfortunately require some gender-essentialist visual shorthand, this tropes is overdone and exhausting.

Please let me know what you think; I'm totally open to feedback on all of this and happy to make modifications. I've attached the resulting tiles file to this post in png form.

The alterations made in this pull request are as follows:

- Female ants are slightly larger than male ants, just like in real life. I could have added wings to the male ants, but I felt that doing so would lead to some confusion.
- Female wolves are slightly smaller than male wolves. There wasn't a great way to show this without making winter wolves look very similar to winter wolf cubs, so I just made the female wolves tails slightly shorter.
- Calico cats are almost exclusively female, so I turned the female cats into calico cats. The other piece of logic behind this choice was that players will probably really enjoy seeing different variants of their pets.
- Female hobbits, minotaurs, humans, werecreatures, and aleaxes wear slightly different clothing.
- Dwarfs are **not** differentiated in any way whatsoever. According to Terry Pratchett (in Unseen Academicals, if I remember correctly) it is almost impossible to tell what gender a dwarf is, even for fellow dwarfs. I strongly believe that NetHack should follow this tradition.
- Female leprechauns, archons, frost giants, guards, and all types of gnomes are clean-shaven. Although of course not one hundred percent accurate, it's convenient visual shorthand.
- Centaur tiles have no differentiation because the different types of centaurs are already extremely difficult to tell apart from one another.
- Female ogre tyrants and elven monarchs have slightly different crowns.
- Female quantum mechanics have a different hairstyle and no beard. Genetic engineers look the same, because the genetic engineer tile is perfect.
- Female barrow wights look like old grandmothers with flyaway hair. I kept the hair color the same and used a similar quantity of pixels so that they look similar enough to the males that you can tell they are barrow wights.
- Female archeologist tile is a reference to a certain archeologist known for raiding tombs.
nhtiles


k21971:
This is awesome.

k21971:
Me nitpicking here - any worker ants or bees one would encounter are always female. In monst.c killer bees are already flagged as M2_FEMALE, any ants should be as well.

copperwater:
Though you might find male bees or ants in an anthole or beehive. This is a great pull request @NullCGT!

nhmall:
Thank you for contributing this!

jeremyn:
Hi, I'm a total nobody in this project, just a long-time player, but by sheer coincidence I just made a comment on the foocubi gender issue at https://github.com/NetHack/NetHack/issues/368#issuecomment-751554066 where I suggested non-binary NPCs and used a captain as an example: "The captain swings their silver saber! They miss!" I'm wondering how a non-binary captain would be displayed using this tileset. Maybe pick one at random? Please don't take this as criticism, the art here is fantastic. If this question is out of scope for this PR then please let me know.

Also, I think that in Discworld female dwarves were originally totally indistinguishable from male dwarves, but due to the "corrupting" influence of Ankh-Morpork, some female dwarves "scandalously" started wearing female clothing. They stayed bearded, though. However, the underground dwarf community was more conservative and kept all-male dress for longer, and since NetHack is underground we can assume the dwarves we see are part of that culture, so not differentiating them here seems fine to me, but there is the option of creating a more female-looking dwarf tile and assigning it randomly to some of the female dwarves to reflect the cultural shift. That's probably a totally unnecessary complication though.

NullCGT:
@jeremyn 

You're totally right on the Pratchett lore! It's been far too long since I've read Unseen Academicals.

Not to advertise myself, but I maintain the variant SpliceHack, which offers the option for players to be nonbinary, and (currently somewhat limited) support for nonbinary monsters. I'll be pulling the gendered tiles commit into SpliceHack at some point in the future, at which point I'll have to figure out how to reconcile the changes here with existing nonbinary player tiles that I created. Feel free to pop over there and look at that implementation.

Handling nonbinary entities is definitely outside the scope of this feature, though. It required a huge amount of work in order to get the game to smoothly handle nonbinary characters in the first place, and I still handle nonbinary NPCs with only a moderate amount of success.

jeremyn:
Sure, implementing is definitely outside the scope of this artwork PR. I suppose the concern is futureproofing, like if you had a set of tiles where one looked definitely masculine and the other definitely feminine, and then non-binary players/monsters were added, which tile would you assign? The Indiana Jones/Lara Croft tiles (they're the ones under Death, right?) for archaeologists are an example.

Anyway -- if this interests you and you want to take the time, I'm certainly not saying you should feel obligated -- you might want to take this opportunity to create a third less gender-coded tile for those few edge cases where the male and female tiles are both "too" gendered, anticipating and enabling future NetHack changes, even if the new tiles aren't used now. Maybe you could bring some over from SpliceHack, if you've already done the work there.

(I use ASCII graphics anyway, not tiles, so this wouldn't affect my play either way, full disclosure.)

NullCGT:
Followup sent.

PR #437, opened 2021-01-05
Add the monstercolor system, letting players choose monster colors

copperwater:
This existed as one of the 3.4.3-nao patches, but rather than porting that patch to the current codebase I started from scratch, because the system as I saw it had two big drawbacks:

1. Mutated mons[], which is just not a great idea in my book.
2. Colors were only settable via config file, not in-game. I envisioned monster colors working like msgtypes or autopickup exceptions, which can be modified in-game.

To that end, monstercolors are settable in-game via a submenu in the "Other settings" section of the options, like autopickup exceptions or msgtypes, as well as via "MONSTERCOLOR=monname:color" in the config. Also like those settings, they do not persist through saving and restoring, but any config lines will be loaded as normal upon restore.

Instead of directly editing mons[], it edits a new variable 'monstercolors' in instance_globals. I thought this might be unnecessary
at first, but then realized that if it changed mons[].mcolor like the NAO patch, then it would be impossible to count the number of currently set monstercolors because there would be no default to compare against.

### DIFF
--- a/include/color.h
+++ b/include/color.h
@@ -56,4 +56,8 @@ struct menucoloring {
     struct menucoloring *next;
 };
 
+/* For monstercolors: special value that indicates a monstercolor isn't
+ * configured for a given monster and it should use the default */
+#define MONSTERCOLOR_DEFAULT (CLR_MAX + 1)
+
 #endif /* COLOR_H */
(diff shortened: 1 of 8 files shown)


copperwater:
This has partially been superseded by the NetHack 5.0.0 glyph option, but it has some limitations (apparently only works with certain symsets, requires 24-bit color support). It would be good to support this option for all symsets and all types of color support, such as 16-bit color by accepting color names in addition to RGB triads.

But since I think MONSTERCOLOR as implemented in this pull request is likely to be redundant, I'll close this.

PR #438, opened 2021-01-05
Refactor getobj() to use callbacks on candidate objects

copperwater:
This replaces the arcane system previously used by getobj where the caller would pass in a "string" whose characters were object class numbers, with the first up to four characters being special constants that effectively acted as flags and had to be in a certain order.  Because there are many places where getobj must behave more granularly than just object class filtering, this was supplemented by over a hundred lines enumerating all these special cases and "ugly checks", as well as other ugly code spread around in getobj callers that formatted the "string".

Now, getobj callers pass in a callback which will return one of five possible values for any given object in the player's inventory. The logic of determining the eligibility of a given object is handled in the caller, which greatly simplifies the code and makes it clearer to read.  Particularly since there's no real need to cram everything into one if statement.

This is related to pull request #77 by FIQ; it's largely a reimplementation of its callbacks system, without doing a bigger than necessary refactor of getobj or adding the ability to select a floor/trap/dungeon feature with getobj. Differences in implementation are mostly minor:

- using enum constants for returns instead of magic numbers
- 5 possible return values for callbacks instead of 3, due to trying to make it behave exactly as it did previously. PR #77 would sometimes outright exclude objects because it lacked semantics for invalid objects that should be selectable anyway, or give slightly different messages.
- passing a bitmask of flags to getobj rather than booleans (easier to add more flags later - such as FIQ's "allow floor features" flag, if that becomes desirable)
- renaming some of getobj's variables to clearer versions
- naming all callbacks consistently with "_ok"
- generally more comments explaining things

The callbacks use the same logic from getobj_obj_exclude, getobj_obj_exclude_too and getobj_obj_acceptable_unlisted (and in a few cases, from special cases still within getobj). In a number of them, I added comments suggesting possible further refinements to what is and isn't eligible (e.g. should a bullwhip really be presented as a candidate for readying a thrown weapon?)

This also removed ALLOW_COUNT and ALLOW_NONE, relics of the old system, and moved ALLOW_ALL's definition into detect.c which is the only place it's used now (unrelated to getobj). The ALLOW_ALL functionality still exists as the GETOBJ_PROMPT flag, because its main use is to force getobj to prompt for input even if nothing is valid.)

I did not refactor ggetobj() as part of this change.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -1011,7 +1011,8 @@ E boolean NDECL(wearing_armor);
 E boolean FDECL(is_worn, (struct obj *));
 E struct obj *FDECL(g_at, (int, int));
 E boolean FDECL(splittable, (struct obj *));
-E struct obj *FDECL(getobj, (const char *, const char *));
+E int FDECL(any_obj_ok, (struct obj *));
+E struct obj *FDECL(getobj, (const char *, int (*)(OBJ_P), unsigned int));
 E int FDECL(ggetobj, (const char *, int (*)(OBJ_P), int,
                       BOOLEAN_P, unsigned *));
 E int FDECL(askchain, (struct obj **, const char *, int, int (*)(OBJ_P),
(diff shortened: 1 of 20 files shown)


copperwater:
Updated with some feedback from FIQ:

- unsqueak_ok() wasn't correctly downplaying all other potions (which would include informally identified oil)
- adjust_ok() has a side effect of calling impossible, but I consider this an acceptable side effect because it will never happen during correct execution of the game, so I updated the comment about side effects accordingly.

copperwater:
Squashed the commits and rebased on top of the current master branch from @paxed's request.

Issue #441, opened 2021-01-08
Bug in wear menu


Meklon2007:
Hello, 

1. My inventory:
![1](https://user-images.githubusercontent.com/12417354/104011839-d2aba600-51bf-11eb-9d51-55c4bcb4b70a.png)
2. Then i press W for wear menu:
![2](https://user-images.githubusercontent.com/12417354/104011960-01c21780-51c0-11eb-98ca-039e849bf835.png)
3. Then i take a blindfold and press W for wear menu:
![3](https://user-images.githubusercontent.com/12417354/104012067-3b931e00-51c0-11eb-9737-cd389f18dc62.png)

looks like a bug, because i see many items. Version of 3.7.0.30 (being developed).

P.S. I've got one more question. In the console version for windows of the game often displays item menus (for example, when reading an identity scroll). I can scroll down the list of items, but can I scroll up?

entrez:
You have `force_invmenu` on, right?

This is caused by the recent changes to getobj, specifically here:
https://github.com/NetHack/NetHack/blob/2f76d0e7014bd1dd4656155a49672d778d7d2cd5/src/invent.c#L1600-L1601

`forceprompt` shouldn't be used here I think; the rule is really supposed to be "set `ilet` to `'*'` only if nothing would show up when `'?'` is used". Before the `getobj` refactor the ternary expression was `*let ? '?' : '*'` and the equivalent variable to `forceprompt`, `allowall`, wasn't involved.

It seems to me that changing this to something like
```c
if (!oneloop) 
    ilet = (*lets || *altlets) ? '?' : '*'; 
```
might restore the previous behavior; it seems to do so for the case mentioned in the issue at least.

Also, '<' should scroll back up a page -- not sure if that's universal to every platform and windowport though...

Meklon2007:
yes, force_invmenu is on in the options.

P.S. thanks - yes '<' was worked.

copperwater:
@entrez, I'm not sure that *let is the same as (*lets || *altlets). The semantics of *let are, basically, whether any concrete object class was passed to getobj - e.g. `{ALLOW_ALL, FOOD_CLASS, TOOL_CLASS, 0}` would have *let be nonzero (FOOD_CLASS, because preprocessing used to advance it past up to 4 special constants including ALLOW_ALL), whereas `{ALLOW_ALL, 0}` would have *let be 0. The semantics of "getobj treats at least one concrete non-gold object class as generally acceptable" aren't there anymore.

[...]

PR #442, opened 2021-01-08
force_invmenu getobj listings

entrez:
Instead of listing the entire inventory when `force_invmenu` is on and `getobj` flags include `GETOBJ_PROMPT` (as described in #441), only applicable objects will be printed as long as any exist (either suggested or alternative choices).

However, a `getobj` callback that returns `GETOBJ_SUGGEST` for every item in inventory (such as `any_obj_ok`, used for the <kbd>d</kbd>rop command among other things) would produce a list of allowable objects that is in effect identical to the entire player inventory. Because of the way `display_pickinv` works, when this list is passed to it as a list instead of explicitly as the entire inventory, a useless "* - (list everything)" item is appended to it.  In order to fix this behavior, this patch would also add a test before adding the '*' item to ensure it will not be included for lists that already encompass the entire inventory.

There's more discussion of the ways `(*lets || *altlets)` is and isn't comparable to `*let` in pre-refactoring `getobj` in #441.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -1598,7 +1598,7 @@ unsigned int ctrlflags;
         else if (iflags.force_invmenu) {
             /* don't overwrite a possible quitchars */
             if (!oneloop)
-                ilet = forceprompt ? '*' : '?';
+                ilet = (*lets || *altlets) ? '?' : '*';
             if (!msggiven)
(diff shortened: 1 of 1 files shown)


entrez:
Fix incorporated into 81a1b9b

Issue #443, opened 2021-01-11
X11 status_hilite Not supported


theexternvoid:
The new status_hilite rules aren't supported in X11.

pat-rankin:
Issue #443 and pull request #444 were closed by
commits acbf423734a3b09a47c2c7329cf18404234016c8
and a2a9cb7b4ffd438ba8b5e2a9e3eb7b8b1021b40e
on Jan 26 2021.

PR #444, opened 2021-01-11
Fix Issue 443 - X11 Support for status_hilite Rules

theexternvoid:
For the X11 UI, I implemented the most important status_hilite rules: colors and invert. Bold, blinking, etc. are still not supported in this update. Also makes X11 honor the number specified in the statushilites config option.


paxed:
Setting `NetHack*fancy_status: False` in NetHack.ad causes a segfault.

theexternvoid:
> Setting `NetHack*fancy_status: False` in NetHack.ad causes a segfault.

It works fine for me. Can you post a stack trace?

paxed:
> > Setting `NetHack*fancy_status: False` in NetHack.ad causes a segfault.
> 
> It works fine for me. Can you post a stack trace?

I can do that, if necessary, but the problem is that you changed the color XtNfoo defines in xwindow.h to actual colors, when they're used as names for Xresources in init_menu_nhcolors, not as actual colors. NetHack.Ad has `NetHack*map*black: grey15`, but after your change, the code is trying to look up `NetHack*map*grey15` instead ...

theexternvoid:
Thanks! I thought those were unused so repurposed them. Sounds easy to fix. I'll get on it when I have some spare time, probably not this week unfortunately.

-David Mooter.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

theexternvoid:
I addressed the above concerns by restoring the XtNfoo defines in xwindow.h and making my code use its own hard coded values. It'd be nice if it pulled colors from NetHack.ad via get_nhcolor(). The challenge is that winstat.c is supporting two flavors: tty and fancy. The tty code seems to be using NetHack.ad via get_nhcolor(). The fancy code doesn't, hence doesn't load the necessary data for get_nhcolor() to work without more changes. My thought for long term is it may be easier to manage by splitting the tty and fancy flavors of winstat.c into two separate files so it's clear which is what, then update the fancy code to load colors via get_nhcolor() or something similar that pulls color codes from the resource file. But this is my first time in this code, so I may be wrong due to unfamiliarity.

pat-rankin:
Issue #443 and pull request #444 were closed by
commits acbf423734a3b09a47c2c7329cf18404234016c8
and a2a9cb7b4ffd438ba8b5e2a9e3eb7b8b1021b40e
on Jan 26 2021.

PR #450, opened 2021-01-18
fix issue 449 - dishwasher gender should be compatible with hero

mlehotay:
assuming the hero is straight

see [issue 368](https://github.com/NetHack/NetHack/issues/368) for related discussion


mlehotay:
@nhcopier fixed this in a8f4a54

PR #451, opened 2021-01-22
Scrap xlogfile flags field, replacing it with "mode"

copperwater:
As of the recent commit introducing a "bones" field in the xlogfile,
there is no more particular need for the 0x4 bit in flags. Its semantics
of "if flags & 0x4 is set, then this game did NOT load any bones" were
confusing and unintuitive anyway.

Then, with bones now out of the flags field, the only thing that flags
is used for is marking games as being in wizard mode or explore mode.
Despite the flags encoding function treating them as potentially
independent variables, they seem to be mutually exclusive to me: the
game doesn't allow you to start in both wizard and explore mode, and if
you do #exploremode while in wizard mode you get a warning that you are
leaving wizard mode.

Thus, this commit gets rid of flags entirely, replacing it with a
"mode=" field in the xlogfile. Possible values are "wizard", "explore",
and "normal".

### DIFF
--- a/src/topten.c
+++ b/src/topten.c
@@ -63,7 +63,7 @@ static void FDECL(readentry, (FILE *, struct toptenentry *));
 static void FDECL(writeentry, (FILE *, struct toptenentry *));
 #ifdef XLOGFILE
 static void FDECL(writexlentry, (FILE *, struct toptenentry *, int));
-static long NDECL(encodexlogflags);
+static const char *NDECL(encodegamemode);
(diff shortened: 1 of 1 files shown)


copperwater:
Withdrawing this because devteam members have objected to breaking xlogfile backwards compatibility by removing the flags field. (My opinion remains that the xlogfile has never been constant from major version to major version anyway, and the game ossifies faster the more things are in it that aren't allowed to be changed, but it looks like there's no will on the devteam's part to take this).

PR #453, opened 2021-01-22
Allow custom increment amounts in #wizintrinsic

entrez:
Let users of `#wizintrinsic` set a specific amount by which they want to increment each particular timeout, by entering a count/typing in numerals before selecting the item in the list.  I think doing it this way should be pretty intuitive to players who are familiar with the 'enter count -> make selection' flow of normal inventory menus in NetHack, and is simpler than using a separate prompt, as was mentioned as a possibility in a comment.

When an intrinsic is selected without entering a count, the increment will continue to default to the previous value of 30; this value is set via the `DEFAULT_TIMEOUT_INCR` macro.

I also deleted a redundant declaration of `wiz_intrinsic`; the duplicate seems to have been added by mistake in ff6139c6c5.

(Edit: merged via 77d4b4c955a9256129b601f0af70033821c3c502)

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -130,7 +130,6 @@ static int NDECL(dotravel);
 static int NDECL(doterrain);
 static int NDECL(wiz_wish);
 static int NDECL(wiz_identify);
-static int NDECL(wiz_intrinsic);
 static int NDECL(wiz_map);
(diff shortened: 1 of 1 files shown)


paxed:
Good idea. Pushed, and will show up in a bit.

PR #454, opened 2021-01-23
Allow running over water with IDed water walking

entrez:
Currently, only levitation or flight allow a character to run and use the travel command over water or lava.  The water walking extrinsic doesn't provide the same ability, for fear that it would give players a risk-free way to test whether something provided water walking: wear the item, run at some water, and see whether you're stopped at the edge or not:

https://github.com/NetHack/NetHack/blob/96ee175ce89ab71aa038c33d7d96c538d3debf8e/src/hack.c#L2897-L2900

This risk can be defused, though, by requiring not just water walking, but water walking provided by a known source (i.e. already-identified water walking boots).  Allowing heroes with an identified source of water walking to travel/run across the water doesn't provide the same opportunity for risk-free identification, and can make it a lot more convenient to traverse watery levels.

Identified water walking boots still wouldn't allow running over lava, even if they are known to be fireproof.

This would also add transition points between water/lava and land as "interesting" features that will pause running started by <kbd>G</kbd><kbd>k</kbd> or <kbd>g</kbd><kbd>k</kbd> (though not running by <kbd>Shift</kbd><kbd>k</kbd>), regardless of whether the hero can traverse the terrain on the other side of the transition.  This would allow people with water walking boots, a ring of levitation, etc, to use <kbd>G</kbd><kbd>k</kbd> to run within an island on Medusa's level, say, without continuing out into open water by accident.

I also changed the phrasing of the `flags.mention_walls` messages associated with stopping short of one of these transition points, since the previous phrasing of "you stop at the edge of the %s" wasn't totally clear about which side of the edge you were on -- since now these transitions between water and land can be approached from either direction, I tried to make the messages clearer about where exactly the character stopped.


entrez:
Since this has gotten hopelessly messed up and conflicted, I rewrote it from scratch in #802 -- that doesn't include some of the additional stuff in this PR like stopping at water-to-land transitions in certain run modes or modifications to the `mention_walls` messages, just sticks to the core "allow running and travel over water if you know that you have water walking" change.

PR #455, opened 2021-01-25
Unify code for extracting an object from a monster's inventory

copperwater:
The code for doing this (basically an obj_extract_self() call plus
handling if the object was worn or wielded) was duplicated all over, and
inconsistent - for instance, though all of them updated the monster's
misc_worn_check to indicate it was no longer wearing something in
whatever slot, only one call also set the bit that flags the monster to
consider putting on other gear afterwards.

Under a new function, extract_from_minvent, all this extra handling is
checked in one function, which can simply replace the obj_extract_self
call.

A few callers (such as stealing) have some common code *after* the
object is extracted and some other things happen such as message
printing, such as calling mselftouch if the object was worn
gloves. extract_from_minvent does not handle these cases.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -3101,6 +3101,8 @@ extern void bypass_objlist(struct obj *, boolean);
 extern struct obj *nxt_unbypassed_obj(struct obj *);
 extern struct obj *nxt_unbypassed_loot(Loot *, struct obj *);
 extern int racial_exception(struct monst *, struct obj *);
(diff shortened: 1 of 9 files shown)


nhmall:
[update-pr455.diff.txt](https://github.com/NetHack/NetHack/files/5887871/update-pr455.diff.txt)

copperwater:
Rebased and updated.

PR #458, opened 2021-02-05
Remove chaotic stormbringer peaceful penalty

RojjaCebolla:
Characters of all alignments, including Knights, can already remove shopkeepers
    and guards without penalty. From the wiki:
    "The simplest way to get rid of a peaceful human while avoiding the penalty for
    murder is to let your pet do the dirty work. You can also polymorph your victims
    before killing them. Note that a failed polymorph, " shudders." (system shock),
    still counts as murder."
    
    When a player is chosen to "steal souls" for the glory of whomever is concerned,
    it's a pretty big flavor glitch for stormbringer to start blasting the user for
    successfully snuffing out a soul with the artifact weapon that encourages murder
    (Not to mention the weird fantasy racism in which elf, orc, & dwarf deaths don't
    cause nearly the same level of punishment as human deaths do.)
    
    And in general, it's weird for chaotics to be god-penalized for any "crime".
    
    So the proposed change is that chaotic stormbringer users don't lose huge chunks
    of alignment for hitting peacefuls. I did not mess with the luck penalties, as
    those do not cause artifact blasting. While i find them unnecessary, they don't
    seem like a significant flavor glitch.

### DIFF
--- a/src/mon.c
+++ b/src/mon.c
@@ -2935,7 +2935,10 @@ xkilled(
             You_hear("the rumble of distant thunder...");
         else
             You_hear("the studio audience applaud!");
-    } else if (mtmp->mpeaceful)
+    } else if (mtmp->mpeaceful
+               && !(wep && wep->oartifact == ART_STORMBRINGER
+               && u.ualign.type == CHAOTIC))
+               /* it's not murder if stormbringer made you do it! */
         adjalign(-5);
 
     /* malign was already adjusted for u.ualign.type and randomization */


src/mon.c
```diff
@@ -2928,9 +2928,12 @@ xkilled(
             u.ublessed = 0;
         if (mdat->maligntyp == A_NONE)
             adjalign((int) (ALIGNLIM / 4)); /* BIG bonus */
-    } else if (mtmp->mtame) {
+    } else if ((mtmp->mtame) 
+               && (wep && wep->oartifact != ART_STORMBRINGER)
+               && (u.ualign.type != CHAOTIC)) {
```
entrez: This will evaluate to `FALSE` as long as the character is chaotic, even if they aren't wielding Stormbringer and haven't been crowned. It will also evaluate to `FALSE` if the character isn't wielding a weapon, regardless of their alignment, since `(wep && wep->oartifact != ART_STORMBRINGER)` is always false if `wep` is 0.

If the goal is not to punish the character as long as they have been "chosen to steal souls" maybe this would work better:
```c
} else if (mtmp->mtame && u.uevent.uhand_of_elbereth != 3) {
```
Or if you want explicitly to exclude Stormbringer:
```c
} else if (mtmp->mtame && !(wep && wep->oartifact == ART_STORMBRINGER)) {
```
You could also combine the two in order to specifically remove the penalty for crowned chaotics wielding Stormbringer:
```c
} else if (mtmp->mtame && !(wep && wep->oartifact == ART_STORMBRINGER
                            && u.uevent.uhand_of_elbereth == 3)) {
```

Edit: Actually, I guess you still have to check the current alignment in addition to `uhand_of_elbereth` in case the character has changed alignment since being crowned?
```c
} else if (mtmp->mtame && !(wep && wep->oartifact == ART_STORMBRINGER
                            && u.uevent.uhand_of_elbereth == 3
                            && u.ualign.type == CHAOTIC)) {
```

dnadlinger: Isn't the logic backwards here? `mtmp->mtame && !(wep && wep->oartifact == ART_STORMBRINGER && u.ualign.type == CHAOTIC` seems to be the intended condition.

src/mon.c
```diff
@@ -2896,7 +2896,7 @@ xkilled(
         if (Blind && !Blind_telepat)
             see_monsters(); /* Can't sense monsters any more. */
     }
-    if ((mtmp->mpeaceful && !rn2(2)) || mtmp->mtame)
+    if (mtmp->mpeaceful && !rn2(2)) || mtmp->mtame)
```
dnadlinger: Spurious change.

dnadlinger (changes requested):

src/mon.c
entrez: Also, this part only handles killing your own pet, not attacking shopkeepers and guards.  The `else if` right below it is the one that deals with peaceful non-pets.

https://github.com/NetHack/NetHack/blob/3db658de21e9694145413d6728674b8814ce6735/src/mon.c#L2941-L2942

entrez:
Another potential issue is that even when the logic is corrected, it is still just checking your currently wielded weapon -- not necessarily what killed the monster. So zapping a wand of death may or may not have negative consequences depending on whether you happen to be wielding Stormbringer, which seems pretty arbitrary.

Maybe Stormbringer should just be left out of it entirely, since the "chosen to steal souls" line doesn't explicitly say "with Stormbringer" -- I think it'd be more consistent to just exclude all chaotics who have been crowned Glory of Arioch, regardless of their current weapon.

src/mon.c
RojjaCebolla: whups. thanks!

RojjaCebolla:
Thanks for the corrections, @entrez and @dnadlinger ! 

To be more specific about intent, i updated the commit message. But in short, if i were making a NH fork, i would throw out all penalties for crime when the player is chaotic.  I figured a more conservative approach would be welcome in NetHack - if you're chaotic, you're generally encouraged to be more criminal, and if you're also wielding Stormy, you're being *strongly* encouraged to kill indiscriminately.

It's being blasted ***by Stormy*** for doing "#StormyThings" that feels most counterintuitive. Being crowned is simply the most common scenario in which chaotics receive stormbringer.

PR #463, opened 2021-02-19
Make "joined" in Lua and level structs always take a boolean

copperwater:
It is only ever used as a boolean, and having joined=BOOL for regions and joined=INT for rooms is confusing. So this PR changes all types of joined to booleans:

- g.coder->lvl_is_joined
- joined in struct room
- needjoining in struct mkroom
- and all places that did joined=INT in lua files are changed.


paxed:
Pushed.

PR #467, opened 2021-02-27
Two getobj fixes

### LINKED ISSUE #441: Bug in wear menu

copperwater:
Discovered a bug in my refactor of getobj a couple of months ago in which the sub-GETOBJ_SUGGEST cases didn't work for selecting '-', even though they are intended to make it selectable. This, for instance, makes it impossible to select your hands at a getobj prompt if the callback returns GETOBJ_DOWNPLAY - valid but not presented outright as an option.

I also pulled qt's fix of #441 into this pull request, at his request, because why not wrap two getobj fixes together?

Fixes #441 and #442.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -1649,7 +1649,7 @@ getobj(const char *word,
             /* guard against the [hypothetical] chace of having more
                than one invent slot of gold and picking the non-'$' one */
             || (otmp && otmp->oclass == COIN_CLASS)) {
-            if (obj_ok(otmp) <= GETOBJ_EXCLUDE) {
+            if (otmp && obj_ok(otmp) <= GETOBJ_EXCLUDE) {
                 You("cannot %s gold.", word);
                 return (struct obj *) 0;
             }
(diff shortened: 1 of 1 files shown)


src/invent.c
```diff
@@ -1479,10 +1479,17 @@ getobj(register const char *word,
...
     /* is "hands"/"self" a valid thing to do this action on? */
-    if ((*obj_ok)((struct obj *) 0) == GETOBJ_SUGGEST) {
-	allownone = TRUE;
+    switch ((*obj_ok)((struct obj *) 0)) {
+    case GETOBJ_SUGGEST:
         *bp++ = HANDS_SYM;
         *bp++ = ' '; /* put a space after the '-' in the prompt */
+        /* FALLTHRU */
+    case GETOBJ_DOWNPLAY:
+	allownone = TRUE;
```
entrez: yall need expandtab

PR #494, opened 2021-04-19
Fix: encyclopedia typo of "Galadrim"

copperwater:
Should be "Galadhrim", with an h.

### DIFF
--- a/dat/data.base
+++ b/dat/data.base
@@ -1558,7 +1558,7 @@ elven cloak
 	The Elves next unwrapped and gave to each of the Company the
 	clothes they had brought.  For each they had provided a hood
 	and cloak, made according to his size, of the light but warm
-	silken stuff that the Galadrim wove.  It was hard to say of
+	silken stuff that the Galadhrim wove.  It was hard to say of
 	what colour they were: grey with the hue of twilight under
 	the trees they seemed to be; and yet if they were moved, or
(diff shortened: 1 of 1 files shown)


pat-rankin:
I didn't add that quote but my copy of the Lord of the Rings spells it the way data.base already has it:  "Galadrim" (without 'h'), both in the "Farewell to Lorien" chapter in Volume I/Book II and in the index at the end of Volume III after Book VI and various appendices.

It's a boxed set of oversized softcovers from Ballantine Books in the mid 1970s, so nowhere close to the original but definitely not some cheap knockoff.

The index of names in the Silmarillion doesn't have that word at all.  The Tolkien Companion by J.E.A. Tyler has it with the 'h', but that's hardly canonical.

amateurhourrrr:
According to http://tolkiengateway.net/wiki/Galadhrim#Etymology

> In the first edition of The Lord of the Rings, Galadhrim was spelt Galadrim. In later editions, this was emended to the former, which Tolkien had decided was the correct Sindarin form.[10]

With a reference to The Lord of the Rings: A Reader's Companion, p. 305

Do you have a copy of that to check the reference @pat-rankin ?

nhmall:
Since the spelling currently present in the source tree matches the spelling in at least one edition of the book, let the current spelling stand for now.

PR #501, opened 2021-05-05
Fix: uninitialized buffer in mhitm theft feedback

entrez:
If a monster with a theft attack (nymph or leprechaun) stole something from an invisible monster (e.g. while under the influence of conflict), and the attacking monster was not itself invisible, the monster name buffer used when printing the "\<foo\> suddenly disappears!" message would be used while still uninitialized.  The attacking monster's name was only copied into the buffer if the defending monster was visible, but would be used regardless to print the pline if the attacking monster was visible and teleported away successfully after the attack.

uninitialized-buffer-use


paxed:
Committed

PR #502, opened 2021-05-05
Fix: nymph theft vs monster

entrez:
Nymphs' item theft attack against other monsters was broken in 1696019,
when a `break` used to select a particular item in the target monster's
inventory was changed to an early return.


paxed:
Committed.

Issue #510, opened 2021-05-11
Vomiting on the altar does not anger the Priest


davidssmith:
I was eating my lunch on a cross-aligned altar during a break in an arduous BUC testing session.

An expired tin of dog meat made me predictably sick and confused and eventually I wolfed all over the altar. I quietly slinked out, and the priest did nothing. 

Seems like soiling or defiling the altar should anger the Priest.

NullCGT:
[Thank you for the inspiration.](https://github.com/NetHack/NetHack/pull/519)

PR #519, opened 2021-05-25
Additional vomit effects.

NullCGT:
As @davidssmith mentioned in Issue #510 , vomiting on an altar currently has no effect. Given that the simple act of kicking or sitting on an altar provokes the direct intervention and wrath of a deity, this seems like an oversight. Of course, if we're adding some internal consistency to vomit, we might as well go all the way. This patch makes the following changes:

1. Vomiting on an altar provokes the wrath of a deity.
2. If the player is an acidic monster, vomiting on ice will cause the ice to melt, since they have very strong stomach acid (strong enough to reverse stoning).
3. If the player is a monster with an acidic breath weapon (i.e. a yellow dragon) they are treated as if they have used their breath weapon on themself.

This is, quite possibly, the worst thing I have ever written. I'm going to bed.

### DIFF
--- a/src/eat.c
+++ b/src/eat.c
@@ -3375,6 +3375,15 @@ vomit(void) /* A good idea from David Neves */
            reaches 0, but only if u.uhs < FAINTING (and !cantvomit()) */
         if (u.uhs >= FAINTING)
             Your("%s heaves convulsively!", body_part(STOMACH));
+         /* Vomiting on an altar is, all things considered, rather impolite. */
+        if (IS_ALTAR(levl[u.ux][u.uy].typ)) {
+            altar_wrath(u.ux, u.uy);
+        }
+        if (acidic(g.youmonst.data)) {
+            if (is_ice(u.ux, u.uy)) {
+                melt_ice(u.ux, u.uy, "Your stomach acid melts straight through the ice!");
+            }
+        }
     }
 
     /* nomul()/You_can_move_again used to be unconditional, which was
--- a/src/eat.c
+++ b/src/eat.c
@@ -3361,6 +3361,8 @@ floorfood(const char *verb,
 void
 vomit(void) /* A good idea from David Neves */
 {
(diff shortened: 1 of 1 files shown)


davidssmith:
This is awesome. I can't wait to see the ensuing hilarious YASDs.

src/eat.c
```diff
@@ -3373,6 +3375,15 @@ vomit(void) /* A good idea from David Neves */
...
             Your("%s heaves convulsively!", body_part(STOMACH));
+         /* Vomiting on an altar is, all things considered, rather impolite. */
+        if (IS_ALTAR(levl[u.ux][u.uy].typ)) {
+            altar_wrath(u.ux, u.uy);
+        }
+        if (acidic(g.youmonst.data)) {
+            if (is_ice(u.ux, u.uy)) {
+                melt_ice(u.ux, u.uy, "Your stomach acid melts straight through the ice!");
+            }
+        }
```
copperwater, line 3386: This part doesn't make sense to me. Even if it's super powerful acid (which to my knowledge doesn't have incredibly melty effects on ice), you're not really adding enough thermal energy to the ice to turn it back into a deep pool of water.

NullCGT: That's a fair point; I was enticed by the idea of ice/vomit-based YASDs. I wonder if there's an alternate way this could be implemented.

Henry-Hart: > 
> 
> This part doesn't make sense to me. Even if it's super powerful acid (which to my knowledge doesn't have incredibly melty effects on ice), you're not really adding enough thermal energy to the ice to turn it back into a deep pool of water.

Interesting. Wouldn't the acid be hot after it left the player's stomach? That thermal energy would melt some of the ice, and the rest would be produced in the resulting reaction. Also, since the acid is so strong and there is a significant quantity of it, it COULD produce an explosion...

PR #521, opened 2021-05-28
Correct quest guardian, Valkyrie PM alignments

actual-nh:
Correct to neutral the alignments for Attendants (who can grow up into neutral Healer player monsters), Warriors (who can grow up into Valkyrie player monsters), and Valkyrie player monsters. The first is the most egregious problem; the second and third put them into correspondence with the Valkyrie role.


copperwater:
Agree with this, currently the valkyrie monster and guardians are chaotic, which makes no sense.

PR #541, opened 2021-07-01
Fix: itemized billing and perm_invent

entrez:
When buying multiple items from a shop via itemized billing with
perm_invent enabled, paying for any single item would immediately
conceal the prices of all the remaining unpaid items in the inventory
pane, even those that the player declined to purchase.  In other words,
if holding three items from a shop, paying for one of them via itemized
billing would cause the other two items to be listed in the perminvent
pane as though they had already been paid for as well.

Deactivate iflags.suppress_price before the call to update_inventory in
dopayobj(shk.c) so that the inventory pane's item names accurately
reflect which items are still unpaid.

### DIFF
--- a/src/shk.c
+++ b/src/shk.c
@@ -1621,6 +1621,7 @@ dopayobj(
                   ltmp, "");
     obj->quan = save_quan; /* restore original count */
     /* quan => amount just bought, save_quan => remaining unpaid count */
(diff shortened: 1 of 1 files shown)


entrez:
This issue was reported in copperwater/xNetHack#49 but applies to NetHack as well.

PR #543, opened 2021-07-03
implement realtime/wallclock statusline timer

aoeixsz4:
this pr implements the ui clock developed by @bhaak in such a way that it can update independently of user input

for tty it uses pselect(), for curses they have their own timeout functionality

i am unsure about the portability of this, only tested on Ubuntu 20.04 with Linux kernel 5.8.0-59-generic

### DIFF
--- a/win/tty/wintty.c
+++ b/win/tty/wintty.c
@@ -3677,27 +3677,28 @@ static const char *encvals[3][6] = {
     { "", "Brd",      "Strs",     "Strn",     "Ovtx",      "Ovld"       }
 };
 #define blPAD BL_FLUSH
-#define MAX_PER_ROW 15
+#define MAX_PER_ROW 16
 /* 2 or 3 status lines */
 static const enum statusfields
     twolineorder[3][MAX_PER_ROW] = {
     { BL_TITLE, BL_STR, BL_DX, BL_CO, BL_IN, BL_WI, BL_CH, BL_ALIGN,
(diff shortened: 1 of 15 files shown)


win/tty/wintty.c
```diff
@@ -3513,11 +3518,37 @@ tty_nhgetch(void)
         i = randomkey();
     } else {
 #ifdef UNIX
+#ifdef REALTIME_ON_BOTL
+        fd_set read_set;
```
moon-chilled: i think something like

```
struct pollfd pf;
pf.fd = 0;
pf.events = POLLIN;
for (;;) {
        if (poll(&pf, 1, 1000)) {
               tgetch stuff
        } else {
                update stats
```

win/tty/wintty.c
```diff
@@ -77,6 +77,11 @@ extern short glyph2tile[];
 #define HUPSKIP_RESULT(RES) /*empty*/
 #endif /* ?HANGUP_HANDLING */
 
+#ifdef REALTIME_ON_BOTL
+#include <sys/select.h>
```
moon-chilled: should be guarded by #ifdef UNIX or so too

aoeixsz4: ah right

win/tty/wintty.c
aoeixsz4: oooo u can give the poll thing a timer
got it

PR #544, opened 2021-07-05
Fire vortices shouldn't extinguish flaming objects in hero's inventory

copperwater:
Small change that just makes sense: if the thing whirling around you is
composed of fire that is burning you to a crisp, your candles, lamps,
and lit oil ought to stay burning instead of being snuffed out.


ostrosablin:
I suppose it should then light up burnable items even if they are snuffed out.

copperwater:
That already effectively happens; it's not automatic, but taking fire damage can light ignitables you're carrying.

PR #553, opened 2021-07-21
Fix: off-by-one error in movecmd

entrez:
The for loop which iterates through the list of movement keys in
movecmd(cmd.c) was updated in 5abf948116 to count down to 0 instead of
up to the end of the list.  This commit inadvertently introduced an
off-by-one error which started the loop one past the actual end of the
array.  On my system this made 'H' stop working as the 'run West' key.


paxed:
Committed.

PR #554, opened 2021-07-21
Fix: engraving with dry magic marker

entrez:
Engraving with a marker can use up remaining charges to write
semi-permanently in ink, or--if the marker has no remaining charges--can
draw in the dust as with a normal blunt item.  Handling in engrave after
a recent refactor (9c6a5fc423) didn't fully cover this second case, and
as a result using an empty marker to engrave would deduct a charge per
attempt and cause an impossible.  This could be exploited by repeatedly
engraving with the same empty marker to underflow the marker charges.


paxed:
Committed.

PR #556, opened 2021-07-22
termcap.c: Don't use hilites[CLR_BLACK] = NULL for ANSI_DEFAULT

heiner:
This didn't crash since xputs is defensive and checks if it got a NULL
string, but it also didn't display black entries (dragons, unicorns,
orcish arrows, ...) as black or otherwise different from gray entries.


paxed:
Merged.

PR #557, opened 2021-07-22
Fix: crash when appending to long engraving

entrez:
If an engraving was appended to repeatedly, it could eventually exceed
BUFSZ and cause a crash.  Add some checks to prevent this from happening
and inform the player when she runs out of space to engrave on a
particular square.


paxed:
Committed.

PR #561, opened 2021-07-31
Non-metallic gloves protect worn rings from AD_ELEC in destroy_one_item

splatpope:
Based on a suggestion by luxidream (and a bunch of other good people)

The rationale behind this is that a lot of people feel that the now properly working shocking sphere explosion is imbalanced, causing a lot of grief and bitterness due to the loss of important wands and rings it causes. Even other electrical attacks are now considered under a new, harrowing light.

This is the first and the least disruptive of a series of 3 patches I've come up with to address this apparent (and admittedly minor) imbalance, the rest will be submitted if the reception on this one is favorable.

This PR makes it so that rings worn under non-metallic gloves are protected from destruction by electrical attacks. Because, you know, metals conduct electricity and all.
Since the only metallic gloves are the Gauntlets of Power, which are almost always considered the best choice for any ascension-worthy character, this introduces a new layer of tactical choices (Do I want an easy 25 str or do I want my worn rings to be safe ?).

### DIFF
--- a/src/zap.c
+++ b/src/zap.c
@@ -5003,7 +5003,7 @@ destroy_one_item(struct obj *obj, int osym, int dmgtyp)
         switch (osym) {
         case RING_CLASS:
             if (obj->otyp == RIN_SHOCK_RESISTANCE ||
-                    (obj->owornmask && uarmg && !is_metallic(uarmg))) {
+                    (obj->owornmask & W_RING && uarmg && !is_metallic(uarmg))) {
                 skip++;
                 break;
             }
--- a/src/zap.c
+++ b/src/zap.c
@@ -5002,16 +5002,9 @@ destroy_one_item(struct obj *obj, int osym, int dmgtyp)
         quan = obj->quan;
         switch (osym) {
         case RING_CLASS:
-            if (obj->otyp == RIN_SHOCK_RESISTANCE) {
+            if (obj->otyp == RIN_SHOCK_RESISTANCE || (obj->owornmask && uarmg && !is_metallic(uarmg))) {
                 skip++;
                 break;
-            }
-            if (obj->owornmask && uarmg) {
-                if (obj->owornmask & W_RING && !is_metallic(uarmg)) { /* metallic gloves conduct electricity */
-                    pline("Your %s is protected by the %s you're wearing !", xname(obj), xname(uarmg)); /* for debug purposes */
-                    skip++;
-                    break;
-                }
             }
             dindx = 5;
             dmg = 0;
--- a/src/zap.c
+++ b/src/zap.c
@@ -5002,9 +5002,8 @@ destroy_one_item(struct obj *obj, int osym, int dmgtyp)
         quan = obj->quan;
         switch (osym) {
         case RING_CLASS:
-            if (obj->otyp == RIN_SHOCK_RESISTANCE ||
-                    ((obj->owornmask & W_RING) && 
-                        uarmg && !is_metallic(uarmg))) {
+            if (((obj->owornmask & W_RING) && uarmg && !is_metallic(uarmg)) ||
+                    obj->otyp == RIN_SHOCK_RESISTANCE) {
                 skip++;
                 break;
             }
--- a/src/zap.c
+++ b/src/zap.c
@@ -5006,6 +5006,13 @@ destroy_one_item(struct obj *obj, int osym, int dmgtyp)
                 skip++;
                 break;
             }
+            if (obj->owornmask && uarmg) {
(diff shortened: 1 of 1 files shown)


src/zap.c
```diff
@@ -5002,7 +5002,8 @@ destroy_one_item(struct obj *obj, int osym, int dmgtyp)
         quan = obj->quan;
         switch (osym) {
         case RING_CLASS:
-            if (obj->otyp == RIN_SHOCK_RESISTANCE) {
+            if (obj->otyp == RIN_SHOCK_RESISTANCE ||
+                    (obj->owornmask && uarmg && !is_metallic(uarmg))) {
```
entrez: `obj->owornmask` will also have a value for stuff like wielded objects, so this would protect rings you have quivered, wielded, or readied in your alternate weapon slot.  `(obj->owornmask & W_RING)` would check more specifically if the item is being worn in one of the ring slots.

pat-rankin:
The concise version has left out something from the first version.
The owornmask check needs to be '(obj->owornmask & W_RING) != 0'
otherwise someone could protect up to three extra rings by putting them
into the weapon, swap-weapon, and quiver slots (temporarily, if they have
enough time as a shocking monster approaches).

splatpope:
Fixed (good of you guys to explain me how owornmask worked, I didn't figure that rings being wielded would count and thought I could just omit that check !)

Nonetheless, I feel that this line is getting increasingly ugly for no other reason that limiting the declaration of local variables and staying under 80 columns. Do you feel like there would be a good point in defining a "ring_under_glove" function or macro ?

entrez:
If the glove thing came before the existing `otyp` check, it would fit into two lines:

```c
if (((obj->owornmask & W_RING) && uarmg && !is_metallic(uarmg))
    || obj->otyp == RIN_SHOCK_RESISTANCE) {
```

splatpope:
Should be good, hit me up if there's more stuff to fix or if you want me to squash the commits

copperwater:
> This is the first and the least disruptive of a series of 3 patches I've come up with to address this apparent (and admittedly minor) imbalance, the rest will be submitted if the reception on this one is favorable.

Do you mind sharing what the other 2 are?

splatpope:
> > This is the first and the least disruptive of a series of 3 patches I've come up with to address this apparent (and admittedly minor) imbalance, the rest will be submitted if the reception on this one is favorable.
> 
> Do you mind sharing what the other 2 are?

* adjust probabilities of wands exploding to 1 for cursed wands and 1/9 for blessed wands
* non-metallic wands and rings flat out resist electrical damage

FredrIQ:
Personally, the way I saw it when it comes to shock damage, I never considered wands blowing up a problem, because wands are consumables (Although there is an argument to be made that it should zap wand charges instead of the entire wand, unless a wand's charges become 0 as a result), but rings are equippables and thus shouldn't be destroyed that easily. Weapons can't be destroyed at all (except for ammo mulching) and armor can only be destroyed by "destroy armor" (resisted by MR) or disintegration (prevented by reflection). But even with shock resistance, rings still blow up.

This PR is a perfectly sensible solution for rings IMO. While I wouldn't mind the other 2 feature examples you brought up on a personal level, it's mostly because then I'd have less to worry about as a player, and not because I feel the current behaviour of those are unbalanced.

copperwater:
> non-metallic wands and rings flat out resist electrical damage

I'd note that this interferes with a NetHack design principle some people try to adhere to - the principle of minimizing the amount of things that:
- are determined at the start of the game
- are hidden to the player
- have an effect on the long term difficulty of the game

For instance, if I start a game in which the wand of death, wand of wishing, wand of teleportation, ring of free action, ring of conflict, ring of teleport control etc. etc. are all randomized to be metal, that game will have some difficulties that it wouldn't have if they all generated as nonmetal, and it's not really fair to the player because they had no idea what the items were randomized to.

NetHack already has some things that break this principle, so I don't think it's disqualifying of the idea, but it's an argument against adding more things like that. (Color alchemy is another proposed system along the same lines - some games will let you mix cheap common potions into powerful ones, and some just won't, and you don't learn that until well into the game.)

paxed:
Move the || to the start of the line, and squash the commits.

entrez:
The wraparound line is also indented too much. It should look like the comment I made earlier in terms of formatting:

```c
if (((obj->owornmask & W_RING) && uarmg && !is_metallic(uarmg))
    || obj->otyp == RIN_SHOCK_RESISTANCE) {
```

I tried to get a message to you on Discord (which I don't use, so I was probably unsuccessful) but there's a file in DEVEL/ called code_style.txt that describes a lot of this stuff.

paxed:
Committed.

PR #565, opened 2021-08-04
Amplify booze effect if drinking on empty stomach

entrez:
Ok, this is admittedly a sort of silly idea, but has [basis in reality](https://www.nytimes.com/2005/12/06/health/the-claim-never-drink-on-an-empty-stomach.html).

Drinking booze on an empty stomach will amplify its effects
(i.e. increase the duration of the resulting confusion); stuffing
yourself before drinking will have the opposite effect.


ostrosablin:
Sounds good, another little situation contributing to NetHack's TDTTOE experience.

PR #567, opened 2021-08-07
Display a permanent list of nearby characters

NullCGT:
## Overview

This pull request adds a window in the curses window port which displays
nearby characters on the map. Pressing a key (bound to ']' by default)
changes whether the window is displaying nearby monsters or nearby
objects.

This pull request is a proof of concept; I am very happy to receive and implement feedback.

## Rationale for Change

This is, first and foremost, an accessibility feature. Prospective NetHack
players are often discouraged by the interface, and the lack of explanation
of various characters and symbols. One way to combat this is by displaying
a permanent list of nearby glyphs. Displaying a list of nearby items and
monsters is a feature in many popular modern roguelikes, such as
Cataclsysm: Dark Days Ahead, Brogue, and Dungeon Crawl Stone Soup. In order
to increase its appeal to new players, NetHack should offer this as an
option as well.

Furthermore, this feature can benefit experienced players as well, allowing
them to determine at a glance whether a magenta h is a dwarf king or a
mind flayer. A nearby characters window should greatly speed up play, as well
as reduce some of the tedium of farlooking monsters in order to determine
their identities.

## Technical Notes

While much of the code for this feature is taken from the look_all code in
pager.c, there are a number of differences. First, coordinates are omitted
from the window. While displaying coordinates is useful, we risk overwhelming
new players by presenting them with too much information (something NetHack
is already quite good at). Second, we visually distinguish peaceful monsters
with a different bullet point. Third, pet attributes and color are displayed
in this window, so it is easier to read at a glance.

I made the decision to deviate from the standard set by cursinvt.c and avoid
saving the text in memory, since that would greatly complicate the code for
little benefit.

I chose to implement a toggle for the type of nearby characters rather than
a scroll button like the one used in the permanent inventory. In the future,
a scroll button could potentially be added.

## Considerations

There are a few considerations about this feature that are important to note.
The first is that updating the list of nearby characters is a potentially
expensive operation. While I do not believe this is of great concern, since
the window is only available in the curses window port, which is likely only
to be run on powerful computers, it is still important to note.

When the permanent inventory and permanent nearby windows are displayed at
the same time, the screen becomes rather cramped. I attempted to stack them
on top of one another, but was unable to determine how to do so.

Finally, I recognize that the existing update_nearby() calls are likely
insufficient, and there may be a number of edge cases where the window updates
late.

## Ports
* Curses: Enabled.
* Qt: Enabled.
* TTY: Functions are stubbed out so the port compiles. Strongly suggest not implementing, since the look_all() functionality accessed with the '/' key essentially duplicates the function of this window, and adding a key to display this window would be an unnecessary duplication of code and functionality.
* Windows: Functions are stubbed out so the port is playable. Not implemented.
* Shim: Functions are stubbed out so the port is playable. Not implemented.
* X11: Functions are stubbed out so the port is playable. Not implemented.

Note that stubbed out functions are written in the manner of perm_invent functions in window ports where that option is not enabled.

## Conclusion

That was a lot of words! Feedback is more than welcome.


paxed:
What about the Windows, X11, and Qt windowports? TTY obviously can't permanently display the window, but there should have a command to show it and wait for user.

NullCGT:
@paxed I was planning on porting it to X11 and Qt. (I don't have a way of testing a windows port at the moment). I wanted to get this PR out here first, in order to see if there was interest, and if the implementation was sound.

EDIT: I'll be doing my best to port this to available platforms.

NullCGT:
@paxed After running into numerous headaches, I'm hesitant to port this to Qt and X11 for the following reasons:
* The permanent inventory option causes a softlock followed by a segfault in X11 (tested on latest version of 3.7-Dev, not noted in X11-issues).
* In Qt, the inventory display code already uses a major "hack" in order to work.
While this feature isn't the same, it's similar enough that I'm concerned about touching those ports. I'm interested in porting this to tty and Windows, but I don't think Qt and X11 are ready to receive new features.

EDIT: I've got Qt working, but the X11 port does not seem worth taking a look at until it is fixed.

entrez:
> TTY obviously can't permanently display the window, but there should have a command to show it and wait for user.

Wouldn't it be a bit redundant to have this as well as the existing "list nearby objects" and "list nearby monsters" functionality under <kbd>/</kbd>?  The big difference/advantage is that the list in the patch is shown in a persistent window; if that's gone, it doesn't really seem strongly differentiated from the existing options to me.

NullCGT:
Funnily enough, I updated the description to say the same thing at almost the exact moment you commented that, entrez. That's where my feelings on this lie as well, now that I've done more digging into things.

ostrosablin:
Interesting. Thank you, entrez, today I've learned that under `/` there's similar monster/object (non-permanent) listing functionality already built in. Will be very useful in crowded situations. NetHack is certainly a game that you're constantly learning and mastering for years. And even paxed seems to be unaware that such feature already exist. NetHack will never stop to surprise me. Love such "wow, so it WAS possible?" moments.

pat-rankin:
tty could display them below the status line on terminals that have a high row count specified.  It would lose usefulness when there are too many lines of info, but the persistent inventory window already has that problem on other interfaces.

I wonder whether this feature is useful enough to justify the maintenance headache it will bring with it (across all interfaces, not just 'maybe tty').  'perm_invent' has needed a seemingly endless stream of fixes.

NullCGT:
I'm planning on releasing an update for my variant fairly soon. I could take this feature and throw it in the release in order to gauge the community's reaction and determine how useful it is.

ostrosablin:
@pat-rankin To be honest, perm_invent below status line in tty windowport would be (in my opinion) much more useful than list of nearby characters (as long as there's a way to scroll it like curses counterpart), and it would be a lifesaver for playing on mobile SSH client (because smartphones, even with on-screen keyboard usually still have a plenty of unused vertical screen space, while having a very limited horizontal screen space - this renders curses windowport unusable for smartphones, but perm_invent is just so useful).

shootmeinthehead:
your colors for the game is so good @NullCGT can i get them,  cuz the font and colors are really well matched, if necesary u can share your .nethackrc. Thx anyway. I have issue with the perm_invent not working even with the OPTIONS=perm_invent and the #perminv

PR #568, opened 2021-08-08
Make enchant armor prompt for target item

amateurhourrrr:
This prompts the player for a worn armor piece to enchant when reading a
scroll of enchant armor.

Although this is more flexible than before, the player must select a
worn piece of armor or opt out entirely - they cannot enchant something
that isn't currently equipped.

This should make it more ergonomic to enchant armor, without having
to derobe and shuffle armour pieces around

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -1029,6 +1029,19 @@ display_stinking_cloud_positions(int state)
     }
 }
 
+/* getobj callback for armor object to enchant */
+static int
+armor_enchant_ok(struct obj* obj)
+{
+    if (!obj)
+        return GETOBJ_EXCLUDE;
+
+    if (obj->owornmask & W_ARMOR)
+        return GETOBJ_SUGGEST;
+
+    return GETOBJ_EXCLUDE_INACCESS;
+}
+
 static void
 seffect_enchant_armor(struct obj **sobjp)
 {
@@ -1036,11 +1049,16 @@ seffect_enchant_armor(struct obj **sobjp)
     register schar s;
     boolean special_armor;
     boolean same_color;
-    struct obj *otmp = some_armor(&g.youmonst);
     boolean sblessed = sobj->blessed;
     boolean scursed = sobj->cursed;
     boolean confused = (Confusion != 0);
     boolean old_erodeproof, new_erodeproof;
+
+    struct obj *otmp = getobj("enchant", armor_enchant_ok, GETOBJ_NOFLAGS);
+    while (otmp && !(otmp->owornmask & W_ARMOR)) {
+        You("cannot target armor that is not worn.");
+        otmp = getobj("enchant", armor_enchant_ok, GETOBJ_NOFLAGS);
+    }
 
     if (!otmp) {
         strange_feeling(sobj, !Blind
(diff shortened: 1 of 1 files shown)


src/read.c
```diff
@@ -1029,19 +1029,37 @@ display_stinking_cloud_positions(int state)
...
     register schar s;
     boolean special_armor;
     boolean same_color;
-    struct obj *otmp = some_armor(&g.youmonst);
     boolean sblessed = sobj->blessed;
     boolean scursed = sobj->cursed;
     boolean confused = (Confusion != 0);
     boolean old_erodeproof, new_erodeproof;
 
+    struct obj *otmp = getobj("enchant", armor_enchant_ok, GETOBJ_NOFLAGS);
```
copperwater: Since this makes it a prompting scroll, I would add a "This is an enchant armor scroll." message here before the prompt, similar to charging or genocide scrolls, as well as learnscroll() for perminvent update prior to the getobj (then possibly remove a makeknown() call later on that previously did the job).

```c
    if (!objects[SCR_ENCHANT_ARMOR].oc_name_known) {
        pline("This is an enchant armor scroll.");
        learnscroll(sobj);
    }
    otmp = getobj(...)
```

src/read.c
```diff
@@ -1046,13 +1046,21 @@ static void
...
+    int otyp = sobj->otyp;
     register schar s;
     boolean special_armor;
     boolean same_color;
     boolean sblessed = sobj->blessed;
     boolean scursed = sobj->cursed;
     boolean confused = (Confusion != 0);
     boolean old_erodeproof, new_erodeproof;
+    boolean already_known = (sobj->oclass == SPBOOK_CLASS /* spell */
+                             || objects[otyp].oc_name_known);
```
copperwater, line 1061: For NetHack code style consistency, you're also going to want to move the `struct obj *otmp;` declaration up here separate from its initialization.

src/read.c
```diff
@@ -1029,18 +1029,45 @@ display_stinking_cloud_positions(int state)
...
+static int
+armor_enchant_ok(struct obj* obj)
+{
+    if (!obj)
+        return GETOBJ_EXCLUDE;
+
+    if (obj->owornmask & W_ARMOR)
+        return GETOBJ_SUGGEST;
+
+    return GETOBJ_EXCLUDE_INACCESS;
```
entrez: The way `GETOBJ_EXCLUDE_INACCESS` is used elsewhere, it implies that the action could apply to this item, were it not for some specific issue (e.g. wearing armor that's already being worn, dipping a ring that's being worn under gloves).  I think it may not be the best choice as the default case for the callback.

This also doesn't distinguish between valid armor that's just not currently being worn (in which case `GETOBJ_EXCLUDE_INACCESS` would make sense) and random items that are not armor at all -- both will return `GETOBJ_EXCLUDE_INACCESS` and so both will produce the "You cannot target armor that is not worn" message.

amateurhourrrr: There's a difference when using the other excludes in that it can return the zero object (all values set to null) rather than null itself, but good point on the message

entrez: Oh, is that different from `GETOBJ_EXCLUDE_SELECTABLE`?  The comment in hack.h says they behave exactly the same except for the message when nothing applies ("You have nothing else to act on" vs "You have nothing to act on").

amateurhourrrr: will check it out after work, I can't remember why I had to set it to that

PR #573, opened 2021-08-10
Fix: m-prefix movement into warning symbol

entrez:
Moving into a position containing a warning symbol with m-\<direction\> to
'safely' move would still attack as though the 'm' prefix was not
specified.  Ensure warning symbols are counted as 'detected' monsters
for this purpose, to avoid falling through to do_attack().


copperwater:
For context, the specific nethackwiki example that spawned this bug report: you'll get instapetrified by m-moving into a cockatrice next to you that you can see only with warning.

PR #578, opened 2021-08-20
Count all poly'ing tins as potential sliming cures

entrez:
Popeye, used to check whether eating a particular tin is a potential
lifesaving action, considered tins of chameleon meat a cure for sliming
(since you can polymorph into a fiery monster), but didn't credit the
same possible curative power to other tins which can polymorph the hero.

Use the polyfodder macro (used elsewhere to check whether eating
something will polymorph a monster/pet) to determine whether a tin will
polymorph the hero.

### DIFF
--- a/src/eat.c
+++ b/src/eat.c
@@ -3546,7 +3546,7 @@ Popeye(int threat)
                           && (mndx == PM_LIZARD || acidic(&mons[mndx])));
     /* polymorph into a fiery monster */
     case SLIMED:
-        return (boolean) (mndx == PM_CHAMELEON);
+        return (boolean) polyfodder(otin);
     /* no tins can cure these (yet?) */
     case SICK:
     case VOMITING:


src/eat.c
```diff
@@ -3546,7 +3546,8 @@ Popeye(int threat)
                           && (mndx == PM_LIZARD || acidic(&mons[mndx])));
     /* polymorph into a fiery monster */
     case SLIMED:
-        return (boolean) (mndx == PM_CHAMELEON);
+        return (boolean) (mndx == PM_CHAMELEON || mndx == PM_GENETIC_ENGINEER
```
copperwater: Perhaps instead of enumerating specific monsters in a duplicate of cpostfx, this should be moved to a function (which would also then be used in the polyfodder() macro used in dog_eat.c)? Yes, I know cpostfx enumerates them in a large switch statement, but I think it's worth it.

entrez: Yeah, the reason I just listed them out directly was because of the switch statement -- it slipped my mind that there were other places it might be used/useful.  I'll have to look at it again.

entrez: Maybe it would be best/simplest to just use `polyfodder(otin)` rather than creating a new function, unless there's some difference in practice between that and the list in `cpostfx`? Are there any corpses that will polymorph a monster but not the hero, or vice-versa?

copperwater: That sounds better actually, yeah. polyfodder appears to be defined in terms of 1) shapeshifters and 2) monsters with a polymorph attack, which should be this same set.

PR #579, opened 2021-08-22
Remove g.monstermoves

### LINKED ISSUE #288: Contradictory comments about moves and monstermoves
This implies that moves tracks player moves whereas monstermoves tracks monster moves (turns):
https://github.com/NetHack/NetHack/blob/67b485601a3a4b2b126a09114879643263765159/include/decl.h#L838

But in allmain.c, moves and monstermoves are incremented together, with a comment stating that monstermoves' increment is obsolete, and there never seems to be a separate increment of just moves:
https://github.com/NetHack/NetHack/blob/67b485601a3a4b2b126a09114879643263765159/src/allmain.c#L196-L197

so which is it? Do the variables ever diverge, and if not, why are they tracked separately?

copperwater:
It's redundant with g.moves, so there is no more need for it.

Way, way back, it looks like g.moves and g.monstermoves can and did
desync, where g.moves would track the amount of moves the player had
gotten (and would therefore increase faster if the player were hasted)
and g.monstermoves would track the amount of monster move cycles, aka
turns. But this has not been the case for a long time, and they both
increment together in the same location in allmain.c. There are no
longer any cases where they will not be the same value.

This is a save-breaking change because it changes struct
instance_globals, but I have not updated the editlevel in this commit.

Also, fixes #288 (the comment in decl.h is wrong).

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -834,8 +834,7 @@ struct instance_globals {
     struct mkroom rooms[(MAXNROFROOMS + 1) * 2];
     struct mkroom *subrooms;
     dlevel_t level; /* level map */
-    long moves;
-    long monstermoves; /* moves and monstermoves diverge when player is Fast */
+    long moves; /* turn counter */
     long wailmsg;
     struct obj *migrating_objs; /* objects moving to another dungeon level */    
     struct obj *billobjs; /* objects not yet paid for */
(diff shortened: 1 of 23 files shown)


pat-rankin:
The struct 'instance_globals' is not saved and restored.  However, individual variable 'g.monstermoves' is, so this will break saves, just not for the reason mentioned.

In the fairly distant past, monstermoves was used to catch up for lost time whenever the hero returned to a previously visited level and when migrating monsters eventually arrived at their destination.

copperwater:
This has been added.

PR #580, opened 2021-08-23
Remove requirement of object probs adding to 1000

copperwater:
When discussing the recent commit that removed makedefs -o from the
build process, nhmall pointed out that a sanity check ensuring all
objects within one class add up to 1000 probability had been removed as
well. This requirement was a perennial thorn in the side for anyone
doing anything that touches object probabilities, because allocating
probability to something meant deciding what to take it away from,
without a good way to evenly distribute that across all the other
members of the object class.

I had gotten around this in xNetHack by removing the sanity check and
making mkobj() total up the probability within an object class and then
using that instead of 1000. This commit takes a similar approach, but
instead of inefficiently recalculating the sum every time mkobj() is
called, it instead computes it at the start of the game and stores it in
a global variable.

This fixes a slight bias problem with rings - they are all supposed to
be of equal probability, but there are 28 of them and 1000 is not evenly
divisible by that, so the old formula made the later rings slightly more
likely. Now instead of a 35/1000 or 36/1000 chance, they are all
uniformly 1/28. (Internally they have a oc_prob of 1 now, not 0).

Gems are also weird, because their oc_prob values change every level.
This ought to have still worked without a change, because the arcane
formula for assigning the probabilities would still end up with them
adding to 1000. But I added in code to reset the total gem probability
anyway; this may help make the formula less arcane in the future.

There is still a sanity check against object classes having zero total
probability, in which case the game will not start. I also downgraded
the "probtype error" panic in mkobj() to an impossible because it has a
reasonable failure case - return the first item in that class.

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -1056,6 +1056,7 @@ struct instance_globals {
 
     /* o_init.c */
     short disco[NUM_OBJECTS];
+    short oclass_prob_totals[MAXOCLASSES];
 
     /* objname.c */
     /* distantname used by distant_name() to pass extra information to
--- a/include/extern.h
+++ b/include/extern.h
@@ -1739,6 +1739,7 @@ extern void consoletty_exit(void);
 /* ### o_init.c ### */
 
 extern void init_objects(void);
+extern void init_oclass_probs(void);
 extern void obj_shuffle_range(int, int *, int *);
 extern int find_skates(void);
 extern boolean objdescr_is(struct obj *, const char *);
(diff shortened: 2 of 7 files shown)


copperwater:
Marked draft because I thought everything in instance_globals was saved and restored, which isn't the case. The new field needs to be either added to the save file or regenerated upon restoring the game. Will take a look tomorrow.

copperwater:
Changed it so that the field is populated both at game start and at game restore.
Interestingly, if a game is saved, then the binary is recompiled with different oc_prob values, and then the saved game is restored, it will continue to use the original probabilities from before, because objects[] is written into the save file. I suppose the same thing happens currently, under the constraint that oc_prob must still total 1000, so this isn't really changing how that works.

paxed:
Has some compile warnings in o_init.c and restore.c, plz fix.

paxed:
Committed.

PR #583, opened 2021-09-02
Fix: replace_terrain used wrong y2 coordinate

copperwater:
It erroneously uses x2 again instead of the y2 value. It looks like in most cases where replace_terrain is specified with a region, the region goes to the bottom of the screen, and this effect wasn't noticed; but there are some cases like in Orcish Town where it might have been knocking down the wrong walls.


paxed:
Committed.

PR #598, opened 2021-09-30
Fix: baalz_fixup didn't account for level rotation 

entrez:
Because some spots in the fly's 'legs' on the Baalzebub lair level are
specified as pools in the level file, then later converted to walls in a
post-generation fixup routine, monsters can be generated on those spots
and then left walled up and inaccessible, Cask of Amontillado style.
For the love of God, makemon-tresor!

Some code already existed to relocate these monsters after generating
the level, but it depends on misorientated 'leg segments' being fixed up
in a particular way.  That wasn't being triggered because it didn't
account for the possible rotation of the level; as a result, the
monsters in the leg segments wouldn't be relocated, and the leg segments
themselves would continue to have the wrong orientation.

Account for possible level rotation so that the monsters are relocated
properly (and the leg segments are 'fixed').

### DIFF
--- a/src/mkmaze.c
+++ b/src/mkmaze.c
@@ -418,17 +418,18 @@ baalz_fixup(void)
        both top and bottom gets a bogus extra connection to room area,
        producing unwanted rectangles; change back to separated legs */
     x = g.bughack.delarea.x1, y = g.bughack.delarea.y1;
-    if (isok(x, y) && levl[x][y].typ == TLWALL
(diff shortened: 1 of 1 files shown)


entrez:
Okay, this was a little premature -- there's already code in there that's supposed to solve this problem, but it's not working...

I think I have it figured out now, though the fix is no longer specifically about monsters stuck in the walls.

PR #609, opened 2021-10-18
Have neutral sacrifices disappear in a puff, not a cloud, of smoke.

Vivit-R:
The other two messages, "flash of light" and "burst of flame", each have an evocative onomatopoetic noun describing a sharp, abrupt release of light and flame, respectively. If law is represented by light, chaos by flame, and balance by smoke, the natural word is _puff,_ not _cloud._

### DIFF
--- a/src/pray.c
+++ b/src/pray.c
@@ -1319,7 +1319,7 @@ consume_offering(struct obj *otmp)
              u.ualign.type == A_LAWFUL
                 ? "flash of light"
                 : u.ualign.type == A_NEUTRAL
-                    ? "cloud of smoke"
+                    ? "puff of smoke"
                     : "burst of flame");
     if (carried(otmp))
         useup(otmp);


Planet-Patrick:
Puff sounds a bit sissy; you're sacrificing an animal that you've killed for heaven's sake!

Vivit-R:
I think a less intense message is actually appropriate for the neutral alignment. Besides, its doesn't say how _big_ a puff of smoke it is.

copperwater:
You could also use "great puff of smoke". Though I agree with vivit that "puff of smoke" is fine.

entrez:
"Your sacrifice disappears in a big ol' dang puff of smoke"

copperwater:
> You could also use "great puff of smoke". Though I agree with vivit that "puff of smoke" is fine.

The irony here is that I completely forgot I was the one who added "cloud of smoke" in the first place, and decided against using "puff" like in SpliceHack.

https://github.com/nethack/nethack/commit/c578b95

nhmall:
Stuff on the floor can already cause you to see a "puff of smoke" from fire (zap.c), and when a djinni doesn't emerge from a magic lamp, you can see "a puff of smoke" (apply.c). Why have the language changed to be so similar for an altar?

entrez:
That's a good point.  I agree that "puff" matches more closely than "cloud" to the feel of the other alignments' sacrifice verbiage, but the distinction from the preexisting "puff of smoke" uses seems like a good thing.

Moult:
A gust of smoke? A whirlwind of smoke? A rush of smoke? A gale of smoke?

Vivit-R:
How "burst of smoke" for neutral and "_blast_ of flame" for chaotic?

PR #610, opened 2021-10-18
Have the '$' command include gold in your pack

Vivit-R:
In verbose mode, the gold in your wallet is totaled separately from that in containers in your pack, and the two are listed separately. In terse mode, just print the total of both. Only known gold is mentioned.


paxed:
Merged

PR #618, opened 2021-10-23
Make amnesia drain training to appropriate level

amateurhourrrr:
When amnesia drains your skills the skill training would be set
to a random amount rather than a random valid amount for the new
level of skill.

This meant that, for example, you could have Master skill level in
martial arts but with the training amount of Basic.

Attempts to retrain to level martial arts to Grand Master would
then take an extraordinary amount of time compared to usual.

Fix taken from Evilhack


paxed:
Merged.

PR #624, opened 2021-10-28
Add missing `const`

argrath:
If you want to declare a pointer which the address pointed to is constant,
you should declare it as like `static const char *const var = "...";`.

This commit supplies missing `const` and prevents some programming
error in the future.


nhmall:
I'm unable to re-open this pull request after the branch work someone did on Jan 27. The Reopen button is grayed out.

PR #626, opened 2021-10-30
Turn cancelled bags of holding into sacks

Vivit-R:
A wand or spell of cancellation zapped at a bag of holding now turns the bag into a sack. A wand of cancellation placed directly into a bag of holding now, instead of exploding the bag, just uses one charge and cancels the bag. This is not checked for recursively; you can put a wand of cancellation into a sack and then safely put the sack into a bag of holding.

The biggest consequence of this (other than making wands of cancellation less of a hassle to carry) has to do with looting cursed bags of holding from bones. Formerly, you could just cancel the bag and pick it up. You can still do this to safely retrieve the items, but you don't get to keep the bag of holding. If you want to keep the bag, you have empty it first and risk losing some of the contents. There is now an opportunity cost.

### DIFF
--- a/src/pickup.c
+++ b/src/pickup.c
@@ -2224,14 +2224,12 @@ loot_mon(struct monst *mtmp, int *passed_info, boolean *prev_loot)
 static boolean
 mbag_explodes(struct obj *obj, int depthin)
 {
-    /* these won't cause an explosion when they're empty */
-    if ((obj->otyp == WAN_CANCELLATION || obj->otyp == BAG_OF_TRICKS)
-        && obj->spe <= 0)
+    /* won't cause an explosion when it's empty */
+    if (obj->otyp == BAG_OF_TRICKS && obj->spe <= 0)
         return FALSE;
 
     /* odds: 1/1, 2/2, 3/4, 4/8, 5/16, 6/32, 7/64, 8/128, 9/128, 10/128,... */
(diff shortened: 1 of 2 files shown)


nhmall:
It looks like commit b2d4b77d3a1858869773e5c5bc2384a953b38e6a had a typo causing this PR to be closed. Re-opening.

copperwater:
I believe I said this on IRC or elsewhere but reiterating here: I don't really like this change, not because of its own merits, but because of the door it opens. (Actually, on its own merits I kind of like the decision it sets up when you find a cursed bag of holding in bones - would you tip it and lose some contents but keep the bag as it is, or cancel it and get all the contents?)

That said: If a magic bag can be cancelled into a normal bag, then shouldn't other items turn into nonmagic counterparts whenever possible? A magic harp becomes a wooden harp, magic lamp becomes an oil lamp and so on. It seems almost like polymorph (since that's another way, albeit unpredictable, to turn a magic tool into a nonmagic counterpart) - should never-polymorph-an-item conduct be broken since the item type has changed?

Also, does it really make the game more fun to create enchanted -> nonenchanted mappings for items, without also having the ability to go the other way and enchant a nonenchanted item into the magic counterpart? (I would object to this a lot less if that were the case, and there's a commit in xNetHack that implements it: https://github.com/copperwater/xnethack/commit/05a8c26)

PR #628, opened 2021-11-01
Remove duplicate code

argrath:
Here, `then` clause and `else` clause is identical.

### DIFF
--- a/src/role.c
+++ b/src/role.c
@@ -1406,17 +1406,10 @@ root_plselection_prompt(
     if (alignnum != ROLE_NONE && alignnum != ROLE_RANDOM
         && ok_align(rolenum, racenum, gendnum, alignnum)) {
         /* if race specified, and multiple choice of alignments for it */
-        if ((racenum >= 0) && (aligncount > 1)) {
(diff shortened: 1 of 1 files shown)


pat-rankin:
I don't think this fix is correct.  It retains the current behavior, but I suspect that that behavior is not what is intended.  The code probably started with copy+paste and one side of the then or else should have been revised but wasn't.

The code is constructing a string that might be used when prompting for something like "Pick a gender for your elven ranger:" where the user may have already specified 'chaotic' but it is omitted from the prompt because that is the only possible choice for elf.  The various permutations for allowed role+race+alignment+gender become fairly complicated when the user can specify any of those in advance (possibly by negating values that aren't wanted rather than just by picking ones that are wanted) but isn't required to specify all of them until later.

Most likely that 'else' should just be removed.  But fixing that piecemeal is error prone.  It would probably be best for someone (other than me :-) to construct a program which generates every possible combination and for each one calls that routine and then displays the prompt plus the list of attributes that still need to be specified.  If any of the outcomes aren't right, the routine can be revised then the test can be repeated.

PR #632, opened 2021-11-11
Allow bullwhips to grab items when flying

RojjaCebolla:
Generally Flying is "levitation, but better", and this is reflected by it
occupying a more valuable item slot, an amulet rather than a ring. One case
where that is not yet true is when an item is submerged in water. The item
can be grabbed by a levitating player, but not a flying player.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -2763,7 +2763,7 @@ use_whip(struct obj *obj)
                 (void) fire_damage(uwep, FALSE, u.ux, u.uy);
             return 1;
         }
-        if (Levitation || u.usteed) {
+        if (Flying || Levitation || u.usteed) {
             /* Have a shot at snaring something on the floor */
             otmp = g.level.objects[u.ux][u.uy];
             if (otmp && otmp->otyp == CORPSE && otmp->corpsenm == PM_HORSE) {


entrez:
> One case where that is not yet true is when an item is submerged in water. The item can be grabbed by a levitating player, but not a flying player.

Doesn't trying to grab an item submerged in water with a whip hit the "You cause a small splash." case whether levitating or not?  It comes just before the levitation check modified by the PR.

https://github.com/NetHack/NetHack/blob/90aded3648747179de91b6df9a50bcf670ae715c/src/apply.c#L2760-L2766

I don't see why you shouldn't be able to whip downward while flying though, maybe just an oversight since before 3.7.0 you could only fly via your steed, so the `u.usteed` test was sufficient. (edit: Never mind, I forgot about polyself... Maybe it's just that whipping down while flying doesn't really benefit the player, since any item you could grab with a whip while levitating, you can also grab with <kbd>,</kbd> while flying.  But I don't see why it should be impossible to do.)

RojjaCebolla:
Oh, hey, you're right. I was playing on 3.6 when I noticed the discrepancy. So water whipping is no longer a thing. Still, might wanna be able to whip downwards from flight, to match levitation.
Also it might save the player a turn if they were flying over a pit?

PR #636, opened 2021-11-18
Using the() with monster name

entrez:
the() treats most capitalized words as proper names, which caused
problems when it was used to prepend 'the' to a monster name.

Using the() with monster names is intentionally avoided in most places
for this reason, but there are still quite a few spots where it's used,
e.g., formatting killer names, applying a stethoscope to a statue,
hitting something with a monster egg, attempting to polymorph into an
invalid form...  Rather than replacing all these uses of the(), have it
check whether the string is an exact match for a non-pname monster name,
and if so, ignore capitalization.

### DIFF
--- a/src/objnam.c
+++ b/src/objnam.c
@@ -1812,6 +1812,8 @@ the(const char* str)
 {
     char *buf = nextobuf();
     boolean insert_the = FALSE;
+    const char *rest;
+    int mnum;
 
     if (!str || !*str) {
         impossible("Alphabet soup: 'the(%s)'.", str ? "\"\"" : "<null>");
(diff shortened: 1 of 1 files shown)


entrez:
Some weird stuff I noticed in the game that prompted this:
* Applying a stethoscope to a statue of an Archon or a Keystone Kop or a Nazgul will produce "Nazgul (or whatever) is in fine health for a statue." rather than "The Nazgul..."
* Attempting to polymorph into the Oracle produces "You can't polymorph into Oracle."

I'm not sure whether it would be a better solution to just wholesale replace all the uses of `the([monster name])` with something that isn't conditional in the same way that `the()` is.  Obviously feel free to close this and do it the other way instead; these are the examples I noticed:

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/apply.c#L274-L276

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/attrib.c#L296-L298

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/polyself.c#L498-L499

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/timeout.c#L610-L612

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/uhitm.c#L995-L998

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/uhitm.c#L3306-L3308

https://github.com/NetHack/NetHack/blob/937eef4b9b3edf6734d33c5acb7f57008b801924/src/uhitm.c#L4362-L4363

PR #637, opened 2021-11-23
Use %d for short

argrath:
obj->otyp is short, so corresponding format string is %d.

(from Coverity Scan)


moon-chilled:
%hd, not %d

PR #646, opened 2021-12-18
Correct the genders of DCSS bogusmons

RojjaCebolla:
Reference: https://github.com/crawl/crawl/blob/master/crawl-ref/source/mon-data.h


paxed:
Merged

PR #656, opened 2022-01-16
Multiple definitions of getreturn_enabled and GUILaunched

feiyunw:
This patch is to fix the following linking error when I was porting NetHack to MSYS2/mingw-w64:
D:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: o/winnt.o:winnt.c:(.bss+0x160): multiple definition of `getreturn_enabled'; o/windmain.o:windmain.c:(.bss+0x100): first defined here
D:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: o/guistub.o:stubs.c:(.bss+0x0): multiple definition of `GUILaunched'; o/nttty.o:nttty.c:(.bss+0x24): first defined here

### DIFF
--- a/sys/winnt/stubs.c
+++ b/sys/winnt/stubs.c
@@ -10,7 +10,7 @@
 #error You cannot compile this with both GUISTUB and TTYSTUB defined.
 #endif
 
-int GUILaunched;
+extern int GUILaunched;
 struct window_procs mswin_procs = { "-guistubs" };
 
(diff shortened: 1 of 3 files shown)


feiyunw:
I need think it further.

PR #660, opened 2022-01-20
Indicate to players that monsters are sleeping.

NullCGT:
This pull request makes the following changes:

* When a player looks at a monster, they can tell whether or not that monster is asleep.
* Display a message when a monster wakes up within sight of the player.
* Add YAFM for waking up a flesh golem.

Note that none of these changes do not consistently impact monsters put to sleep by wands of sleep, since wands of sleep alter mcanmove and mfrozen when putting a monster to sleep for a nonzero number of turns, rather than setting msleeping to 1.

The rationale for these changes is that up until this point, there has been no way to tell whether or not a monster is asleep aside from its lack of movement.

### DIFF
--- a/doc/fixes37.0
+++ b/doc/fixes37.0
@@ -1351,6 +1351,8 @@ drinking blessed potion of full healing heals wounded legs, either hero's or
 	steed's (when riding, wounded leg injury applies to steed, not hero)
 drinking uncursed potion of full healing or blessed potion of extra healing
 	heal hero's wounded legs when not riding; no effect on steed if riding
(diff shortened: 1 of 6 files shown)


copperwater:
See also https://github.com/copperwater/xnethack/commit/fde35333d4bbbf5adca101e6b03fa8c6a230bdc6 (and followup https://github.com/copperwater/xnethack/commit/4bb56a8fc24e735e0d3c3ef0daaee3f77b66696b), which aims to basically turn `mtmp->msleeping = 0;` into a general function for waking up (which can decide whether to tell the player the monster wakes up or not)

PR #668, opened 2022-02-05
Remove redundant code on find_guard_dest()

argrath:
The values of lx and ly are always assigned just before their usage.
So, assignments in advance are redundant.

### DIFF
--- a/src/vault.c
+++ b/src/vault.c
@@ -256,13 +256,13 @@ uleftvault(struct monst *grd)
 static boolean
 find_guard_dest(struct monst *guard, xchar *rx, xchar *ry)
 {
-    register int x, y, dd, lx = 0, ly = 0;
+    register int x, y, dd, lx, ly;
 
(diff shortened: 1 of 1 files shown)


nhmall:
In 3.4.3, it was:
```
	      if(levl[x][y].typ == CORR) {
		  if(x < u.ux) lx = x + 1;
		  else if(x > u.ux) lx = x - 1;
		  else lx = x;
		  if(y < u.uy) ly = y + 1;
		  else if(y > u.uy) ly = y - 1;
		  else ly = y;
		  if(levl[lx][ly].typ != STONE && levl[lx][ly].typ != CORR)
		      goto incr_radius;
```
where all the assignments to lx were contained within "if" statements, and some compilers complain about "potentially uninitialized local variable" in those cases (despite the final "else" clause).

When the code got reworked, the code developer probably didn't go back to figure out if the initializer could safely be  removed from the declaration of the variables involved. Harmless, but as you say, redundant.

PR #673, opened 2022-02-08
Fix: tty map weirdness after 'full-screen' menu

entrez:
When a 'full-screen' (cw->offx and cw->offy both 0) menu was immediately
followed by an offset menu -- as in the case of selecting certain
options from the options menu, or using loot to take out/put in items
after using ':' to describe the contents of a very full container --
there were some odd interactions with the map.

Only certain parts of the map near/under the menu window would be
redrawn if the first offset menu was followed by another one, while
a getlin prompt would cause the entire map to be redrawn, with parts
intersecting the window being drawn on top of it and obscuring it.
Flushing the display immediately after the docrt call when closing a
full-screen menu seems to fix both these issues.

It's a little hard to describe the 'weirdness' so here are some illustrative videos:
Before:

https://user-images.githubusercontent.com/40038830/153090341-45180ef1-09a3-44fa-9225-c20f976fb072.mp4

After:

https://user-images.githubusercontent.com/40038830/153090352-0646ae37-2730-4b0f-9a10-ad04519fe7c4.mp4

### DIFF
--- a/win/tty/wintty.c
+++ b/win/tty/wintty.c
@@ -1573,6 +1573,7 @@ erase_menu_or_text(winid window, struct WinDesc *cw, boolean clear)
             clear_screen();
         } else {
             docrt();
+            flush_screen(1);
         }
     } else {
         docorner((int) cw->offx, cw->maxrow + 1);


pat-rankin:
I reproduced this a couple of times with current code and not with 3.6.4 so I decided to try 'git bisect' to find out where it changed.  No luck; I made it all the way past the commit that was HEAD when you made your change without being able to reproduce the problem again.  I repeated the exercise, this time using ^F and possibly wand of digging to  position the hero where the smaller menu would cover that location.  Still no luck.  So either I screwed up along the way to reproducing this is tricky.

entrez:
Weird -- I can reproduce both aspects of this behavior in 3.4.3, even (well, the version of 3.4.3 on hardfought, so it has some community patches applied).  They happen for me in 3.6.6 as well.  Neither the 'use ':' to make the map suddenly appear' nor the 'multiple menus in succession can cause the intersecting part of the map to be redrawn' things are happening?

https://user-images.githubusercontent.com/40038830/153677714-9b194667-c306-4e27-ad9c-912dcae8e720.mp4


What I have done in the video (opening options, selecting 'autopickup exceptions', then 'list exceptions', spacebar, and pressing ':') is pretty consistent in reproducing it for me, at least when my terminal is small enough (e.g. something like 120 cols, 45 rows for the video in this comment).

PR #684, opened 2022-02-23
Fix: monster hurtling and liquid

entrez:
A monster hurtling over liquid would drown immediately the instant it
touched the first square of water, even if normally it would have kept
moving (e.g. hurtling over a short moat).  Additionally, its placement
on liquid would not take into consideration other monsters, so it could
overwrite an existing monster on that spot and lead to an impossible,
and/or two monsters occupying a single position.

Fix these issues, so that liquid effects like drowning only happen if
the monster ends up in liquid at the end of the hurtle, and so that
other monsters in the way will stop it early even if they're floating
over or swimming on a pool/water/lava square.

Also use canspotmon instead of canseemon for the wiztelekinesis debug
command.

### DIFF
--- a/include/hack.h
+++ b/include/hack.h
@@ -297,6 +297,7 @@ typedef struct sortloot_item Loot;
                                * to make an extra call to goodpos()]        */
 #define GP_ALLOW_U  0x100000L /* don't reject hero's location */
 #define MM_NOEXCLAM 0x200000L /* more sedate "<mon> appears." mesg for ^G */
+#define MM_IGNORELAVA   0x400000L /* ignore lava when positioning */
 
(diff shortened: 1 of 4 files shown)


entrez:
I marked this as a draft because I realized I wasn't sure if 'hurtling' entails sliding across the floor (in which case it makes sense that the first square with liquid would end it) or being thrown through the air; after some discussion on IRC most people seem to think that it probably involves flying through the air, so I'm reopening it, at least for now... if it should be more like sliding on the floor, it ought to be a pretty small tweak to bring it in line with that.

paxed:
Merged

PR #699, opened 2022-03-14
Reduce eucalyptus leaf nutrition to 1

copperwater:
Eucalyptus leaves are famously inedible except by certain animals such
as koalas. I consider it very strange that a single leaf in NetHack
gives you six meatballs' worth of calories.

I considered making it 0 nutrition, but am not sure if a 0-nutrition
comestible would end up violating some assumption that all food is at
least 1 nutrition. If that's overly cautious and it's safe to do this, feel free
to make it 0 instead of 1.


paxed:
Committed.

PR #700, opened 2022-03-14
Implement selection addition and difference

copperwater:
Selection difference is something I have found myself wanting a lot when
working on levels, and have had to defer to a clunkier xor-then-and
approach. This commit implements the TODO-ed addition and subtraction
operators on two sets.

I don't see how the addition operator would be any different from
logical or, so it just calls l_selection_or rather than implement a new
function.

### DIFF
--- a/doc/lua.adoc
+++ b/doc/lua.adoc
@@ -793,11 +793,13 @@ Example:
 
 === Logical or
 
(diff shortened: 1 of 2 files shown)


paxed:
Committed.

PR #701, opened 2022-03-15
Fix selection "random" grow direction, and other code cleanup

copperwater:
Noticed that when I set a selection to grow in a random direction, it
instead grew in all directions, which is not what I wanted. Turns out
the -1 random dir ended up being passed straight to the code which
checks bitmasks, without any form of randomizing among directions.

So this adds code to do that, and defines W_RANDOM as -1 rather than
using a magic number. In the process I also noticed that specifying
"random" as the wall for a door in a room made it rerandomize the
direction every iteration of its loop, essentially rolling two rn2(4)s
and only proceeding if they matched. That was pointless so I cleaned it
up a bit.

Also added safety checks in the form of an impossible for des.corridor()
being called with "random" as either walldir, because this is not
implemented currently.

And lastly, I noticed that create_secret_door was entirely unused
(secret door creation is handled in create_door), so I deleted it.

The only behavior change caused by this is that the Valkyrie quest lava
pools will be a little smaller, which is the only place grow is
currently used. If it's desired to keep them the same, that should be
changed to "all".

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -2496,7 +2496,6 @@ extern void flip_level_rnd(int, boolean);
 extern boolean check_room(xchar *, xchar *, xchar *, xchar *, boolean);
 extern boolean create_room(xchar, xchar, xchar, xchar, xchar, xchar, xchar,
                            xchar);
-extern void create_secret_door(struct mkroom *, xchar);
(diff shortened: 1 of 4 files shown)


paxed:
Committed

PR #703, opened 2022-03-15
Fix: levitating/flying monsters moving over liquid

entrez:
Commit c1a6dd4 was meant to prevent flying, levitating, and clinging
monsters from considering walls of water as acceptable movement
destinations, even outside of the Plane of Water.  However, it was
evaluating the monster's starting position instead of possible places to
move to, and the evaluation was 'backwards' (the equivalent of
IS_WATERWALL, instead of !IS_WATERWALL).

The result was that non-swimming monsters could only move onto any kind
of water or lava square if the position they were moving from was a
WATER square.  Change this so that instead of the starting position,
each potential destination spot's status as a wall of water is evaluated
in turn, and reverse the effect of the test so that it blocks walls of
water instead of allowing them.

### DIFF
--- a/src/mon.c
+++ b/src/mon.c
@@ -1813,8 +1813,7 @@ mfndpos(
 
     nodiag = NODIAG(mdat - mons);
     wantpool = (mdat->mlet == S_EEL);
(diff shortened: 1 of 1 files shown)


paxed:
Committed.

PR #705, opened 2022-03-18
Collisions while hurtling can stone participants

entrez:
Hurtling into a monster is described as "bumping into" it, so it makes
sense that hurtling willy-nilly into a cockatrice (or vice-versa) could
result in petrification.  Since hurtling for the hero usually involves
"floating in the opposite direction" (presumably backwards) after
throwing an item, check whether the hero is wearing any body armor which
would cover their torso rather than looking for gloves.  Do the same for
monsters on the general basis that it's a bodily collision, and for the
sake of consistency.

https://user-images.githubusercontent.com/40038830/158918345-8b427077-1d4c-4670-b6db-da07da79d6aa.mp4


paxed:
Committed

PR #708, opened 2022-03-19
Fix: repeated 'hit with a wielded weapon' logging

entrez:
The "hit with a wielded weapon for the first time" livelog line could be
produced repeatedly: it was triggered by hitting a monster with a
wielded object of any sort, but the u.uconduct.weaphit counter was only
incremented if hitting with an actual 'weapon' (a WEAPON_CLASS or
is_weptool item).  As a result, if a non-weapon-using hero whipped out a
non-weapon item -- a cockatrice corpse, for example -- and started going
to town on some monsters, the livelog message would be repeated with
every hit.

### DIFF
--- a/src/uhitm.c
+++ b/src/uhitm.c
@@ -1323,6 +1323,8 @@ hmon_hitmon(
 
     if (!already_killed) {
         if (obj && (obj == uwep || (obj == uswapwep && u.twoweap))
+            /* known_hitum 'what counts as a weapon' criteria */
+            && (obj->oclass == WEAPON_CLASS || is_weptool(obj))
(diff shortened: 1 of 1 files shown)


entrez:
I feel like ideally, it'd be preferable to have the livelog message printed at the same time as the counter increments, rather than trying to sync the criteria for the two events as I've done here.  But after reading this comment, which sounds as though they were decoupled for a good reason, I didn't really look into that possibility any further for fear of messing things up...

https://github.com/NetHack/NetHack/blob/64371dc8b39e86fc95b67ea1ab0bda645fc22bce/src/uhitm.c#L747-L750

PR #711, opened 2022-03-23
Make nh.getmap() return information relative to the most recent des.map

copperwater:
There are many possible use cases for nh.getmap during level creation,
but it's rendered mostly unusable by virtue of always returning data
about the exact x,y coordinate in g.level.locations. (In particular,
it can't currently be used in themed rooms at all, because the themed
room could be anywhere on the level.) This is inconsistent with how most
other coordinate-based functions work following a des.map, which use
coordinates relative to the 0,0 point of the map.

This changes it so that during level creation only, if nh.getmap is used
following a des.map statement, it will look up the coordinates relative
to the origin of the map, consistent with the other functions.

### DIFF
--- a/dat/themerms.lua
+++ b/dat/themerms.lua
@@ -80,8 +80,7 @@ themerooms = {
                     if (percent(25)) then
                        local mintime = 1000 - (nh.level_difficulty() * 100);
                        local ice_melter = function(x,y)
-                          local ax,ay = nh.abscoord(x,y);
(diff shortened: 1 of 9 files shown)


paxed:
Needs to be updated. Tests should be changed to match.

copperwater:
Updated and expanded the scope of this pull request. This now standardizes all functions in `nh` and `obj` which take coordinate input to interpret that input as relative to the last defined map or room rather than absolute, clearing up a complaint in lua.adoc about them being inconsistent between those two namespaces and `des`.

paxed:
committed

PR #735, opened 2022-04-15
Fix: repeating 'free' actions with ^A takes time

entrez:
The do_repeat function returned ECMD_TIME whenever it found something to
try to repeat, including 'free' actions like checking inventory, or
even pressing a key that's not bound to any function.  Returning ECMD_OK
has the opposite problem; repeating an action never takes time, even if
it otherwise would.

Add a new return value, ECMD_PASS, which preserves the relevant
variables set by the actual action being repeated rather than resetting
or modifying them, so that actions which take time will still take time
when repeated and 'free' actions won't.


entrez:
Fixed another way in 12a63216a

PR #737, opened 2022-04-18
Monster AI Improvements and monmove.c refactoring.

NullCGT:
This pull request contains a number of minor AI tweaks and improvements:

- When the player is on a set of stairs, tame monsters acts as if the player is carrying a treat. This reduces the frustration experienced by many players when they attempt to bring their pets with them, while still allowing for intentionally leaving pets behind on other dungeon levels.
- Enable offensive use of wands of teleportation by monsters. While comments in the code indicate that this behavior was disabled simply because it never occurred, I have found that in practice, this is not the case. When this code is enabled, monsters zap wands of teleportation at players standing on staircases, exactly as intended.
- Allow piranhas to eat corpses in the same manner that purple worms do. In pop culture, piranhas are infamous for their ability to skeletonize animals remarkably quickly, and I thought this would be nice to represent in NetHack.

This pull request also contains some significant refactoring and readability improvements for monmove.c:

- Pull illithid mind blast behavior into its own function. This shouldn't create much overhead, and enormously improves the readability of dochug().
- Eliminate a number of gotos in movemon.c.
- Replace magic numbers in movemon.c with constants.

I'm far from done with refactoring movement code, but I think this pull request is a good place to start. Dochug() is arguable the worst spaghetti code in the entire codebase, and I think that addressing it starts with making it more legible.

### DIFF
--- a/src/monmove.c
+++ b/src/monmove.c
@@ -1828,13 +1828,14 @@ set_apparxy(register struct monst* mtmp)
      */
 
     /* pet knows your smell; grabber still has hold of you */
-    if (mtmp->mtame || mtmp == u.ustuck)
-        goto found_you;
-
-    /* monsters which know where you are don't suddenly forget,
+    if (mtmp->mtame || mtmp == u.ustuck ||
+        /* monsters which know where you are don't suddenly forget,
        if you haven't moved away */
-    if (u_at(mx, my))
-        goto found_you;
+        u_at(mx,my)) {
+            mtmp->mux = u.ux;
+            mtmp->muy = u.uy;
+            return;
+    }
 
     notseen = (!mtmp->mcansee || (Invis && !perceives(mtmp->data)));
     notthere = (Displaced && mtmp->data != &mons[PM_DISPLACER_BEAST]);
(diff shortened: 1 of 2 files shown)


NullCGT:
I just learned that in the time since I worked on this, 3.7 has added constants in monmove.c, meaning that this pull request will not cleanly merge.

If there is still interest, I can update the PR.

paxed:
Please do update this.

NullCGT:
> Please do update this.

Updated!

entrez:
> Enable offensive use of wands of teleportation by monsters. While comments in the code indicate that this behavior was disabled simply because it never occurred, I have found that in practice, this is not the case.

The comment is a little misleading and I read it this way too, but if you look at 8d1e0d17565ca7567cd6c75565a3e5c6e87f8db1 basically it used to be the case that it was "supported" but never selected, then full support for monster offensive use of teleportation was added by @pat-rankin but deliberately left disabled.

NullCGT:
@entrez Excellent catch.

paxed:
Any chance you could rebase it on top instead of merging?

NullCGT:
Rebased.

paxed:
Pushed.

PR #748, opened 2022-04-30
Mark high altars in altarmask, fix display of unaligned temple altars

entrez:
I noticed I had a branch from a few months ago with this change to altarmask,
which I don't *think* I ever submitted to the devteam (my apologies if I did
and just forgot about it!).  I looked it over and it seemed like a decent idea
to me, so I thought I may as well send it in now -- and if there's some big
mistake or other reason I never submitted it that I've since forgotten,
hopefully that will become apparent. :)

While testing that commit, I also noticed that altars in unaligned temples
(like the one in the Valley) weren't being displayed as the right color.

- Designate high altars with dedicated altarmask bit
- Fix coloring of unaligned temple altars

### DIFF
--- a/include/align.h
+++ b/include/align.h
@@ -34,11 +34,13 @@ typedef struct align { /* alignment & record */
 /* Some altars are considered shrines, add a flag for that
    for the altarmask field of struct rm. */
 #define AM_SHRINE       0x08
+/* High altar on Astral plane or Moloch's sanctum */
+#define AM_SANCTUM      0x10
 
 /* special level flags, gone by the time the level has been loaded */
-#define AM_SPLEV_CO     0x10 /* co-aligned: force alignment to match hero's  */
-#define AM_SPLEV_NONCO  0x20 /* non-co-aligned: force alignment to not match */
-#define AM_SPLEV_RANDOM 0x40
+#define AM_SPLEV_CO     0x20 /* co-aligned: force alignment to match hero's  */
+#define AM_SPLEV_NONCO  0x40 /* non-co-aligned: force alignment to not match */
+#define AM_SPLEV_RANDOM 0x80
 
 #define Amask2align(x) \
     ((aligntyp) ((((x) & AM_MASK) == 0) ? A_NONE                \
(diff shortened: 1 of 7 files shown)


include/align.h
```diff
@@ -34,11 +34,13 @@ typedef struct align { /* alignment & record */
 /* Some altars are considered shrines, add a flag for that
    for the altarmask field of struct rm. */
 #define AM_SHRINE       0x08
+/* High altar on Astral plane or Moloch's sanctum */
+#define AM_SANCTUM      0x10
```
copperwater, line 38: I can see the choice of `AM_SANCTUM` misleading people to think it _only_ means the Sanctum altar, since nothing in the game calls the temples on the Astral Plane sanctums. Shouldn't it be `AM_HIGH` or `AM_HIGH_ALTAR` to correspond with the usual terminology?

entrez: Yeah, I know what you mean... the reason I called it `AM_SANCTUM` is because that's the altar type in the level files (it dates back to the old des format):

https://github.com/NetHack/NetHack/blob/721b1cc9b92a802cdf542a09404c8562e16799d1/dat/astral.lua#L82-L91

entrez: I wasn't sure about it either, but since they're called 'sanctums' internally I figured this would more consistent with the rest of the level generation stuff...  You can see another reference here:

https://github.com/NetHack/NetHack/blob/721b1cc9b92a802cdf542a09404c8562e16799d1/src/sp_lev.c#L2330-L2334

It parallels the use of the internal/level file term 'shrine' in `AM_SHRINE` for describing temple altars, too.  But you're right that the potential for confusion is higher in this case.

copperwater: I'd argue this is a good opportunity to adjust sp_lev.c to accept "high" instead of "sanctum" for high altars.

entrez:
Now that `EDITLEVEL` has been incremented in 44d5be6 this PR no longer modifies patchlevel.h in a meaningful way.  Since there hasn't been a great rush to get this merged, I'm just going to remove the `EDITLEVEL` change for now (by rebasing on 44d5be6) so I don't have to force-push it every time editlevel is increased.  I can always put it back in the future if need be.

include/align.h
entrez: I mentioned this on IRC but just for the record: "high" or "highaltar" does seem like a more intuitive name for the type, though that change feels a bit outside the intended scope of this PR.  I'll definitely update it to be consistent with the internal terminology if/when it changes.

PR #749, opened 2022-05-01
add msghandler support for win32

argrath:


pat-rankin:
This is missing an update to src/mdlib.c in the code that makedefs uses to construct dat/options which is displayed during play by nethack's '#version' command.

argrath:
Updated for it.

PR #761, opened 2022-05-13
Prevent hero from writing Pratchett books

entrez:
The hero's ability to channel Pratchett and write his books with a magic
marker once she had read or IDed at least one of them seemed strange,
especially cases like an illiterate hero doing it as her first
introduction to the written word.  Block the hero from writing random
novels with a marker.

The image of the hero sitting down in the dungeon to write a novel is
funny, so it feels like a good spot for a funny message.  I'm not
sure if what I have there is perfect, but it can always be changed.

### DIFF
--- a/src/write.c
+++ b/src/write.c
@@ -220,6 +220,11 @@ dowrite(struct obj *pen)
         You_cant("write that!");
         pline("It's obscene!");
         return ECMD_TIME;
+    } else if (i == SPE_NOVEL) {
+        You(
+         "prepare to write the Great Yendorian Novel, but lack inspiration.");
+        You("give up on the idea.");
+        return ECMD_TIME;
     } else if (i == SPE_BOOK_OF_THE_DEAD) {
         pline("No mere dungeon adventurer could write that.");
         return ECMD_TIME;


k21971:
`You("prepare to write the Great Yendorian Novel, but lack %s.",
         Hallucination ? "perspiration" : "inspiration");`

jeremyn:
If this PR were added it would be nice to balance it out by making it easier to get a novel somewhere else. What about adding a guaranteed novel somewhere? Vlad's Tower or Mine's End would be thematically appropriate, plus an extra one in the Tourist Quest home level. I might make a PR for that if there's support for the idea here.

entrez:
Since novels aren't actively beneficial in the way a spellbook or scroll is, I doubt many people spend their limited marker charges (and blank spellbooks) on them, even if it's technically possible.  I suspect it wouldn't have a significant effect on the number of novels people encounter in-game.

There's already a guaranteed novel in the first book or scroll shop you encounter, by the way -- so they aren't guaranteed in the sense one necessarily appears in every single game, but there are definitely measures in place to make them available.

moon-chilled:
I think 'set down' is better than 'prepare' (and approve of the 'perspiration' bit), but otherwise love it.

jeremyn:
@entrez It's fun trying to ascend with a novel as a silly extra challenge, and if reasonable to do it, swing by Death for a #chat. Novels are pretty rare so getting one can mean either using a wish, using some marker charges, or polypiling. Making novels impossible to write might remove a minor but potentially interesting decision from the player.

I know this sounds like a weird edge case but I actually had a fun and memorable game where I wrote a novel to prepare for the Planes, lost it in a fire attack on Astral, and ended up getting a second novel by wresting a final wish from a wand shortly before ascending. While trying to get that last wish I had been kiting Pestilence around my high altar, zapping him with a wand of death, and avoiding my high priest since I was causing conflict. I ran out of death charges and decided I needed to ascend or risk the entire game, and it was only then that I got that wish. It was awesome.

So, I'm coming at this from the perspective that maybe that fun game would have been a little less fun with this PR. I hope that makes sense.

jeremyn:
It looks like this was merged with some modifications:

* with 2/3s chance you try to write a novel, otherwise you try to write fan-fiction;

* with a separate 2/3s chance, you "give up on the idea" and keep the blank book, otherwise you "tear up" your attempt and lose the blank book.

You never get a novel. Not what I was hoping for but it is what it is.

It would be nice if there were feedback to the player after the first attempt that what they were trying to do will never work, particularly since it works now, unless the intention is to punish players for trying it.

jeremyn:
I should clarify that you don't waste marker charges in any case, despite some small ambiguity in the messages. (Why would you tear up a completely blank book in frustration?) Losing a blank book with a 1/3 chance is a trivial cost so the main penalty is wasting the player's time on repeated attempts.

Specific suggestion: append "It's hopeless!" or "It's no use!" to the final give/tear up message.

PR #762, opened 2022-05-13
Fix: loot not drawn on map when mon dies

entrez:
When a monster died, any loot it carried would not be drawn onto the map
until the square was refreshed in some other way, because the newsym
call was moved (as part of the introduction of mon_leaving_level in
6880d37b3) to before the point its inventory items are actually placed
onto the map.  Move mon_leaving_level to after the relobj call in
m_detach so the new symbol will take into consideration any items placed
onto the map from the monster's inventory.


entrez:
Okay, this needs reconsideration given the apparent need to reset mtrapped, etc, before the call to `relobj` (based on the comment, I'm guessing that's the case).

PR #768, opened 2022-05-21
Livelog refused wishes and (non-cursed) genocides

vultur-cadens:


entrez:
My two cents: using "refused to perform genocide" for a failed attempt to genocide something, not just an actual refusal, seems weird to me.  I also feel like "refused" can be a somewhat emotionally charged word (maybe this is just me, but "refused to perform genocide" sounds like the hero is taking a moral stand or something to me); in my opinion a more neutral word like "declined" might fit better.

I think the goto labels should not be indented at all, if you want to be consistent with the rest of the code.

Also, it seems like reusing the exact same label in the two functions could potentially be confusing: if you grep for it you might end up in the wrong place.

vultur-cadens:
Updated.

copperwater:
Failing to spell species correctly (or attempting to genocide ungenocidable species) a sufficient number of times doesn't really constitute "declining to perform genocide", since by all indications the player was trying to genocide something but never landed on something that the game would accept. This should only livelog it when the player explicitly says to genocide nothing.

(You could have a separate livelog for the "That's enough tries!" case, but is that really interesting enough to log? I don't think so.)

PR #788, opened 2022-06-03
Fix: wished-for doors in wizmode always vertical

entrez:
Wishing for a door is intended to retain the existing 'horizontal' value
of the surrounding wall or door (see comment in the wizterrainwish
'door' case).  However, the field was being reset by mistake, causing
all door wishes to create vertical doors.  Preserve it as intended.


entrez:
NB: this cannot be tested without something like #787 since wishing for a door (or other terrain/furniture) segfaults as of the current HEAD (dca1cd7)

PR #791, opened 2022-06-06
split cleaning-up on gd_move() into separate function

argrath:


nhmall:
There seems to be some issues:
1. The prototype and the function don't agree on the type of parameter 1 of gd_move_cleanup().
2. There's a variable see_guard that doesn't seem to be used in the function gd_move().

Possible fix:
```
diff --git a/src/vault.c b/src/vault.c
index f75c4e81e..231d389c8 100644
--- a/src/vault.c
+++ b/src/vault.c
@@ -15,7 +15,7 @@ static void move_gold(struct obj *, int);
 static void wallify_vault(struct monst *);
 static void gd_mv_monaway(struct monst *, int, int);
 static void gd_pick_corridor_gold(struct monst *, int, int);
-static int gd_move_cleanup(struct monst, boolean, boolean);
+static int gd_move_cleanup(struct monst *, boolean, boolean);
 
 void
 newegd(struct monst *mtmp)
@@ -813,7 +813,7 @@ gd_move(struct monst *grd)
     long umoney = 0L;
     boolean goldincorridor = FALSE, u_in_vault = FALSE, grd_in_vault = FALSE,
             semi_dead = DEADMONSTER(grd),
-            u_carry_gold = FALSE, newspot = FALSE, see_guard;
+            u_carry_gold = FALSE, newspot = FALSE;
 
     if (!on_level(&(egrd->gdlevel), &u.uz))
         return -1;

```

argrath:
Fix it.

PR #796, opened 2022-06-09
Let fleeing monsters dig pits in undiggable floor

entrez:
When the hero zaps a wand of digging down in an undiggable level, it
creates a pit.  When a fleeing monster did the same thing, it had no
effect.  Bring this closer to the behavior experienced by the hero: if a
monster tries to use a wand of digging to create a hole in an undiggable
floor, a pit will be made (and the monster will fall into it, if not a
flyer).

### DIFF
--- a/src/muse.c
+++ b/src/muse.c
@@ -781,10 +781,23 @@ use_defensive(struct monst* mtmp)
             return 2;
         }
         if (!Can_dig_down(&u.uz) && !levl[mtmp->mx][mtmp->my].candig) {
-            if (canseemon(mtmp))
-                pline_The("%s here is too hard to dig in.",
-                          surface(mtmp->mx, mtmp->my));
-            return 2;
+            /* can't dig further if there's already a pit (or other trap)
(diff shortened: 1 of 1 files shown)


pat-rankin:
I've put this is (it'll a little while before it propagates to github) but wasn't able to test it properly.  Monsters won't choose wand of digging as defensive item when they can't dig down.

entrez:
Yeah, I think the only place it can happen (i.e. the only circumstance where `Can_dig_down` is false but the rules in `find_defensive` don't exclude using a wand of digging) is non-end-of-branch levels flagged `hardfloor`, which seems to pretty much be true only on certain quest levels.  Maybe the fact monsters can choose to use wands of digging there is a bug -- I didn't really think about that, I think I assumed the rule was meant to exclude levels which are "obviously" non-diggable.

It would be cool if monsters could learn that a level was undiggable from experience, and use that information to realize digging down is futile, instead of never digging down on an undiggable floor -- maybe if something like that is ever implemented this patch would be more useful.

PR #800, opened 2022-06-15
split displaying trap map into separate function

argrath:


nhmall:
What purpose, if any, is the ttmp parameter serving in the added display_trap_map() function?
The first thing that seems to be done with it is a discard of any value it may have held when passed from the caller.

argrath:
Hmm, it seems thinko.
I'll submit a PR to fix it.

PR #802, opened 2022-06-16
Allow travel on water with IDed water walking

entrez:
If you have a known source of water walking (i.e. are wearing formally
IDed water walking boots), allow travel to path you over water and allow
running over water.  A transition from land to water will still cause
the hero to stop in modes other than the shift+dir "move until you hit
something" running mode, so that you don't careen across the entire
level unless you really want to.

Also, fix running over water when levitation or flying is equipped.
This stopped working after f88dce6, only allowing the hero to move
one square at a time.  And prevent travel from pathing the hero into a
wall of water even if flying or levitation is equipped, since they don't
allow you to bypass that terrain.

Edit: I originally had cause_known in here to check for known water
walking, but reconsidered since it's fairly expensive and the boots are
currently the only available source of water walking.  I changed it back
to check the hero's boots directly (as swim_move_danger already does).

### DIFF
--- a/src/hack.c
+++ b/src/hack.c
@@ -39,6 +39,16 @@ static void maybe_wail(void);
 
 #define IS_SHOP(x) (g.rooms[x].rtype >= SHOPBASE)
 
+/* XXX: if more sources of water walking than just boots are added,
+   cause_known(insight.c) should be externified and used for this */
+#define Known_wwalking \
+    (uarmf && uarmf->otyp == WATER_WALKING_BOOTS \
+     && objects[WATER_WALKING_BOOTS].oc_name_known \
+     && !u.usteed)
+#define Known_lwalking \
+    (Known_wwalking && Fire_resistance \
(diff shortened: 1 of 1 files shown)


entrez:
This is a leaner, more focused attempt to revisit #454, which has gained a myriad of merge conflicts since being opened (and also to fix the issue with running over water while levitating).

copperwater (commented):
Rather than allowing this code to become duplicated, shouldn't these checks be turned into macros or functions like `Known_wwalking` and `Known_lwalking` similar to other boolean-evaluatable properties like for instance `Antimagic`?

entrez:
Yes, I think that's a good idea.  To be honest, I thought along the same lines but wasn't sure how best to handle the reference to `known_wwalking` in the definition of `known_lwalking` with regards to macroifying them.  It could be replaced with the `Known_wwalking` macro (or whatever) I guess.

PR #803, opened 2022-06-16
remove unnecessary  `if` on help_dir()

argrath:
If `viawindow` is false, `prefixhandling` is always true.
So, inner `if` statement is unnecessary.

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -4587,15 +4587,11 @@ help_dir(
     /* if '!cmdassist', display via pline() and we're done (note: asking
        for help at getdir() prompt forces cmdassist for this operation) */
     if (!viawindow) {
-        if (prefixhandling) {
-            if (!*buf)
-                Sprintf(buf, "Invalid direction for '%s' prefix.",
-                        visctrl(g.Cmd.spkeys[spkey]));
-            pline("%s", buf);
-            return TRUE;
-        }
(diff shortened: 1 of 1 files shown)


entrez:
Actually, is `prefixhandling` ever true?

`help_dir` is called two times.  One of them, `NHKF_ESC` is passed directly:

https://github.com/NetHack/NetHack/blob/ab169d7961450c1eaabfdc2becee701ba8d73d97/src/cmd.c#L4479-L4482

The other one is given the value of `spkey`, but `spkey` is defined as `NHKF_ESC` and never used or given an opportunity to have another value:


https://github.com/NetHack/NetHack/blob/ab169d7961450c1eaabfdc2becee701ba8d73d97/src/cmd.c#L4080-L4084

...

https://github.com/NetHack/NetHack/blob/ab169d7961450c1eaabfdc2becee701ba8d73d97/src/cmd.c#L4248-L4249

The return value of `ch2spkeys`, which no longer even exists, was previously used for the value of `spkey`.  This seems like a leftover that the rhack refactor in e7fa065203 failed to convert to work within the new system (or remove).

Edit: I think that second `help_dir` is mostly superseded by the "The foo command does not accept 'bar' prefix" and "'bar' prefix has no effect for the foo command" feedback, but it still kicks in when you type a prefix followed by a key that's not assigned to any command, like <kbd>m</kbd><kbd>M</kbd>.

PR #821, opened 2022-07-15
Prevent amorphous mon from carrying hero into door

entrez:
An amorphous engulfer like a fog cloud could engulf the hero, then carry
him into a closed door.  If it was killed or decided to spit out the
hero, he would be left occupying the same spot as a closed/locked door.
Make an amorphous monster unable to move into a door if currently
engulfing the hero.

Something more complicated could be done along the lines of allowing the
move if the hero is himself in an amorphous polyform, but that verges on
being a little too silly, maybe.

I also included fixes to a couple miscellaneous, unrelated formatting
issues that I noticed recently.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -3164,7 +3164,7 @@ find_poleable_mon(coord *pos, int min_range, int max_range)
                 && (mtmp->mtame || (mtmp->mpeaceful && flags.confirm)))
                 continue;
             if (glyph_is_poleable(glyph)
-                    && (!glyph_is_statue(glyph) || impaired)) {
+                && (!glyph_is_statue(glyph) || impaired)) {
                 if (mpos.x)
                     return FALSE; /* more than one candidate location */
                 mpos.x = x, mpos.y = y;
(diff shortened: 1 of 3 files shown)


entrez:
The negations here are definitely somewhat tortured:

```
&& !((amorphous(mdat) || can_fog(mon))
     && !(u.uswallow && mon == u.ustuck))
```

I tried to think of a way to do it that's a little easier to parse but I couldn't figure one (besides using a variable for this 'able to move into door' property and stretching it out over a couple lines, maybe that's an alternative option).

pat-rankin:
I haven't looked at the proposed commit yet, but in the snippet I think the second line could be simplified to
 '&& mon != u.ustuck'.
That should give the right answer when mon is ustuck, when some other monst is ustuck, and when ustuck
is Null.  Checking uswallow seems superfluous.

entrez:
That's how I had it at first, but I was worried it would prevent an amorphous grabbing monster (if such a monster existed) from walking away through a door on the turn it lets go of the hero.

That's because I noticed this block, which makes it seem as though the 'unsticking' happens only after the grabber has selected a spot to move to and then moved (rather than letting go of the hero, and then moving somewhere, so that `u.ustuck` would be unset in `mfndpos`):

https://github.com/NetHack/NetHack/blob/5aca2fc590f5ee533ce89760b7bfdf2ad1d689dd/src/monmove.c#L739-L748

Edit: Or to put it more succinctly, I thought by checking `u.uswallow` it would limit the restriction only to `u.ustuck` monsters that would actually carry the hero along with them as they moved.

PR #823, opened 2022-07-16
Fix: getpos cmap vs typ/glyph confusion

entrez:
When some parts of getpos were rearranged in 7404597, tests of terrain
types and glyphs were moved to a loop which iterated through the defsyms
array without being updated to handle cmap values instead.  Consequently
they weren't excluding certain terrain types from being 'jumped to' as
intended.  (Though it seems as though they actually worked by chance to
some extent, just because there's some overlap between terrain types and
defsyms in terms of where the 'walls/doors' blocks start and end.)

I also noticed that cmap_to_type was missing S_darkroom (it fell through
to the default case so that the function returned STONE), so I added it.

### DIFF
--- a/src/do_name.c
+++ b/src/do_name.c
@@ -926,10 +926,11 @@ getpos(coord *ccp, boolean force, const char *goal)
                 int pass, lo_x, lo_y, hi_x, hi_y, k = 0;
 
                 (void) memset((genericptr_t) matching, 0, sizeof matching);
-                for (sidx = 1; sidx < MAXPCHARS; sidx++) { /* [0] left as 0 */
-                    /* don't even try to match some terrain: walls, room... */
-                    if (sidx <= S_hcdoor
-                        || sidx == S_room || sidx == S_darkroom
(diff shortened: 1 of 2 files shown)


entrez:
I added this additional commit that is a bit more 'efficient', but not sure if it is worth being a bit harder to comprehend.

I have also been wondering whether it's better to use `cmap_to_type` instead of baking in these assumptions about the order of `defsyms`.  I had it like that at first, which is why the edit to `cmap_to_type` is in that first commit, but it means you're feeding a lot of non-cmap stuff to it and getting `STONE` for them all, and therefore can't check for `STONE` to exclude it -- so you have to start the loop at 1, which is once again an assumption about the order of `defsyms`.  Not sure what the best way to go is.

entrez:
Okay, here's another way to do it that doesn't encode any assumptions at all about the order of defsyms.  Still not sure what the best way to do it is -- these followup commits can be squashed/discarded as needed.

By the way, `S_ndoor` isn't included in any of the `is_cmap_foo` macros?  Am I reading that right?

PR #829, opened 2022-07-28
Fixes related to running and traps

entrez:
I noticed some issues(?) with how traps interact with running: one recent-ish
change to behavior when running which removed an existing way for players to
override the normal run-stopping (using runmode 1), and a problem that
prevented the trap check from being evaluated most of the time in lookaround.

- Restore old behavior for running onto trap
- Fix: lookaround trap detection


paxed:
Pushed.

PR #830, opened 2022-07-29
Some lit corridor display tweaks

entrez:
A couple small tweaks to how lit corridors are displayed in various scenarios,
intended to address some minor issues I've noticed recently.

- Don't show corridors as lit in #terrain
- Update lit corridor display if dark_room toggled


paxed:
Pushed.

PR #831, opened 2022-07-31
Remove operator with side effect (increment) from array index accessing

un-ch:


pat-rankin:
Why?

paxed:
Not a "side effect"

un-ch:
The reason of current commit was to make code more readable: tried to write accessing to the array without using the value of increment. Thus, we do not have to "remember" differences between "prefix" and "postfix" operators while code reading.

PR #833, opened 2022-08-02
Lua test-related changes

entrez:
Some changes to improve the Lua tests.  Some of them I'm not too sure about:
they fix problems for me, but I'm not certain if it's the 'proper' fix (in one
case because it seems like there's some context I'm missing for how it came to
be the way it is, and in another because the NHL_SANDBOX system is complex and
I'm not sure if I'm using the flags correctly).

- Allow lua test files to call error()
- Fix: movement tests
- Fix: test_src -eaux plurals
- Add makesingular to test_src


paxed:
Committed.

PR #835, opened 2022-08-04
Fixes related to command counts

entrez:
Fix some regressions caused by 3.7 refactors: using a count with an altmeta
meta key command, and specifying a count greater than 2 (this would not fail
exactly but would never repeat the requested command more than twice).

- Fix: get_count and altmeta
- Fix: command counts greater than 2


paxed:
Pushed.

PR #837, opened 2022-08-08
Refer to non-damaging gas cloud as "steam"

copperwater:
You can create steam clouds that are gray and deal 0 damage when zapping
fire at or over water sources. If you happen to be in the cloud, it
produced the "You are enveloped in a cloud of noxious gas!" line, and
steam isn't noxious.

Other uses of "gas cloud" could have potentially been changed, but I
left them alone.


paxed:
Pushed.

PR #839, opened 2022-08-10
Improve feedback for trying to bribe the watch

entrez:
Members of the watch can never be bribed, no matter how much gold is
thrown at them, but when a bribe was attempted they would say "that's
not enough" (implying incorrectly some higher amount would be enough to
pacify them).  Also, an already-peaceful mercenary would take on an
unusually aggressive tone when given any amount of gold, giving the
player the same message to "beat it" as when a bribe succeeds in
pacifying them.

Adjust the mercenary bribery code a bit so that a more friendly message
is given if the monster is already peaceful, and so that unbribeable
monsters don't imply they are just not being given enough gold.  I think
the actual effects of bribery shouldn't be affected by this commit, just
the messages/feedback.


paxed:
Pushed.

PR #840, opened 2022-08-10
Describe engulf attack by animal as "swallowing"

entrez:
Suggested by aosdict: instead of describing all gulp attacks as
"engulfing you", say "The <monster> swallows you whole" for purple worms
(or any other animal engulfer) to visibly differentiate between
engulfing and swallowing.


paxed:
Pushed.

PR #841, opened 2022-08-10
Fix: m-prefix looting a container

entrez:
Using #loot with the 'm' prefix would claim there was nothing to loot
even if the hero was standing on a box, as long as there was no adjacent
monster.  'm' prefix is an explicit request to select a direction so
make it work regardless of whether a monster is there.  Once I did that
I noticed that it was providing no feedback for using '.' or '>' as the
direction if no container was available, so adjust that to give the
"there is nothing here to loot" feedback.


paxed:
Pushed.

PR #845, opened 2022-08-11
Monster looting improvements

NullCGT:
- Implemented cursed bag of holding behavior for when monsters loot a cursed bag of holding.
- Allow monsters to loot containers at their location instead of picking them up.

I avoided adding behavior for monsters unlocking containers. It would not be difficult to add, but the number of times we would need to iterate through a monster's inventory for such a niche behavior was difficult for me to justify.

### DIFF
--- a/src/muse.c
+++ b/src/muse.c
@@ -1831,6 +1831,8 @@ find_misc(struct monst* mtmp)
                         }
                     }
     }
+    if (nohands(mdat))
+        return 0;
     if (OBJ_AT(mtmp->mx, mtmp->my) && !rn2(2)) {
         for (otmp = g.level.objects[mtmp->mx][mtmp->my]; otmp; otmp = otmp2) {
             otmp2 = otmp->nexthere;
@@ -1838,12 +1840,9 @@ find_misc(struct monst* mtmp)
                 && Has_contents(otmp) && !otmp->olocked && !otmp->otrapped) {
                 g.m.has_misc = MUSE_BAG;
                 g.m.misc = otmp;
-                return TRUE;
             }
         }
     }
-    if (nohands(mdat))
-        return 0;
 
     /* normally we would want to bracket a macro expansion containing
        'if' without matching 'else' with 'do { ... } while (0)' but we
@@ -2060,10 +2059,11 @@ mloot_container(
                 break; /* out of takeout_count loop */
         } /* can_carry */
     } /* takeout_count */
-    /* Monster rummaged through a bag but found nothing of interest, possibly because
-       some items vanished. */
+    /* Monster rummaged through a bag but found nothing of interest,possibly
+       because some items vanished. */
     if (!res && vismon) {
-        pline("%s %s inside %s.", Monnam(mon), mon->mcansee ? "looks" : "feels about", 
+        pline("%s %s inside %s.", Monnam(mon), 
+                                  mon->mcansee ? "looks" : "feels about",
                                   contnr_nam);
     }
     return res;
--- a/src/muse.c
+++ b/src/muse.c
@@ -1784,6 +1784,8 @@ boolean
 find_misc(struct monst* mtmp)
 {
     register struct obj *obj;
+    struct obj *otmp;
+    struct obj *otmp2;
     struct permonst *mdat = mtmp->data;
     coordxy x = mtmp->mx, y = mtmp->my;
     struct trap *t;
@@ -1829,6 +1831,16 @@ find_misc(struct monst* mtmp)
                         }
                     }
     }
+    if (OBJ_AT(mtmp->mx, mtmp->my) && !rn2(5)) {
+        for (otmp = g.level.objects[mtmp->mx][mtmp->my]; otmp; otmp = otmp2) {
+            otmp2 = otmp->nexthere;
+            if (Is_container(otmp) && otmp->otyp != BAG_OF_TRICKS && !otmp->olocked) {
+                g.m.has_misc = MUSE_BAG;
+                g.m.misc = otmp;
+                return TRUE;
+            }
+        }
+    }
     if (nohands(mdat))
         return 0;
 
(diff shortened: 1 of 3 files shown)


src/muse.c
```diff
@@ -2055,6 +2060,12 @@ mloot_container(
                 break; /* out of takeout_count loop */
         } /* can_carry */
     } /* takeout_count */
+    /* Monster rummaged through a bag but found nothing of interest, possibly because
+       some items vanished. */
+    if (!res && vismon) {
+        pline("%s %s inside %s.", Monnam(mon), mon->mcansee ? "looks" : "feels about",
```
entrez: Formatting thing, this line and the above comment is longer than 80 characters

src/pickup.c
```diff
@@ -2608,7 +2607,7 @@ mbag_item_gone(int held, struct obj *item, boolean silent)
             You("%s %s disappear!", Blind ? "notice" : "see", doname(item));
     }
 
-    if (*u.ushops && (shkp = shop_keeper(*u.ushops)) != 0) {
+    if (*u.ushops && (shkp = shop_keeper(*u.ushops)) != 0 && user == &g.youmonst) {
```
entrez: Also longer than 80 characters

src/muse.c
```diff
@@ -1957,9 +1967,8 @@ mloot_container(
 
     if (!container || !Has_contents(container) || container->olocked)
         return res; /* 0 */
-    /* FIXME: handle cursed bag of holding */
     if (Is_mbag(container) && container->cursed)
-        return res; /* 0 */
+        boh_loss(mon, container, container->ocarry == mon);
```
entrez, line 1970: Is there anything to stop a monster from repeatedly looking into a bag of holding without "learning" it's cursed and losing all the items?

src/muse.c
```diff
@@ -1829,6 +1831,17 @@ find_misc(struct monst* mtmp)
...
                     }
     }
+    if (OBJ_AT(mtmp->mx, mtmp->my) && !rn2(2)) {
+        for (otmp = g.level.objects[mtmp->mx][mtmp->my]; otmp; otmp = otmp2) {
+            otmp2 = otmp->nexthere;
+            if (Is_container(otmp) && otmp->otyp != BAG_OF_TRICKS
+                && Has_contents(otmp) && !otmp->olocked && !otmp->otrapped) {
+                g.m.has_misc = MUSE_BAG;
+                g.m.misc = otmp;
+                return TRUE;
```
entrez: Why does a container being on the monster's current spot override any of the other misc items it can use half of the time?  It feels to me like maybe the `return TRUE` shouldn't be here, so that the other rules for things that the monster actually has a need to use can take precedence further down, and the container looting will only happen as a "well, nothing else interesting to do" thing.  That's basically how bags in inventory work, right?  They are the lowest precedence action in `find_misc`.  Is the idea that it should take precedence because the monster might move off the spot and lose the chance to open it?

src/muse.c
```diff
@@ -1829,6 +1831,17 @@ find_misc(struct monst* mtmp)
...
+            }
+        }
+    }
```
entrez, line 1845: Does placing this above the `if (nohands(mdat)) return 0;` mean that monsters without hands can still open a container on their location?

src/muse.c
NullCGT: Ugh, thanks. I'm not in my usual editor.

src/muse.c
NullCGT: No. It's a similar problem to monsters not being able to remember whether something is locked.

src/muse.c
```diff
@@ -1983,6 +1991,8 @@ mloot_container(
            from one item to the next */
         Strcpy(mpronounbuf, mhe(mon));
     }
+    Strcpy(contnr_nam, an(nearby ? xname(container)
+                                         : distant_name(container, xname)));
```
entrez, line 1995: The '?' and ':' should probably line up like they used to

src/muse.c
```diff
@@ -2055,6 +2059,13 @@ mloot_container(
                 break; /* out of takeout_count loop */
         } /* can_carry */
     } /* takeout_count */
+    /* Monster rummaged through a bag but found nothing of interest,possibly
+       because some items vanished. */
+    if (!res && vismon) {
+        pline("%s %s inside %s.", Monnam(mon), 
+                                  mon->mcansee ? "looks" : "feels about",
+                                  contnr_nam);
```
entrez, line 2067: The broken up parts should be aligned with the opening paren of pline, like
```c
        pline("%s %s inside %s.", Monnam(mon), 
              mon->mcansee ? "looks" : "feels about", contnr_nam);
```
or
```c
        pline("%s %s inside %s.", Monnam(mon), 
              mon->mcansee ? "looks" : "feels about",
              contnr_nam);
```
etc

src/pickup.c
```diff
@@ -2608,7 +2607,8 @@ mbag_item_gone(int held, struct obj *item, boolean silent)
             You("%s %s disappear!", Blind ? "notice" : "see", doname(item));
     }
 
-    if (*u.ushops && (shkp = shop_keeper(*u.ushops)) != 0) {
+    if (*u.ushops && (shkp = shop_keeper(*u.ushops)) != 0 
+                      && user == &g.youmonst) {
```
entrez, line 2611: Broken up part should be aligned with the opening paren, like
```c
    if (*u.ushops && (shkp = shop_keeper(*u.ushops)) != 0 
        && user == &g.youmonst) {
```

entrez (commented):
A couple more formatting nitpicks I wanted to point out.

pat-rankin:
This has significant ramifications for bones levels.  If the dead hero was carrying a bag of holding it will be cursed and on the floor.  Depending on where the new hero arrives on the level, any item-using monster in the bag's vicinity could have lots of time to repeatedly loot it before the new character eventually makes it to that location, potentially destroying a lot of items.  (That isn't necessarily a bad thing but most players probably won't like it.)

Right now bags and boxes with lots of stuff are protected by virtue of being too heavy to lift.

src/muse.c
copperwater: Since monsters implicitly know items as if they were fully identified, shouldn't they respond to a cursed bag of holding by dumping out all the contents so as not to lose items looking in each time?

copperwater:
Wrote this in response to a review comment before seeing it brought up in the main thread:

Since monsters implicitly know items as if they were fully identified, shouldn't they respond to a cursed bag of holding by dumping out all the contents (i.e. tipping it) so as not to lose items looking in each time?

Without something like that, yeah, I can't really get behind allowing monsters to loot bags on the ground.

PR #849, opened 2022-08-14
Remove obsolete find_skates function

copperwater:
find_skates was still in use for its one intended case, but objdescr_is
has been around for a few years now and can do just as good a job
without having to hardcode the first and last boots in objects[].


paxed:
Pushed.

PR #850, opened 2022-08-14
Level teleporters always print "momentarily disoriented" message

copperwater:
Removing the message "You are momentarily blinded by a flash of light"
and keeping the alternate message that was formerly shown only when
blind. There are three reasons:
1. This effect doesn't actually blind you.
2. This effect *does* momentarily disorient you (i.e. confuse you.)
3. This is the exact same message as the summoning/blinding effect of a
   magic trap, and it's easy to conflate the two.

### DIFF
--- a/src/teleport.c
+++ b/src/teleport.c
@@ -1191,10 +1191,7 @@ level_tele_trap(struct trap* trap, unsigned int trflags)
         You_feel("a wrenching sensation.");
         return;
     }
-    if (!Blind)
-        You("are momentarily blinded by a flash of light.");
-    else
-        You("are momentarily disoriented.");
+    You("are momentarily disoriented.");
     deltrap(trap);
     newsym(u.ux, u.uy); /* get rid of trap symbol */
     level_tele();


entrez:
In my opinion "momentarily" is suggestive of an effect that's true only for an instant, too fast to produce a status effect, and then passes.  For that reason I think the "you are momentarily blinded by a flash of light" actually fits better than "you are momentarily disoriented", for the opposite reason than you -- the first one could be true, but the second one is definitely not true since the effect isn't "momentary".  I think if this were to happen it should be rephrased to eliminate the word "momentarily".

copperwater:
It only confuses you for 3 turns. I would call that momentary.

entrez:
The message suggests to me that the effect has already passed by the time it is printed (otherwise how is it known that it will be only momentary?), and that also seems to be how it was intended since when it was written neither message referred to an actual status effect.

copperwater:
Why then does "You are momentarily blinded by a flash of light!" from a magic trap blind you for 10-15 turns? :-)

entrez:
That also seems wrong to me -- maybe it's just me, though!

PR #852, opened 2022-08-14
Fix a couple latent bugs with initial rings/other charged items

copperwater:
These bugs are currently latent only because no role's starting
inventory happens to satisfy the conditions under which it would show
up. This commit contains the xNetHack fixes for them.

Bug 1: if a role received a specific charged ring (versus a random
ring), it would be possible for that ring's charge to be <= 0, since the
code currently only caught this case for random rings. Move that clause
outside the if (UNDEF_TYP) clause.

Bug 2: non-ring oc_charged items of UNDEF_TYP that randomly rolled a
starting charge/enchantment <= 0 would always be enchanted, sometimes
very highly, because their enchantment was getting set to rne(3) by
hitting this case, which is only intended for rings as a comment
indicates. In particular, Archeologists' starting pick-axe would hit
this case after moving it out of the UNDEF_TYP block, but it would also
apply to any role receiving a random tool or weapon-tool at the start
of the game.

### DIFF
--- a/src/u_init.c
+++ b/src/u_init.c
@@ -1050,10 +1050,6 @@ ini_inv(struct trobj *trop)
                     break;
             }
 
-            /* Don't start with +0 or negative rings */
-            if (objects[otyp].oc_charged && obj->spe <= 0)
-                obj->spe = rne(3);
-
             /* Heavily relies on the fact that 1) we create wands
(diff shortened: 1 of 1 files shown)


paxed:
Committed.

PR #855, opened 2022-08-17
remove `GCC_WARN`

argrath:
It's unlikely to use gcc without enabling warnings, so:
 * substitute `GCC_WARN` to `__GNUC__` in sources
 * remove them from hints


argrath:
Add a change in comments and force-pushed.

entrez:
`GCC_WARN` is not really tied to using GCC right now: it is enabled by 
default on \*nix systems regardless of the compiler used.  This would
change preprocessor output for people building with llvm/clang on OSX,
for example.  I don't know if `GCC_WARN` is necessary across the board on
UNIX or whether it was really meant to be GCC-specific, but that seems
worth investigating before replacing it with `__GNUC__`.

pat-rankin:
GCC_WARN was originally called GNU_LINT which caused confusion because it seemed to be indicating the existence of a tool that no one ever saw.  At some point (many years ago) it got renamed to better reflect that it was used with code intended to pacify 'gcc -Wall' (and 'gcc -Wall -W', now 'gcc -Wall -Wextra').

Like traditional lint, it tends to lead to the bad practice of introducing snippets of fake code to mollify the testing tool, with the corresponding real code not getting whatever testing is being done.

In nethack's case, it's usually used to add initialization to avoid "might be used uninitialized" warnings.  That warning is often inaccurate (discounting the wishy-washy "might be") when compiling with optimization suppressed.

Over time, lots of the GCC_WARN conditionalized code has become unconditional so that the compiler is always operating on the same code, and newer code has lots of superfluous initializations.  Unfortunately that runs the risk of other compilers being smart enough to complain that an initialization is redundant.  So blanket unconditionalization has the potential to introduce its own problems.

I don't know whether replacing GCC_WARN with __GNUC__ is a good idea.  Ideally there shouldn't be any '#if __GNUC__' constructs in src/*.c except for few that were introduced to workaround old compiler bugs.  [Particularly the infamous "strength reduction" bug affecting optimized loop code generated for little-endian CPUs.  That was present across many early versions of gcc and the cause of mystery bugs for folks who enabled optimization, being hidden from anyone trying to debug them without optimization.]

argrath:
Recent commits reduced the use of `GCC_WARN`, so I'll take down this PR and submit a new one.

PR #858, opened 2022-08-24
Describe engulf attacks a bit more consistently

entrez:
Use verbiage for mon vs mon and hero (mostly hero) engulf attacks that
matches recent changes to monster vs hero engulf attacks more closely
(e.g. "swallows whole" instead of "engulfs" for purple worm, other
changes in b07fe59...).  Also ensure non-AD_DGST engulf attacks
(e.g. from revamped trapper or lurker above polyforms) aren't treated as
"eating" (or as involving "debris").

Also change the enfolds and digests macros so they produce booleans
rather than attack pointers (I got a compiler warning about casting
struct attack * to boolean when I did 'boolean b = digests(ptr);').

### DIFF
--- a/include/mondata.h
+++ b/include/mondata.h
@@ -86,8 +86,10 @@
 #define is_wooden(ptr) ((ptr) == &mons[PM_WOOD_GOLEM])
 #define thick_skinned(ptr) (((ptr)->mflags1 & M1_THICK_HIDE) != 0L)
 #define hug_throttles(ptr) ((ptr) == &mons[PM_ROPE_GOLEM])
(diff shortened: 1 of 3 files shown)


entrez:
> Also change the enfolds and digests macros so they produce booleans 
> rather than attack pointers (I got a compiler warning about casting
> struct attack * to boolean when I did 'boolean b = digests(ptr);').

I guess dmgtype_fromattack is only ever used for "does this exist or 
not", so instead of changing digests and enfolds the function itself 
could also just be updated to return a boolean.  Unless there's some 
reason to keep it as struct attack * that I'm overlooking.

PR #862, opened 2022-08-28
Enhance feedback from detect unseen

copperwater:
I was never too happy with how this was a silent effect that required
you to watch the map to see if anything changed. It might count as
an accessibility issue as well, not sure.

This change adds specific feedback for all the possible things that
might get revealed by detecting unseen. If you reveal a secret door or a
trap, you now get a message indicating that.


paxed:
Merged.

PR #865, opened 2022-08-31
Fix inaccessible spaces and shop backdoors in Bustling Town

copperwater:
Perennial problem: since Bustling Town consists of a fixed map which was
shorter than the moveable area overlaid onto a cavern fill, sometimes
the cavern fill made spaces above or below the walled edges of the town,
which then got cut off when the town was placed. Since one of the valid
ways to generate a way out of an inaccessible space is to create a door
connecting it to an accessible space, this meant rooms in Minetown,
including shops, could get a second door leading into that tiny space.
And if it was a shop, the shopkeeper could not block exit from the
second door.

This fixes that issue by expanding the map vertically so that it will
overwrite the whole cavern fill in the town area of the map segment
where it might create cut-off spaces like this. (Not the whole area -
since we now have the 'x' mapchar to leave existing terrain in place,
areas adjoining open space where inaccessible pockets won't get
created can retain any existing fill to keep the town from being exactly
the same every time).

### DIFF
--- a/dat/minetn-6.lua
+++ b/dat/minetn-6.lua
@@ -11,55 +11,62 @@ des.level_flags("mazelevel", "inaccessibles")
 
 des.level_init({ style="mines", fg=".", bg="-", smoothed=true, joined=true,lit=1,walled=true })
 
+-- Map extends the full height of the playable area in order to prevent any of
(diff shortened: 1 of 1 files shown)


paxed:
Committed

PR #866, opened 2022-09-01
Make des.mineralize use default probabilities

copperwater:
Calling des.mineralize() with no arguments was equivalent to calling it
and manually specifying gem_prob = 0, gold_prob = 0, etc. Which meant
that no mineralization would actually happen.

Instead, make this match the intuitive behavior, and pass in -1
probabilities as defaults -- which the mineralize() function interprets
as the caller wanting to use the standard probabilities for a level of
that depth, as if it were not a special level.

This change does not affect any special level files since des.mineralize
is not currently used in any of them.

### DIFF
--- a/src/sp_lev.c
+++ b/src/sp_lev.c
@@ -3782,10 +3782,11 @@ lspo_mineralize(lua_State *L)
     create_des_coder();
 
     lcheck_param_table(L);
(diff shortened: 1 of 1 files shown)


paxed:
Committed

PR #867, opened 2022-09-03
Move levitation potion headache false rumor into true rumors

jpsenior:
Quaffing a cursed potion of levitation does indeed cause a headache, up to d10 damage.  Assuming one was quaffing 'many' unidentified potions, it's possible one of them is indeed cursed when the player hits their head on the ceiling.

https://github.com/NetHack/NetHack/blob/f5a9901db1d9e7ebc97f4ad1fdffebb695ca2b79/src/potion.c#L1182

### DIFF
--- a/dat/rumors.fal
+++ b/dat/rumors.fal
@@ -289,7 +289,6 @@ They say that only big spenders carry gold.
 They say that orc shamans are healthy, wealthy and wise.
 They say that playing NetHack is like walking into a death trap.
 They say that problem breathing is best treated by a proper diet.
-They say that quaffing many potions of levitation can give you a headache.
 They say that queen bees get that way by eating royal jelly.
 They say that reading a scare monster scroll is the same as saying Elbereth.
 They say that real hackers always are controlled.
--- a/dat/rumors.tru
+++ b/dat/rumors.tru
@@ -367,3 +367,4 @@ You're going into the morgue at midnight???
 Your dog knows what to eat; maybe you should take lessons.
 Zap yourself and see what happens...
 Zapping a wand of undead turning might bring your dog back to life.
+They say that quaffing many potions of levitation can give you a headache.


entrez:
I think there are quite a few rumors in rumors.fal that aren't strictly false but just aren't directly helpful or useful to a player.

jpsenior:
> I think there are quite a few rumors in rumors.fal that aren't strictly false but just aren't directly helpful or useful to a player.

@entrez I can go through each rumor if we think there are more that may need candidates for moving, but to be honest this is the only one I considered in this commit. Would you like me to evaluate each rumor? I do understand what you mean by 'strictly false' in that some can be ambiguous, but in my opinion this one is wrong.

This actually came up in context of live-watching a player who encountered this rumor and some conversations came up on the truthiness of it.

copperwater:
In xNetHack, the rule of thumb for true rumors is that they must be "helpful to a player encountering the game for the first time", though they don't have to be straightforward. I would consider this potion of levitation rumor to qualify as true by that standard. 

(Meanwhile, false rumors allow "pretty much anything NetHack-related, as long as it's not actual valid advice. Cryptically hinting at mechanics that don't exist, recommending that the player do useless things, even total tangents, as long as they make sense in context.")

pat-rankin:
I doubt if this change is going to be incorporated.  The rumor implies that the number of potions of levitation you drink can have an effect other than levitating and that is not helpful or useful.  Contorting it to mean "if you drink a lot of them, some are likely to be cursed" isn't either.  If the first one you drink happens to be cursed, does "one" count as "many"?

What really ought to be done--I'm not suggesting that you do it, and I'm certainly not going to do it--is to review all the rumors looking for ones that are classified as True but are no longer True or classified as False but are no longer False after umpteen years of gameplay changes.  That would be tough even when the meanings are straightforward and they often aren't.

PR #869, opened 2022-09-08
Spelling Fixes and Standardizations

klorpa:


amateurhourrrr:
'behaviour' isn't a typo, it's the conventional spelling everywhere outside the US
'iff' may not be a typo in some of those locations, as it's shorthand for if-and-only-if

targeted vs targetted isn't very important even if the former is the dominant choice but fair enough on that one

nhmall:
Thanks for bringing attention to these. A subset of the changes in the pull request have been manually applied and are in the pipeline.

The comment above by @gebulmer regarding "behaviour" vs "behavior" is valid, but decades ago it was decided to settle on U.S. spellings in the NetHack sources just for consistency. It may have been the equivalent of a "coin toss." Quotations from non-U.S. sources in the database or tribute files are an exception. It is an ongoing challenge, for those of us not in the U.S., to get it "right."

```
behaviour  to behavior
speach     to speech
knowlege   to knowledge
incrments  to increments
stethscope to stethoscope
staiway    to stairway
arifact    to artifact
extracing  to extracting
targetting to targeting
```

The uses of "iff" were left alone because they appear to be quite deliberate on the part of the developer.
doc/Fixes* files from versions prior to 3.7 were also left alone.

@klorpa Were these found by eyeballing the sources, or via a tool or combination of tools?

PR #880, opened 2022-09-22
Hole/trap door destination fixes

entrez:
Attempted fixes for various issues stemming from 05761ba, mostly interactions
between holes/trapdoors and bones files: e.g. holes can deposit you backwards,
onto a higher level than you started on, or into the Sanctum prematurely (even
after applying ccaadaa).

- Fix: antigravity trap doors
- Have monsters' hole destination match the hero's
- Prevent hero from falling past end of dungeon
- Refine attempt to clamp trap door fall destination
- Apply dest. limit to monster trap door usage


paxed:
Pushed.

PR #881, opened 2022-09-28
Fix undefined behavior when exiting Curses

chasonr:
* When saving: curses_exit_nhwindows calls curses_uncurse_terminal, which calls endwin. curses_exit_nhwindows then calls raw_print, which calls more Curses functions after endwin has been called. Fix this by having curses_raw_print use puts if window_inited is false.

* When dying, quitting, etc.: really_done opens the "Goodbye" window, which refreshes the other windows when it closes. But the status window (and possibly the map and message windows) are gone by that point. The window pointers are properly NULLed, but the NULL is then passed to touchwin. Fix this by checking window pointers for NULL.

These bugs manifest as crashes on exit when the @Bill-Gray fork of PDCurses (https://github.com/Bill-Gray/PDCursesMod) is linked to a cross-compile to MS-DOS. This fork has the potential to offer a much-improved Curses port on DOS, but I also uncovered some bugs in PDCursesMod.

### DIFF
--- a/win/curses/cursmain.c
+++ b/win/curses/cursmain.c
@@ -855,7 +855,11 @@ curses_raw_print(const char *str)
 #ifdef PDCURSES
     /* WINDOW *win = curses_get_nhwin(MESSAGE_WIN); */
 
-    curses_message_win_puts(str, FALSE);
+    if (iflags.window_inited) {
+        curses_message_win_puts(str, FALSE);
+    } else {
+        puts(str);
+    }
 #else
     puts(str);
 #endif
--- a/win/curses/curswins.c
+++ b/win/curses/curswins.c
@@ -176,12 +176,18 @@ curses_refresh_nethack_windows(void)
         touchwin(stdscr);
         refresh();
     } else {
-        touchwin(status_window);
-        wnoutrefresh(status_window);
-        touchwin(map_window);
-        wnoutrefresh(map_window);
-        touchwin(message_window);
-        wnoutrefresh(message_window);
+        if (status_window != NULL) {
+            touchwin(status_window);
+            wnoutrefresh(status_window);
+        }
+        if (map_window != NULL) {
+            touchwin(map_window);
+            wnoutrefresh(map_window);
+        }
+        if (message_window != NULL) {
+            touchwin(message_window);
+            wnoutrefresh(message_window);
+        }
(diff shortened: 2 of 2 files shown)


Bill-Gray:
Relying on Curses functions after calling `endwin()` is somewhat risky.  The results are (mostly) undefined and/or implementation-specific.  Going with `puts()` is a good move.

PDCursesMod does abort (in debug compiles) via `assert()` if you use NULL pointers.  [This was a deliberate choice](https://github.com/Bill-Gray/PDCursesMod/commit/96f4984f9fb76a0546b71e3018e441224fe747ed) (commit message explains all).  I'd consider a `-DALLOW_NULL_POINTERS` compilation flag,  particularly in cases where the code using Curses can't be modified for some reason... but passing a NULL pointer to those functions _usually_ means a bug and shouldn't be silently ignored.  Though the Curses specification calls for us to do exactly that.

If you're seeing other bugs,  please do let me know.  (I should admit that I only occasionally test PDCursesMod with DOS,  sometimes with your quite nice DOSVGA port,  sometimes with the "plain vanilla" DOS port.  Bugs are more likely to slide through in DOS than they are elsewhere.)

chasonr:
Two bugs I observed in PDCursesMod.

The first one seems specific to DOSVGA. If PDC_WIDE is not used, window borders draw with the wrong characters. I have not tracked this one down.

Using PDC_WIDE fixes the borders; and, as the NetHack in progress seems to be trying to support Unicode map symbols, it's probably the way to go. But it reveals another bug, which is also present in the original PDCurses. wgetnstr is broken in this configuration if wint_t is larger than wchar_t. For NetHack, this bug manifests right at the beginning: when you type the name of your new hero, only the first character gets through. I have issued pull requests to PDCurses and to PDCursesMod for this bug.

Bill-Gray:
Just compiled and ran `testcurs` with DOSVGA in non-wide mode,  and I see the same 'broken' box characters in the ACS Test section (image below).  Will investigate.  [Edit : realized it was actually a rather dumb mistake on my part and fixed it.]

I'm also seeing that when I set PDC_FONT=any of the various PSF fonts,  I get garbage.  Will investigate that as well.  I know it worked Once Upon A Time;  some bisection will probably turn up a commit that broke that capability.
z

PR #882, opened 2022-09-28
A couple genocide prompt tweaks

entrez:
Treat hitting Enter on an empty line consistently between genocide and class
genocide prompts, and use flags.cmdassist to decide whether to show additional
help.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -2463,7 +2463,7 @@ do_class_genocide(void)
 {
     int i, j, immunecnt, gonecnt, goodcnt, class, feel_dead = 0;
     int ll_done = 0;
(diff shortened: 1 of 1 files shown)


pat-rankin:
I think input handling for class genocide should have been changed to match single genocide rather than the other way around.

A loop which gets input shouldn't be allowed to run an unlimited number of times since a hangup signal at the wrong moment could result in consuming all available CPU or in the process being killed off before it reaches a point where it can perform an emergency save.  NetHack might successfully switch to the hangup interface so deliver ESC as the response to any input request, but that's not necessarily reliable.

entrez:
That makes sense, I'll update the branch to flip-flop it.

Issue #886, opened 2022-09-30
CURSES_UNICODE vs IBMgraphics


entrez:
The various IBMgraphics symsets no longer work on Hardfought when using the curses windowport, instead showing up like this (this is Minetown):

    ┌────────────────────────────────────────────────────────────────────────────────┐
    │                                                                                │
    │                                                                                │
    │                                                                                │
    │                   M-0                             M                            │
    │                   MM-3             %              M                            │
    │                   MM-3     Me    M-?    +     M-? M                            │
    │                   MM-3 M/3(M  M-3M-M-3 M  M-3 M-3                              │
    │                   MM-3 M@++M  M-3M-M-~%M  M(( M-3 M                            │
    │                   MM-3 M)3oM G     M-@ M-AM-4 M-3 M                            │
    │                   MM-3 M-M-M+A            M-@M-~I M                            │
    │                   MM-3   f   B               %    M                            │
    │            M      MM-3             M-~ M-? )[     M                            │
    │    M-3     M    MMMM-~@          M-3   M-3    (   M                            │
    │    M-3      MMMMM MM-3   M-~M-?  M-3   M-3  M-M-? M                            │
    │    M-3  <  M  MMMMMM-3  M-4   MM)M-3[    MM-4+M-3 M                            │
    │    M-@             M-M  M@((M-3M-M-3   M-+M-? M-3                              │
    │                    M-+  M/((M-3M-M-@   M-4M]3 M-3 M                            │
    │                    M-M-@M-A M-A        M-@M-A     M                            │
    │                    M-3                             M                           │
    │                    M-@                                                         │
    │                                                                                │
    └────────────────────────────────────────────────────────────────────────────────┘

Even when using the normal curses symset, some suspiciously-similar looking things pop up under certain circumstances, e.g. farlooking a floor space:

    M-~        a doorway or the floor of a room or the dark part of a room or ice (dark part of a room)

K2 did some experimentation and found out that this is triggered by the CURSES_UNICODE define introduced to support ENHANCED_SYMBOLS in curses.  Without that define it doesn't happen (but the Enhanced1 symset doesn't work in curses, of course).  The issue doesn't show up in TTY either way.

On my local system (running OSX) I don't seem to have this problem, but in addition to Hardfought it's been reproduced on another system running Ubuntu 22.04.1 LTS.  Not sure if this is an issue with NetHack or with the ncursesw library on Ubuntu.  K2 can probably provide more details as required.

[...]

PR #889, opened 2022-10-03
Provide Unicode support to DOS Curses port

chasonr:
These changes offer support for Unicode in the DOS Curses port, and also
close issue #886.

1) Offer PDCursesMod as an alternative implementation of PDCurses for the DOS port.

PDCursesMod is available at https://github.com/Bill-Gray/PDCursesMod . This fork of PDCurses offers an implementation that works through the VGA graphical modes, offering greater resolutions and support for Unicode-capable fonts.

This change loads PDCursesMod at lib/pdcursesmod (or lib/pdcurmod if building natively on DOS). If WANT_DOSVGA=1 is specified to the make command, the build will look for PDCursesMod and its "dosvga" port. If not, the build will look for mainline PDCurses as before.

TODO: PDCursesMod also offers a "wingui" port that can work with Windows.

2) Provide the UTF-32 code point to the windowing code

A field utf32ch is added to struct unicode_representation. This gives the Unicode code point from the symbols file. Curses and many other user interface libraries more easily work with this code point than with UTF-8.

TODO: Use this to prove Unicode support for X11, Windows graphical and the DOS TTY interface.

3) Call the wide character function to display a Unicode map symbol. When H_IBM is in effect, convert from IBM437 to Unicode.

4) Provide external fonts for the DOS Curses build

These fonts are derived from the Terminus font project, hosted at https://sourceforge.net/projects/terminus-font/ . A program, makefont.lua, is provided to convert the BDF files to the PSF form that PDCursesMod expects. Because building the fonts from the cross-compile proved overly complicated, I have provided the compiled fonts in binary form.

TODO: The DOS TTY port could use all of these fonts in VESA mode, and the 8x16 fonts in 16-color mode.

5) Fix the native build for DOS, and make it accept WANT_WIN_CURSES and WANT_DOSVGA as the cross-build does.


nhmall:
Perhaps this patch can work around your cross-compile font conversion issues and avoid adding the binary fonts to the repo?
https://gist.github.com/nhmall/34317ae9fc7b2c8e94c297a94a9b795b

Note: sys/unix/setup.sh will have to be used with the linux.370 or macOS.370 hints file to distribute updated Makefiles after applying that patch.

chasonr:
Your patch did most of the heavy lifting. I did have to add some rules so that make would actually build the fonts.

I confirmed that the fonts build if, and only if, WANT_DOSVGA is specified. If you accept this change, I'll be around with another one for the TTY mode; that change will use the fonts always.

nhmall:
Another follow-up suggestion with accompanying patch.

The existing terminus-fonts are not being actively altered or maintained as part of the NetHack source tree, and they seem to be copied as-is from the source. My suggestion is to hoist the fetch of those outside the NetHack build process and place them in a lib subfolder prior to starting the build of NetHack. 

To that end, the patch below adds the fetch of the bdf font files to the cross-compiler fetch procedure, but they would have to be placed into the lib subfolder through external means prior to starting a native DOS build.

Having them outside the NetHack distribution avoids ongoing pass-through storage of the bdf files as part of NetHack. It does add on-going maintenance to the process if the source location changes, but that seems like an acceptable trade-off at this time.

Patch:
https://gist.github.com/nhmall/51b5d078bc9b2f30de89e5ef70ae6ed9

The sys/msdos/fonts/*.bdf files would also be removed, but I did not inflate the above patch with that step.

chasonr:
A couple of things I did differently: first, for the native build, the font package is in lib/terminus for 8.3 compliance; second, I put back the section to fetch PDCursesMod, and edited Install.dos to clarify that this is used when the Unicode build is selected. Downloading both PDCurses and PDCursesMod is less than ideal, but doing differently requires either using PDCursesMod with the non-Unicode build or adding a parameter to fetch-cross-compiler.sh.

I suppose the sys/msdos/fonts directory really should go away, but it makes a convenient place for the compiled fonts. I'll add a readme to that effect.

PR #902, opened 2022-10-16
The word 'Current' is unnecessary here (and in general).

zomGreg:
On the enlightenment output, the word Current is superfluous. In general, the word Current is unnecessary.

### DIFF
--- a/src/insight.c
+++ b/src/insight.c
@@ -898,7 +898,7 @@ status_enlightenment(int mode, int final)
      *     should be discernible to the hero hence to the player)
     \*/
     enlght_out(""); /* separator after title or characteristics */
-    enlght_out(final ? "Final Status:" : "Current Status:");
+    enlght_out(final ? "Final Status:" : "Status:");
 
     Strcpy(youtoo, You_);
     /* not a traditional status but inherently obvious to player; more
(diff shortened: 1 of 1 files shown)


entrez:
It doesn't seem completely superfluous to me.  It indicates to some degree that your attributes can change and aren't fixed over the course of the game.  That is more self-evident for "status" but not as much for "attributes" I think.

zomGreg:
When invoking or otherwise attaining an enlightenment event, it is a snapshot of the stats. The mutability of stats is immaterial. Also, if using the "stats can change" argument then the output of ^x needs to be fixed to *add* Current to the Background and Basic sections, since those can both change. The "things can change" argument for Current quickly reaches absurdity, as you'd have to add it to every stat that can change for the sake of consistency.

We don't say "Current turn count", because everyone knows that number will change.

On the Final attributes page/summary, a stronger argument for current can be made since your final stats may be different than your max stats, so the reference is meaningful. Although I like the way it is presented now, ex: Your constitution was 18 (limit:18). 

As argued in #nethack on IRC, this is a minor issue that is important only insofar as anyone cares about "tightening" up the messages used in nethack. 

Huge fan of the game. Love the work you're all doing.

Umbire:
I would think it's necessary because of stat loss that can be reversed, isn't it?

zomGreg:
No. The display is a moment in time. Almost all numbers/attributes in a game can change (I can't think of a counter example except maybe role). We don't say currently chaotic. It's just chaotic, even though a helm of opposite alignment is a possibility. Whether or not an attribute/value can change is immaterial.

Removing 'currently' tightens up the display and language used in the game. Makes it cleaner, neater, and shows good attention to detail.

PR #904, opened 2022-10-19
Fix: duplicate invlet from throwing obj with count

entrez:
Specifying a count of 1 when throwing an object could leave you with two
stacks sharing one inventory letter.  The second stack gets split off
when the player specifies a count (e.g. 't1o'), but keeps its original
invlet.  Some early returns, like trying to throw at yourself with '.',
could fail to unsplit the stack.  Theoretically, specifying multiple
items to multishot and then failing to throw them all could also leave a
partial stack; I don't think this is actually possible right now with
't' but I tried to make sure it won't become a problem if greater counts
than 1 are ever allowed.

The fix doesn't affect 'f', which can be a combined "create a quiver
stack and throw" action and doesn't have the issue with duping invlets.
Specifying a count to split off a new quiver stack with 'f' shouldn't be
reverted if the throwing fails or only part of the stack is thrown,
because the newly created stack may be intended for continued use as the
quiver in future turns.  This slightly changes the behavior of the
existing unsplit when cancelling the throw (which previously unsplit the
newly created quiver and quivered the entire parent stack), but I think
this actually makes more sense -- the player only declined to throw the
new stack, not to create it (as if they canceled earlier in the action).

I routed a couple early returns through the stack unsplitting that
shouldn't actually need it (like Mjollnir and welded items) for
consistency's sake; I don't think it hurts anything.

### DIFF
--- a/src/dothrow.c
+++ b/src/dothrow.c
@@ -85,20 +85,14 @@ throw_obj(struct obj *obj, int shotlimit)
     schar skill;
     long wep_mask;
     boolean twoweap, weakmultishot;
+    int res = ECMD_TIME;
+    struct obj_split save_osplit = g.context.objsplit;
 
     /* ask "in what direction?" */
     if (!getdir((char *) 0)) {
-        /* No direction specified, so cancel the throw;
-         * might need to undo an object split.
-         * We used to use freeinv(obj),addinv(obj) here, but that can
-         * merge obj into another stack--usually quiver--even if it hadn't
-         * been split from there (possibly triggering a panic in addinv),
-         * and freeinv+addinv potentially has other side-effects.
-         */
-        if (obj->o_id == g.context.objsplit.parent_oid
-            || obj->o_id == g.context.objsplit.child_oid)
-            (void) unsplitobj(obj);
-        return ECMD_CANCEL; /* no time passes */
+        /* No direction specified, so cancel the throw */
+        res = ECMD_CANCEL; /* no time passes */
+        goto unsplit_stack;
     }
 
     /*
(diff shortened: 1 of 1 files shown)


entrez:
I don't know if there's a neater way to handle this; I thought about 
just putting the unsplit into dothrow (I don't know if that works 
because obj can potentially be freed...), and removing the ability to 
specify a count in dothrow at all (but you need it for throwing specific 
amounts of gold, even though it's limited to 1 for other types of 
objects).  Maybe the devteam will think of something better than this.

entrez:
Another way to fix it that is simpler, but would completely ignore the count (so you could end up multishotting even if you enter 't1o'), might be something like this:

```diff
diff --git a/src/invent.c b/src/invent.c
index 01794c024..c8d2df629 100644
--- a/src/invent.c
+++ b/src/invent.c
@@ -1795,10 +1795,14 @@ getobj(
                "ready" or "fire" and this restriction doesn't apply */
             if (cnt == 0)
                 return (struct obj *) 0;
-            if (cnt > 1 && (ilet != def_oc_syms[COIN_CLASS].sym
-                && !(otmp && otmp->oclass == COIN_CLASS))) {
+            if (ilet == def_oc_syms[COIN_CLASS].sym
+                || (otmp && otmp->oclass == COIN_CLASS)) {
+                /* respect count for throwing gold */
+            } else if (cnt > 1) {
                 You("can only throw one item at a time.");
                 continue;
+            } else {
+                cntgiven = FALSE; /* ignore count of 1 */
             }
         }
         g.context.botl = 1; /* May have changed the amount of money */
```

This would also leave 'f' with quiver selection inconsistent, in that canceling when selecting a direction would unsplit the stack (and make the parent stack the new quiver), but other early returns wouldn't.  So I think there are at least some ways the original patch is a better approach than this, despite being a bit more complicated.

PR #906, opened 2022-10-22
Carrying a blessed loadstone bestows steadfastness

Theyflower:
I was poking around to read the source code of the new knockback change and mentioned to some people that holding Giantslayer prevents the wielder from being knocked back. Many of them mused possessing a loadstone should also grant steafastness. This is my attempt at implementing such a change.

### DIFF
--- a/src/uhitm.c
+++ b/src/uhitm.c
@@ -4614,8 +4614,19 @@ m_is_steadfast(struct monst *mtmp)
              : (is_flyer(mtmp->data) || is_floater(mtmp->data)))
         return FALSE;
 
+    /* holding giantslayer makes you steadfast */
     if (is_art(otmp, ART_GIANTSLAYER))
         return TRUE;
+
+    /* having a blessed loadstone in your inventory makes you steadfast */
+    //there are a few first person comments written by deadnoob/Theyflower
+    /*i am choosing to iterate over the knockback victim's inventory to check for a loadstone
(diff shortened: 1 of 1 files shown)


src/uhitm.c
```diff
@@ -4614,8 +4614,19 @@ m_is_steadfast(struct monst *mtmp)
...
         return TRUE;
+
+    /* having a blessed loadstone in your inventory makes you steadfast */
+    //there are a few first person comments written by deadnoob/Theyflower
+    /*i am choosing to iterate over the knockback victim's inventory to check for a loadstone
+    i am not sure if this actually does that, but I'm pretty sure it should*/
+    otmp = is_u ? g.invent : mtmp.minvent; //we already used otmp so i'm just going to reuse it. this SHOULD make it have "mtmp"'s inventory as something I can iterate over... I think
+    //this should iterate through the entire inventory, i see a similar thing done elsewhere in the code, i'm guessing nobj is "next object" in a linked list
+    for(otmp;otmp;otmp->nobj)
+        if ((obj->otyp == LOADSTONE || obj->blessed))
```
vultur-cadens: Should be `&&` instead of `||` here, otherwise this means that having a loadstone OR any blessed object makes you steadfast.

Theyflower:
```suggestion
        if ((obj->otyp == LOADSTONE && obj->blessed))
```
Good catch.

PR #917, opened 2022-10-31
Message for giving unnameable monster blank name

entrez:
The message for trying to (re)name an unnameable monster was weird when
the player entered a blank name (a name consisting of only spaces), in
an attempt to delete the monster's existing name rather than give it a
new one.  Several of the normal messages looked incomplete due to using
the blank string as the "new name" (e.g., "I'm Feyfer, not ."), and the
one which didn't include the name still seemed a little off ("calling
names" is almost the opposite of what the player is doing).  Add a new
message for attempting to erase the name of one of the special
unnameable monsters, e.g. "Juiblex would rather keep his existing name"
or "The Oracle would rather keep her existing title".

I also noticed that the message for trying to give an unrenameable
monster the name it already has (e.g. trying to name Death "Death") was
revealing the genders of the Riders with personal pronouns.  This is
avoided elsewhere (e.g. using "the way" in mdisplacem, "its" baked
into the message in mhitm_ad_deth), so I also added a slightly rephrased
alternate message for Riders which avoids any pronouns.  For the
unnaming message, on the other hand, I just used "its" for the Riders,
since I couldn't think of a way to phrase it that avoided pronouns
entirely.

### DIFF
--- a/src/do_name.c
+++ b/src/do_name.c
@@ -1162,24 +1162,38 @@ christen_monst(struct monst *mtmp, const char *name)
 }
 
 /* check whether user-supplied name matches or nearly matches an unnameable
-   monster's name; if so, give alternate reject message for do_mgivenname() */
+   monster's name, or is an attempt to delete the monster's name; if so, give
+   alternate reject message for do_mgivenname() */
 static boolean
 alreadynamed(struct monst *mtmp, char *monnambuf, char *usrbuf)
 {
     char pronounbuf[10], *p;
 
-    if (fuzzymatch(usrbuf, monnambuf, " -_", TRUE)
-        /* catch trying to name "the Oracle" as "Oracle" */
-        || (!strncmpi(monnambuf, "the ", 4)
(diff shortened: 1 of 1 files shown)


entrez:
I wasn't sure if a message should only be printed if the monster 
actually has a given name, and not for monsters where removing the name 
would be a no-op.

It felt a little tricky because some have a "name" but don't use 
mon->mextra->mgivenname (like shopkeepers -- erasing their mgivenname 
would be a no-op, but that's not obvious to the player), others have a 
"name" but it's really the name of the monster...  And if you extend it 
to all those categories, it may seem strange that nothing is printed for 
unnameable mons that don't fit into any of those categories.  So that's 
the reason I settled upon the possible "title" distinction, and print 
some message for all special monsters that can't be named, but I know a 
different decision could be made there.

Edit: Also, I guess the other thing is that all these monsters currently 
do print a message if you try to name them with a blank name.

PR #920, opened 2022-11-03
Make paralysis intefere with seduction attempts

entrez:
tinklebear on IRC noticed that a hero paralyzed by a floating eye was
still "charmed" and capable of "removing her armor" as part of a nymph's
theft attack.  The same thing was true of foocubus seduction: a
paralyzed hero was still able to respond to the questions about whether
particular pieces of armor should be removed (and also do whatever else
may be involved in a successful attack...).

I think paralysis should prevent both those things.  Nymph theft will
still work, unless she needs the hero's active cooperation in removing a
bulky piece of armor.  Foocubus attacks will be prevented entirely by
paralysis, making it interfere like unconsciousness already does.

Apply a similar constraint to hero vs monster seduction, as well.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -2657,6 +2657,7 @@ extern int num_spells(void);
 extern long somegold(long);
 extern void stealgold(struct monst *);
 extern void thiefdead(void);
+extern boolean unresponsive(void);
 extern void remove_worn_item(struct obj *, boolean);
 extern int steal(struct monst *, char *);
 extern int mpickobj(struct monst *, struct obj *);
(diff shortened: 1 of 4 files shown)


entrez:
This would affect nymph theft when the hero is fainting or otherwise 
unconscious.  Currently, nymphs are able to instantly rouse the hero and 
set him to work taking off his own armor.  I made it prevent the hero 
from assisting in the theft instead, so that all "hero is unresponsive" 
conditions would be handled the same way.  This is also more similar to 
how foocubus seduction treats unconsciousness (it already prevented that 
entirely), but it's a decision that could be made differently.

Just using 'g.multi < 0' instead of this unresponsive() function seemed
tempting, but I think there are some scenarios that would catch (like 
being entangled in a spider web, or fumbling) which I don't think should
really prevent you from responding to a nymph or foocubus.  But that's
another call that could be made differently, if the simplicity might
be worth overapplying the rule in some rare circumstances.

PR #921, opened 2022-11-03
Old software compatibility

chasonr:
* Don't use __attribute__((printf(...))) with function pointers unless GCC is at least 3.1. Older GCCs understand it with functions only.
* Older X servers don't understand "Fuchsia", causing impossibles. Use "magenta", which is the same as "Fuchsia" in the current Xorg.

This pushes back the minimum GCC version to 3.0. NetHack compiled with these changes works with XFree86 4.2.0 and possibly earlier versions as well.

### DIFF
--- a/src/mkobj.c
+++ b/src/mkobj.c
@@ -761,7 +761,7 @@ static const char dknowns[] = { WAND_CLASS,   RING_CLASS, POTION_CLASS,
 void
 clear_dknown(struct obj *obj)
 {
-    obj->dknown = index(dknowns, obj->oclass) ? 0 : 1;
+    obj->dknown = strchr(dknowns, obj->oclass) ? 0 : 1;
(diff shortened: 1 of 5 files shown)


src/mkobj.c
```diff
@@ -761,7 +761,7 @@ static const char dknowns[] = { WAND_CLASS,   RING_CLASS, POTION_CLASS,
 void
 clear_dknown(struct obj *obj)
 {
-    obj->dknown = strchr(dknowns, obj->oclass) ? 0 : 1;
+    obj->dknown = index(dknowns, obj->oclass) ? 0 : 1;
```
entrez: Did you mean to revert this one part of 99a93fe50?

chasonr: That must have got in by mistake.

PR #928, opened 2022-11-18
remove the code to silence lint

argrath:
Warning facilities on recent compilers are incredibly improved, so the code to silence "good-old" lint is much less sense.

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -39,18 +39,6 @@ extern void panic(const char *, ...);
 long *
 alloc(unsigned int lth)
 {
-#ifdef LINT
-    /*
-     * a ridiculous definition, suppressing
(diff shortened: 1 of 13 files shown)


entrez:
Doesn't it enable people to build on older compilers more easily?  Or is any compiler old enough to need this not going to support the C99 things?

Edit: Oh, remembering now that the coverage of this is so limited it surely isn't useful any more, so it seems like a moot point anyway...

pat-rankin:
No compilers ever needed it to build a functional program.  It dealt with alternate code used to prevent warnings from the pre-ANSI 'lint' tool.   Actually building the program with LINT defined would result in lots of serious problems like functions returning pointers to auto arrays instead of static ones hence losing control of whatever they contained.

PR #939, opened 2022-11-29
Add a funny message for smelling yourself

entrez:
I know this is inane, especially for a little-used debug-mode-only
command, but I can't get it out of my head...

The use of the actual monster data instead of the glyph as a proxy ought
to make it somewhat more consistent as well (e.g. hallucination and
'showrace' won't affect the results), though I think that'd hardly be
worth it were it not for the incredibly funny message.

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -1908,14 +1908,14 @@ DISABLE_WARNING_CONDEXPR_IS_CONSTANT
 static int
 wiz_smell(void)
 {
+    struct monst *mtmp; /* monster being smelled */
+    struct permonst *mptr;
     int ans = 0;
-    int mndx;  /* monster index */
-    coord cc;  /* screen pos of unknown glyph */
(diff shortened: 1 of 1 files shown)


pat-rankin:
Catchall of "You see no monster here." doesn't seem right.

entrez:
What would you suggest instead?  I changed it to use "see" because it has the `canspotmon` requirement, though that could be removed -- the current function prints "That is not a monster" but that can be printed when a monster is there but is hiding, mimicking something, invisible, etc (since it relies on the glyph).  So I thought either the message should change or the requirements of monster visibility should be removed, and ended up changing the message, but it was somewhat arbitrary.

entrez:
@pat-rankin -- I added this second commit separately so that it will be easier to drop it from the branch later, since I don't know if it's really going to be what you want, but I can squash the two commits together if it ends up being suitable.

PR #944, opened 2022-12-06
Limit graves to produce only one corpse/undead mon

entrez:
Graves could produce multiple undead and/or corpses.  For example, as a
sort of worst-case-scenario, for a grave in a bones level: the grave was
generated with the former hero's corpse on top of it.  Then a new hero
engraves on the headstone with a wand of digging and wakes an undead
creature from the grave.  Then the hero digs up the grave with a pick
and finds another corpse or undead monster inside.  That's three corpses
or undead creatures associated with a single grave.

Use the existing 'disturbed' field (actually alias for 'horizontal') to
track whether the grave ought to be empty (i.e. whether it has already
produced a corpse somehow), and check its value before creating more
corpses or undead from a particular grave.

### DIFF
--- a/include/rm.h
+++ b/include/rm.h
@@ -331,7 +331,7 @@ struct rm {
 /* horizonal applies to walls, doors (including sdoor); also to iron bars
    even though they don't have separate symbols for horizontal and vertical */
 #define blessedftn horizontal /* a fountain that grants attribs */
-#define disturbed  horizontal /* a grave that has been disturbed */
+#define disturbed  horizontal /* a grave without a corpse in it */
 
 struct damage {
     struct damage *next;
--- a/src/dig.c
+++ b/src/dig.c
@@ -926,6 +926,12 @@ dig_up_grave(coord *cc)
         You("have violated the sanctity of this grave!");
     }
 
+    if (levl[dig_x][dig_y].disturbed) {
+        /* grave already produced a corpse: either corpse was generated on the
+           surface to begin with (bones), or hero disturbed the grave */
+        goto empty_grave;
+    }
+
     switch (rn2(5)) {
     case 0:
     case 1:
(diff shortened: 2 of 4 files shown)


copperwater:
One implication of this change is that the player is now able to engrave on a grave to guarantee a ghoul, which despite its paralysis attack is a fairly weak undead. This will avert the threat of a heavier-hitting monster like a giant mummy or zombie coming out of the grave when it's dug up, while still allowing the player to get at whatever items are buried in the grave.

Note that you do not need a wand of digging: any bladed weapon works for summoning a ghoul, and it does not get dulled in the process because the undead disturbing happens before you can actually engrave.

Possibly a better solution is to have some algorithm that deterministically chooses based on the XYZ coordinates of the grave what sort of thing is in it, with the possible options being ghoul, zombie, mummy, corpse x2, and nothing. Then, engraving on the grave could unleash the monster if it's any of the first three, and subsequent engraving attempts will just fail with some message like "You can't engrave on a headstone".

entrez:
That's a good point.  Could maybe also just include all the monster types in a random roll when engraving (this is what I thought was happening, I think, I didn't realize it was always a ghoul).  It wouldn't be as perfectly consistent and thought-out, but it would avoid the problem you mention.

copperwater:
Actually, I thought of something else too: one reason it is currently dangerous to dig up a grave is not only because you may have unleashed a zombie or mummy you're not prepared to fight, but also because you're now down in a pit and can't easily move away. As implemented now this allows you to get the zombie or mummy out of the way without that drawback.

Two possible solutions:
1. deterministically compute what's in the grave, ghoul-containing graves react to engraving but none of the other types do, graves containing non-ghouls continue to work as normal
2. simply disallow all engraving on graves with a "You can't do that" generic message and remove the interaction where engraving generates an undead at all. If there's a ghoul, it will pop out when the grave is dug up.

entrez:
That seems like a minor enough benefit (and an smart enough trick) that I don't know if it's so bad to allow it, to reward players for thinking outside the box in their graverobbing...  Especially since it's not like it's a bug, but a natural consequence of being able to disturb the undead by scratching on their tombs.

PR #945, opened 2022-12-07
Fix: billing shop boulder pushed thru shared wall

entrez:
I realized that the "move the boulder billing to 'used-up items'" part
of 20392a6 didn't properly handle boulder movement from one shop to
another via a gap in a shared wall, becuase it wasn't looking at the
complete in_rooms() array (plus, it only triggered if the new spot
wasn't in any shops at all).  When I tried to fix that, I realized that
stolen_value() was similarly not working reliably when passed a location
shared between two shops, and for the same reason: it was only using the
first character of the in_rooms() array.

I think this patch fixes both those things, but it would be worth
examining the change to stolen_value() carefully, to ensure getting the
roomno via the shkp won't change anything about its normal functioning.
I'm not totally sure about that -- I didn't notice any problems in some
brief tests of typical stolen_value() uses, but it seems like it
probably has some tricky edge cases.

At the very least, passing a boulder fully through a shared wall between
two shops one way, then back the other way, no longer triggers an unpaid
obj sanity check in my testing.

### DIFF
--- a/src/hack.c
+++ b/src/hack.c
@@ -451,8 +451,10 @@ moverock(void)
                            != 0)
                        && onshopbill(otmp, shkp, TRUE)) {
                 subfrombill(otmp, shkp);
-            } else if (otmp->unpaid && !*in_rooms(rx, ry, SHOPBASE)
-                       && *in_rooms(sx, sy, SHOPBASE)) {
+            } else if (otmp->unpaid
+                       && (shkp = find_objowner(otmp, sx, sy)) != 0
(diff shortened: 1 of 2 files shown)


entrez:
It would be nice to get the roomno directly instead of going through the 
shopkeeper, both in stolen_value() and in moverock(), but a 
"find_objowner(), but it returns a char instead of a monst \*" function 
seemed redundant.  If I am overlooking a function like that, let me 
know...

PR #956, opened 2023-01-16
Replace Win32API sound player with FMOD

MrEveryDay98:
Replaces the very limited Win32API playSound function with the FMOD library


MrEveryDay98:
Need help linking the FMOD libraries, currently new sounds stop playing after a few of them play because the resources are not freed when they're done playing.

PR #957, opened 2023-01-16
Replace Win32API playSound function with FMOD

### LINKED ISSUE #816: NetHackW better sound support
User should be able to set at least 2 flags in the sound options:

1. Allow the specific sound to be played synchronously with other sounds (so that they are not cut off).
2. Play a sound on loop (effectively supporting music playback).

MrEveryDay98:
Replaces the Win32API playSound function with FMOD, allowing for more robust audio playback (example, background music).

Documentation recommendation:
"Music files should have their filename contain 'music_', this way the game will play the file on loop until another 'music_' file is loaded


MrEveryDay98:
Footage: [https://www.youtube.com/watch?v=7PkidwCy8BM](url)

paxed:
Files from external libraries should not be committed to NetHack; those should be under the lib-dir. Neither should there be DLL files.

MrEveryDay98:
> Files from external libraries should not be committed to NetHack; those should be under the lib-dir. Neither should there be DLL files.

I'm not really a programmer (I used OpenAI to help build this) nor do I have Github experience so I'll ask:
What do mean by they should be under the lib-dir and why shouldn't there be DLLs?

nhmall:
> I'm not really a programmer (I used OpenAI to help build this) nor do I have Github experience so I'll ask: What do mean by they should be under the lib-dir and why shouldn't there be DLLs

The GitHub NetHack repository contains the source code for NetHack, NetHack build instructions and scripts, and some NetHack documentation in text form or in source form for producing other types of document files using various tools.

The process of building NetHack generates a bunch of additional new files that are not part of the repository. Those generated files are not source files, they are artifacts that result from the build process. The game executable, generated libraries, binary data files, generated images, dynamic link libraries, etc. all fall into that "generated" category. None of those should ever be committed into the source code repository. Nobody can scrutinize them or verify what they contain.  They would unnecessarily inflate the repository with opaque content specific to one platform (windows in this case), and, besides, they can all be recreated from the source and build process anyway.

If you want to utilize 3rd party libraries with your cloned copy of NetHack, or utilize 3rd party code when you build NetHack, the correct approach is **not** to place a point-in-time snapshot or copy of that 3rd party code/content into the NetHack repository, and distribute that 3rd party code/content as though it were an integral part of NetHack itself.  It might not even be legal to do that in many cases, depending on the license of the 3rd party software. Even if it were legal, it's certainly a bad idea because the 3rd party code is likely maintained somewhere else, by somebody else.  Any copies of it would become stale almost immediately, or (even worse) the copies would have to be updated perpetually from the source and become an ongoing maintenance nightmare.

The lib subdirectory doesn't have any content in the NetHack repository. It's either empty or non-existent when you first clone the NetHack repository.

When you place content into the _lib_ subdirectory of your clone of the NetHack repository, the intention is for it not to be considered something that needs to be committed into the NetHack repository. The contents of _lib_ can still be utilized by the NetHack build process if the build processes expect it.  The content in _lib_ and its subdirectories will be ignored by git itself because there's a line in a file called .gitignore that instructs git to ignore everything in lib.
```
lib/*
```
That line in .gitignore was placed in there deliberately. 

Your pull request #957 (as it stands at the time of this writing) actually includes the removal of that line from .gitignore, thus breaking that necessary usage across all NetHack platforms.

The _lib_ subfolder is meant to be a place that you can manually place 3rd party stuff on your own build machine for use when building NetHack. You **can commit** instructions into the NetHack repository on where the 3rd party stuff can be found on the internet, and where it should be placed in a specific subdirectory of _lib_, and you **can commit** careful modifications to build procedures, Makefiles, or project files to include 3rd party content from a _lib_ subfolder, and **you can** commit text or documentation on how to make use of 3rd party stuff that's in a _lib_ subdirectory. That might be useful to other NetHack builders who might also want to utilize the 3rd party content or library. And of course **you can** commit new _glue code_ to NetHack source files to make use of the 3rd party content, but it should be done in a way that it isn't a requirement or necessity, but rather a build option.

MrEveryDay98:
@nhmall thank you, that was very informative. I'll make some more commits when I have time.

MrEveryDay98:
Current plan:

FMOD must be compiled with NetHack using instructions in sys/wind/Install.windows

both ntsound.c and fmsound.c will exist

code plan:
In applicable files, check nethackrc:
	• OPTION=sound:win32api // this will be the default if the option is not present in nethackrc
		>> Use the vanilla sound function play_usersound(), ignore fmplay_usersound()
	• OPTION=sound:fmod
		>> Use FMOD sound engine and use fmplay_usersound() instead of play_usersound(), init_fmod() to start the sound system

nhmall:
Some changes were just committed to NetHack-3.7 to update the interface for using sound libraries. That's been coming for a while, and wasn't meant to create a moving target for you.

To use the new sound interface layout, you can use the following sample as a skeleton starter:
https://gist.github.com/nhmall/93b1036daa9ba4cafe3fd30b57328a41

You could replace the text "sample" in there with "fmod",  and save it in sound/fmod/fmod.c.
Then add your interface code for invoking your library routines to provide the functionality expected.

Compile with -DSND_LIB_FMOD so that SND_LIB_FMOD is defined during the build.

You don't have to worry about the existing basic win32api functionality, it can sit there side-by-side now.
**ntsound.c** is no longer present in the repository. It has been replaced by **sound/windsound/windsound.c**
which conforms to this new interface.

MrEveryDay98:
Would it be better to close this PR? I'm pretty sure I messed up my repo and accidentally committed things I didn't want to commit. I only meant to send the fmod.c file.

nhmall:
> Would it be better to close this PR? I'm pretty sure I messed up my repo and accidentally committed things I didn't want to commit. I only meant to send the fmod.c file.

There's no harm in closing it, although the commentary that led to this point will not be associated to a new PR. I would say do whatever is easiest for you.

I'll point out some observations that might be useful for your potential next PR:

General

- Your original seemed to be making changes to various files unrelated to your stated goal of using FMOD. I would avoid that, if I were you. It makes your PR less-likely to be accepted, at least quickly.
- Your pull request #957 included creation of src/Makefile.mingw32 and src/Makefile.mingw32.depend, IIRC. The _src_ folder, in addition to containing the NetHack sources shared across all platforms, is where platform-specific Makefiles etc., typically get copied ahead of the build process. That coping of Makefiles is done manually for some platforms or compilers (as it was in your case with mingw32), or by a script file in the case of Unix builds (sys/unix/setup.sh), or by a .bat file for Windows Visual Studio builds (sys/windows/nhsetup.bat). Those build-related Makefiles in _src_, which are there only for the build, should not get committed to the repository. If you need new functionality in the build-related Makefiles to support the changes in your pull request, your pull request should alter the originals, sys/windows/Makefile.mingw32 and sys/windows/Makefile.mingw32.depend in your case.  If the Makefile changes are really invasive, new Makefiles could be made (please avoid that if possible), but they would never go into the _src_ directory, they would go into the subdirectory of _sys_.
- One other general note: if the vendor of the library has header files that your build needs to reference, leave those header files where they are in the vendor library files, and add a -I compiler switch to add the vendor's directory to the include file search path in your Makefile.mingw32 or visual studio vcxproj file changes. (for example -I../lib/vendordir/include).  Don't make copies of vendor-provided header files into one of the NetHack directories (other than lib as previously discussed). I don't know if your most recent changes were doing that, but I think a past set may have been.

Specific to the new sound interface file you included (you seem to be on the right track):

- A minor point: Your code is hinting to the core that all of these are active, because they are all set:
    SNDCAP_USERSOUNDS | SNDCAP_HEROMUSIC
        | SNDCAP_ACHIEVEMENTS |SNDCAP_SOUNDEFFECTS,
 but in reality there seems to be code for just the fmod_play_usersound, so I would remove the others until they actually have working code too, leaving only SNDCAP_USERSOUNDS. That way the core will not even bother invoking the others. That's the intent, at least.

PR #966, opened 2023-01-25
Fix: use-after-free in dog_eat()

entrez:
I think moving the m_consume_obj call (which will free the eaten item)
further down should fix this without causing any really wacky message
sequencing issues, but if maintaining the exact order is important
obj->unpaid and its price could be cached before the free instead.


entrez:
Removed the unrelated music.c bit per b89d2e0ef10d3a732c8a6a74346c67e02d62440a

entrez:
7f4e69f

PR #970, opened 2023-02-04
Fix: default lregion exclusion area occupied real space on the map

copperwater:
The intuitive behavior of des.levregion or des.teleport_region when "exclude" is left unspecified is that there is no exclusion area. However, this wasn't actually the case: since l_get_lregion defaulted the exclusion area to (0,0,0,0) and exclude_islev to 0, this meant that the 0,0 space on the map would always be excluded from regions. In cases where a region was specified with its inclusion area constrained to the 0,0 space of the map, this would create a "Couldn't place lregion" impossible message.

This fixes that issue by defaulting the exclusion area to (-1,-1,-1,-1), and if the exclusion area is left unspecified, forces exclude_islev=1. This means that the exclusion zone will be outside the walkable space of the level where it can't cause any problems.

If a level designer puts negative coordinates in their inclusion or exclusion parameters, this might not work correctly, but negative region coordinates aren't currently used anywhere and probably shouldn't be supported anyway.

### DIFF
--- a/doc/lua.adoc
+++ b/doc/lua.adoc
@@ -896,7 +896,7 @@ Example:
 Example:
 
  des.teleport_region({ region = { x1,y1, x2,y2} });
- des.teleport_region({ region = { x1,y1, x2,y2}, region_islev = 1, exclude = { x1,y1, x2,y2}, exclude_islen = 1, dir = "up" });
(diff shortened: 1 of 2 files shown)


paxed:
Merged

PR #974, opened 2023-02-06
Fix #969: dying in shared shop wall

### LINKED ISSUE #969: place_object: off map error with shops
Scenario: player dies while standing in the space where a hole was dug into the wall of a shop. As their objects are placed in that space, the 'program in disorder' occurs. Version of 3.7 - editlevel 71. Report from the player:

> <desertgreg> Hi K2, I got a bunch of messages while playing nethack on hardfought.org. "place_object: "elven boots" [0] off map <0,0>  Program in disorder!  <desertgreg> What happened is I died in minetown while standing in a hole in the wall of a shop. think the errors are occurring while placing all of my items on the ground.

entrez:
inherits() only examined the first item in u.ushops, so wouldn't work
correctly in some cases where the hero died in a shared shop wall and
neither shopkeeper had been robbed.  This seems to work in some basic
tests of behavior in a shared wall, but I haven't done extensive testing
on all the different combinations of robbed shopkeeper elsewhere in
level, shopkeeper angry but not robbed, etc.

Fixes #969

### DIFF
--- a/src/shk.c
+++ b/src/shk.c
@@ -1924,8 +1924,8 @@ inherits(
     long loss = 0L;
     long umoney;
     struct eshk *eshkp = ESHK(shkp);
-    boolean take = FALSE, taken = FALSE;
-    int roomno = *u.ushops;
+    boolean take = FALSE, taken = FALSE,
+            uinshop = (strchr(u.ushops, eshkp->shoproom) != (char *) 0);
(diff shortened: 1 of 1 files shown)


entrez:
I rebased and squashed these commits to the parts that still seem potentially useful to me even after 1d06fa6.

PR #976, opened 2023-02-09
Fix: 'A'+'P' regression when dropping by type

entrez:
Selecting both 'P' (the just-picked-up items category) and 'A' (the
'auto-select relevant items' flag) when dropping items by type with 'D'
was automatically dropping every item in inventory instead of only items
that were marked as just having been picked up.  Basically, it was
causing 'A' to act like it did before b65c93c amended its functionality.

This commit is more-or-less identical to 991e739, both in terms of the
problem and the fix, except that it applies to dropping items instead of
looting.


entrez:
I had this based on the wrong branch by mistake, so it originally included the commit from #975 and the description/title were messed up.  Should make more sense now.

PR #979, opened 2023-02-10
x_monnam: use 'the' with adjective + given name

entrez:
Most x_monnam articles retained the given article for an adjective +
name phrase, and y_monnam/ARTICLE_YOUR would drop the article even if
there were adjectives before the name.  Some of the resulting messages
and phrases like "Invisible Fido appears" (for ARTICLE_YOUR) or
"interior of an invisible Worm Boy" (for ARTICLE_AN) sound wrong to me;
normally if I am using an adjective with a given name, I will include
use a definite article ("the invisible Sally").  If there's a given name
combined with adjectives, use "the" (and continue to drop the article if
there's a given name on its own).

I have done some testing with this and the results sound more
idiomatic/natural to me, but hard to say for sure there wouldn't be any
strange results given the myriad different circumstances in which
monster names are used.

### DIFF
--- a/src/do_name.c
+++ b/src/do_name.c
@@ -2022,8 +2022,8 @@ x_monnam(
         name_at_start = (boolean) type_is_pname(mdat);
     }
 
-    if (name_at_start && (article == ARTICLE_YOUR || !has_adjectives)) {
-        if (mdat == &mons[PM_WIZARD_OF_YENDOR])
+    if (name_at_start) {
+        if (has_adjectives || mdat == &mons[PM_WIZARD_OF_YENDOR])
             article = ARTICLE_THE;
         else
             article = ARTICLE_NONE;


entrez:
I'm not sure how contentious this might be, whether the existing rules sound better to some people.  I found some mentions of the "definite article with adjective + proper name" thing [here](https://english.stackexchange.com/questions/200045/the-adjective-proper-name) and [here](https://ell.stackexchange.com/questions/214813/using-the-definite-article-the-with-adjective-before-a-persons-name), but the explanations and discussion don't seem very solid.  There's probably a better source for information on this rule(?) out there somewhere.

nhmall:
Workshopping that a little bit further, suppose the message was describing a family member, perhaps your brother named John.

```
interior of the John
interior of a John
interior of John
```
The latter seems grammatically optimal, to me.

```
The invisible John appears.
Invisible John appears
```
It is the latter that sounds more correct, to me.

entrez:
> Workshopping that a little bit further, suppose the message was describing a family member, perhaps your brother named John.
> 
> ```
> interior of the John
> interior of a John
> interior of John
> ```
> 
> The latter seems grammatically optimal to me.

Definitely agree when there aren't any adjectives involved.

> 
> ```
> The invisible John appears.
> Invisible John appears
> ```
> 
> It is the latter that sounds more correct, to me.

I don't think it's necessarily wrong to say "invisible John appears" (though "an invisible John" definitely doesn't seem right) but I think "the invisible John appears" is still how I would say it.  Maybe it's regional, or just a personal preference thing, but to me "invisible John appears" sounds like his invisibility is being especially emphasized as the defining "thing" about John.  I agree with the person who wrote a response at [this link](https://english.stackexchange.com/questions/516680/usage-of-a-definite-article-with-a-proper-name-defining-characteristic) I guess.  [And here's another discussion I found about it.](https://www.englishforums.com/English/TheBeforeAnAdjectiveName/bclclh/post.htm)

I guess one reason I think it's better to lean towards inclusion of the definite article in NetHack is that it's much more common to use mon_nam, which already uses this rule, so your pet is much more frequently referred to as "the invisible Fido".  So especially when sequences mix mon_nam and y_monnam, it feels a bit weird to me, like it's switching styles for no reason: "The invisible Bucephalus turns into an invisible saddled gray dragon!  You adjust yourself in the saddle on invisible Bucephalus."  (This could be made more consistent by dropping all articles if we're dealing with a name, no matter whether there are adjectives or not -- that's just not how I would say it myself.)

Ultimately it's not that big a deal and probably doesn't merit this long response. ;)

nhmall:
If the choices are all grammatically correct, then perhaps this does become a personal preferences discussion.  
You know what that means for NetHack, right?
- someone's preference wins, and those in agreement with the choice are happy.

or

- someone's preference loses, and those in agreement with the choice are dissatisfied, and will bring it up as an addendum to any future discussion about any other topic whatsoever, so as to get a chance to continue to express their dissatisfaction with the choice that was made on the matter.

or

- it becomes a config file option. :-)

entrez:
Well, I think something does need to change about ARTICLE_AN, which is what's producing "interior of an invisible Joe" -- not necessarily ungrammatical (say, if there are multiple "invisible Joes"), but I don't think it's correct in the context of the game.  So there would need to be a choice made to bring that in line with either mon_nam ("the invisible Joe") or y_monnam ("invisible Joe"), and my thought when writing the commit was that you may as well normalize treatment of the adjective + proper name construction across the board while you're in there.

If that change ought to be limited to ARTICLE_AN and leave a mixture of different "writing styles" in the game, that's fine.   But I also doubt anyone really cares about this enough to bear a grudge about it; it's not something I would imagine is going to rile up a big mob of grammaticians either way.\*  :)

\* Gb pryyf bs gur Crbcyr'f Qrsvavgr Negvpyr Eribyhgvbanel Cnegl: gbb zhpu urng evtug abj, jnvg sbe gur fvtany gb fgevxr jura guvatf unir pbbyrq bss.

PR #980, opened 2023-02-10
Invert the behavior of selection.gradient

copperwater:
selection.gradient has some pretty unintuitive behavior, in that it
selects points that are NOT close to the defined center. I've used
gradient selections several times and so far all of them have had to be
negated, because I wanted to select points close to the center with a
decreasing probability further out.

This implements that behavior, and also fixes a bug in which the x,y
coordinates of the gradient center(s) were not converted properly when
used within a des.room or des.map. Also updated the lua documentation
for gradient.

I removed the "limited" argument, as it was previously used to control
whether the rest of the map outside the max given distance would be
included in the selection; now that the area beyond maxdist is naturally
never in the selection, it doesn't have much use. (And I can't think of
a reasonable use case for the inverse: wanting to select points close to
the center, with decreasing chance towards maxdist, but then select the
entire map beyond maxdist.)

Currently this does not affect any special levels or themed rooms
because none of them use selection.gradient.

This pull request also contains a commit to drop the `isqrt()` math for gradients, which removes some bias and chunkiness from them.

### DIFF
--- a/doc/lua.adoc
+++ b/doc/lua.adoc
@@ -1090,11 +1090,23 @@ Example:
 
 === gradient
 
-Create a "gradient" of selected positions.
+Create a "gradient" of selected positions, radiating outward from a center point
+or line.
+x and y are required; x2 and y2 are not required. If they are provided and are
+different from x and y, the center of the gradient will be a line; otherwise it
+will be a point source at (x,y).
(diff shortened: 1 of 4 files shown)


paxed:
Merged.

PR #984, opened 2023-02-13
Don't anger target via mon vs mon zap/breath

entrez:
dobuzz() was modified in 677b32c2a7 to anger the target, but beams
handled by dobuzz() don't necessarily originate from the hero.  Monsters
can zap wands that hit other monsters, dragons can use their breath
attacks, etc.  Those things were causing the targets to become angry at
the hero.


paxed:
Merged.

PR #988, opened 2023-02-24
fix: add build patch for osx build

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/124039


chenrui333:
thanks @nhcopier @nhmall!

PR #991, opened 2023-02-28
Enable statue traps with specific statues of monsters

copperwater:
Previously, attempting to specify a statue trap would have it unconditionally create a statue of a random monster in maketrap() on top of the trap, which is no good if one wants a certain type of monster to be the statue trap. (A poor workaround would be to add another statue of the desired monster on the same space since statue traps animate the topmost statue, but since this would create a pile of statues, it would be readily apparent.)

This commit lets the level designer create a statue trap by specifying the "trapped" field on a statue created with des.object. Instead of setting otrapped on the statue, it will create a statue trap on the floor underneath it.

A des.trap('statue') will still always create a statue of a random monster. I originally implemented an extra flag to pass into des.trap that would disable this behavior, but then figured that a statueless statue trap still needs a des.object('statue') statement to make the statue, which could just be specified as trapped in the first place. So it wouldn't really add anything.

Internally, this moves the statue creation out of maketrap(), so that a trapped statue can call maketrap(STATUE_TRAP) without side effects. This means that responsibility for making a statue now falls to the caller of maketrap, which is only relevant in 3 places (generic mktrap(), wizard mode trap wishing, and adding random traps to a maze).

### DIFF
--- a/doc/lua.adoc
+++ b/doc/lua.adoc
@@ -807,7 +807,7 @@ The table parameter accepts the following:
 | lit         | boolean  | Is the object lit?
 | eroded      | int      | Object erosion
 | locked      | boolean  | Is the object locked?
-| trapped     | boolean  | Is the object trapped?
+| trapped     | boolean  | Is the object trapped? (If the object is a statue, a statue trap will be created underneath it.)
 | recharged   | boolean  | Is the object recharged?
 | greased     | boolean  | Is the object greased?
 | broken      | boolean  | Is the object broken?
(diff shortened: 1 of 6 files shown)


paxed:
Couple problems: 1) Some special levels use des.trap("statue") - those should be fixed. 2) Generating a trapped statue object in random location may put it on top of existing trap.

copperwater:
> Some special levels use des.trap("statue") - those should be fixed.

The initial commit accounted for this - as I said above, "A des.trap('statue') will still always create a statue of a random monster." I've made sure the levels that use statues of random monsters in random locations, like Arc-loca, work as expected. Is there something else you meant?

I've addressed your second point by deleting a trap that a statue trap happens to generate on top of.

Issue #990, opened 2023-02-27
[3.7-73] Truncated message encountered in game window


chappg:
I encountered the following truncated message in the game window:

"The spell hits the dwarf!  You kill the dwarf!  Welcome to experience level"

The game log correctly shows "Welcome to experience level 6", so it appears the display string was truncated for some reason.

Game Settings:  130x30 in terminal with curses and 2 status lines in the game options
Version: 3.7.0-73

entrez:
Did you have a MSGTYPE=stop on the "welcome to experience level" message?  And perm_invent enabled?

chappg:
Good question- I do not have that specific stop msg configured nor on the nearby messages, but I do have others set on potentially deadly conditions:

https://www.hardfought.org/userdata/a/a1/nethack/a1.nhrc

And yes, perm_invent is enabled.

pat-rankin:
Level gain and level loss were recently changed to force --More-- (or >> for curses) so that sound effects for those two events have time to play, so they have an  implicit MSGTYPE:stop now.  (The new pause doesn't depend upon whether sound effects are actually enabled.)

I'm not sure if that was a good idea, but without it #levelchange in wizard mode played the loss effect just once even when dropping multiple levels.  (That behavior might depend on audio hardware.)  For tty, the gain effect was already getting played once per increment due to the more verbose messaging it has.  But that probably wouldn't be the case for a multi-line message window for other interfaces.  (It's not something that comes up in normal play because the change only goes up or down by 1 level at a time.)

entrez:
Seems like this would have fulfilled all the criteria to trigger the #993 issue then, can't say with absolute certainty but I feel pretty sure that must be it.

chappg:
I believe the pull request will fix the issue. I received a 'More' prompt which I immediately ack'd before writing down the bug. Thank you!!!

vultur-cadens:
> I'm not sure if that was a good idea, but without it #levelchange in wizard mode played the loss effect just once even when dropping multiple levels. (That behavior might depend on audio hardware.) For tty, the gain effect was already getting played once per increment due to the more verbose messaging it has. But that probably wouldn't be the case for a multi-line message window for other interfaces. (It's not something that comes up in normal play because the change only goes up or down by 1 level at a time.)

This is kind of annoying in wizard mode because you can't skip all of the level change messages with a single press of ESC anymore (using tty).  It's only wizard mode though.

[...]

PR #993, opened 2023-03-01
Prevent curses 'More' (>>) from overwriting msg

entrez:
In some scenarios where a message prompted a 'more' (in curses shown as
'>>'), the '>>' could overwrite the last one or two characters of the
message.

This could happen if a single message on its own was the length of the
message space minus 2 (e.g. for an 80 character terminal, "h - a concave
amulet named this is an amulet name, it is very long, and so on."), or
if an addition to an existing line brought it to the length of the
message space minus 1 or 2 (e.g. "You materialize on a different level!
You remember this level as testlevlname.").  The two scenarios had
slightly different causes.

I think this fixes both of those scenarios.  I believe this is the
cause of the problem described in GitHub issue #990 but there isn't
enough detail to know that for a fact.

### DIFF
--- a/win/curses/cursmesg.c
+++ b/win/curses/cursmesg.c
@@ -105,9 +105,11 @@ curses_message_win_puts(const char *message, boolean recursed)
         mesg_add_line(message);
     }
 
-    /* -2: room for trailing ">>" (if More>> is needed) or leading "  "
-       (if combining this message with preceding one) */
-    linespace = (width - 1) - 2 - (mx - border_space);
+    /* -3: room for trailing ">> " (in case More>> is needed) */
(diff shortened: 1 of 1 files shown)


entrez:
Forgot to mention, this truncation/overwriting happens only if the message window is "full up" and displaying an extra line would require scrolling all the existing messages.   `curses_block` has this behavior as a deliberate fallback for the bottom line (maybe could be modified to scroll the window if `!noscroll` but even adding that in, it seemed to truncate the 78-character string when I did a quick test).
https://github.com/NetHack/NetHack/blob/76e82d1312d348d2f04ee65be390f15270948af6/win/curses/cursmesg.c#L203-L208

PR #999, opened 2023-03-18
Remove the manual 50% vertical swap of doors in Baalzebub level

copperwater:
This was added before level flipping, and its purpose effectively was to randomize the interior secret door layout (and thus the entire level, because the rest is vertically symmetric) by manually flipping it. Now that the entire level can be flipped and produce this same effect, this is no longer necessary.


paxed:
Committed.

PR #1000, opened 2023-03-18
Make "cookie-only" rumors explicit in rumors file

copperwater:
The prior behavior was an ugly kludge that treated rumors as only-eligible-to-appear-from-fortune-cookies if they contained the string "fortune" or "pity". This commit gets rid of that system and introduces a system where each rumor can be specified as such by prefixing it with "[cookie]".

In order to avoid changing existing behavior, I have applied the [cookie] prefix to all the rumors which were affected under the old rules. However, several rumors seem like they were misclassified under the old rules, and I am in favor of changing them as follows:

Should be made cookie-only (they only really make sense in context of eating a fortune cookie):
- "Ouch. I hate when that happens."
- "PLEASE ignore previous rumor."
- "Suddenly, the dungeon will collapse..."
- "Unfortunately, this message was left intentionally blank."

Should be made non-cookie-only (they make sense if they appear as engravings or artifact whispers):
- "They say that a gypsy could tell your fortune for a price."
- "They say that fortune cookies are food for thought."

### DIFF
--- a/dat/rumors.fal
+++ b/dat/rumors.fal
@@ -22,7 +22,7 @@ A visit to the Zoo is very educational; you meet interesting animals.
 A wand of deaf is a more dangerous weapon than a wand of sheep.
 A wand of vibration might bring the whole cave crashing about your ears.
 A winner never quits.  A quitter never wins.
-A wish?  Okay, make me a fortune cookie!
+[cookie] A wish?  Okay, make me a fortune cookie!
 Afraid of mimics?  Try to wear a ring of true seeing.
 All monsters are created evil, but some are more evil than others.
 Always attack a floating eye from behind!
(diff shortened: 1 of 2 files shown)


pat-rankin:
Those last two might be suitable for engravings or artifacts, but they aren't relevant to oracle messages.  Minor oracles always come from the set of true rumors and those two are both in the false set.

copperwater:
OK, I edited the summary, but it doesn't really change the point: those two probably don't deserve to only come from fortune cookies.

One other thing I forgot to point out is that this change opens up the possibility of true rumors that can exclusively be found in fortune cookies, using the same system. Here's one example from xNetHack:

    [cookie] Fortune cookies are a nice snack, but you'll need some real food at some point.

paxed:
Committed.

PR #1005, opened 2023-03-29
livelog shopkeeper deaths

vultur-cadens:
The livelog message includes the shopkeeper's shop type, since the livelog for stealing also mentions the shop type.

This also adds a revival counter for each shopkeeper.

### DIFF
--- a/src/mon.c
+++ b/src/mon.c
@@ -32,6 +32,8 @@ static struct permonst *accept_newcham_form(struct monst *, int);
 static void kill_eggs(struct obj *);
 static void pacify_guard(struct monst *);
 
+extern const struct shclass shtypes[]; /* defined in shknam.c */
+
 #define LEVEL_SPECIFIC_NOCORPSE(mdat) \
     (Is_rogue_level(&u.uz)            \
      || !gl.level.flags.deathdrops    \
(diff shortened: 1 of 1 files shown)


copperwater:
I don't think it's really important to count how many times a shopkeeper is revived so that it can be livelogged again the 3rd, 5th, ... times. Of course, the livelog shouldn't be spammed with every revival in case someone decides to try stoning and unstoning a shopkeeper a ton of times. But if the player kills and revives a shopkeeper once, it's not very interesting if they do it again.

Instead, why not add a condition such that it only livelogs the shopkeeper's death if `mtmp->mrevived` is false? Then you retain livelogging the death of every separate shopkeeper and avoid spam, but don't have to deal with storing and tracking how many times each shopkeeper has been revived, which ought to only matter in extreme edge cases or extinctionist runs anyway.

vultur-cadens:
I thought, since the other unique monsters are treated the same way (unique demons, Medusa, Vlad -- even though I guess only the Wizard and Riders are actually interesting when killed multiple times), I'd treat shopkeeper logging the same way, for consistency.  The livelog eventually only logs every 50 kills, so I don't think it would be any worse than logging somebody repeatedly killing Medusa.

Pets also have a edog->revivals counter (that doesn't seem to be used for anything), so I thought of adding one for shopkeepers.  I didn't think of using mrevived, but I could re-implement it that way and get rid of the revival counter.

vultur-cadens:
Removed shopkeeper revival counter.  Only the first kill per shopkeeper is logged now.

PR #1010, opened 2023-04-07
Split bestowing artifact into separete function

argrath:


paxed:
Already committed, forgot to close.

PR #1012, opened 2023-04-16
Fix: breathless monsters always generate in water in special levels

copperwater:
... unless explicitly specified to generate at a specific point or within a specific area. But if they are permitted to generate anywhere on the level, and it contains water, they always end up in the water. I noticed this when trying to explicitly specify ghouls to generate anywhere on a level with a minimal amount of water.

This was due to the definition of "amphibious" being conflated with "breathless", such that all breathless monsters counted as amphibious. There are plenty of breathless monsters in the game that decidedly don't normally inhabit water, such as undead, but they would pass the amphibious() check in pm_to_humidity and thus the game decides that they must generate in wet terrain if there is any available.

This fix takes the approach of changing amphibious() so that it no longer checks the M1_BREATHLESS flag and only considers M1_AMPHIBIOUS, then updating the places where amphibious() and Amphibious are used accordingly. I also added a new macro cant_drown() which wraps up swimming, amphibiousness, and breathlessness because these three things are frequently checked together in the context of whether something should drown.

Places where amphibious() or Amphibious did NOT have an extra breathless() or Breathless check added on, and thus where behavior has been changed:
- The pm_to_humidity function (to fix the bug).
- Player vs water in goodpos; it didn't seem like being polymorphed into a breathless non-amphibious monster should make it fair game to randomly teleport into water even though it's technically safe.
- Awarding extra experience when killing an eel. (So the hero will get the extra experience if they are polymorphed into a breathless non-amphibious monster and don't have magical breathing. Very much an edge case.)

### DIFF
--- a/include/mondata.h
+++ b/include/mondata.h
@@ -42,8 +42,8 @@
                        && (!is_clinger(ptr) || !has_ceiling(&u.uz)))
 #define is_swimmer(ptr) (((ptr)->mflags1 & M1_SWIM) != 0L)
 #define breathless(ptr) (((ptr)->mflags1 & M1_BREATHLESS) != 0L)
-#define amphibious(ptr) \
-    (((ptr)->mflags1 & (M1_AMPHIBIOUS | M1_BREATHLESS)) != 0L)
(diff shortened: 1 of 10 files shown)


paxed:
Committed.

PR #1030, opened 2023-05-19
move VDECL definition to vmsconf.h

argrath:
As VDECL macro is used only in sys/vms, move its definition to vmsconf.h.


pat-rankin:
The VMS port is in danger to moving to outdated.  If it survives, pre-ANSI code in it should be replaced.  That would include switching from use of varargs.h to stdarg.h.  But testing changes to it is a problem.

I'm not sure whether the USE_VARARGS/USE_OLDARGS/VDECL stuff in tradstdc.h should be removed altogether.  The code currently assumes--or requires--C99 and that makes the old variable argument stuff obsolete.

argrath:
The fate of the VMS port, VDECL etc. should be discussed at another issue/PR.

PR #1039, opened 2023-05-30
Make loadstone useful

vultur-cadens:
#906 had loadstones give steadfastness (preventing knockback), but loadstones are still too heavy to be practically useful.

This pull request contains 3 commits, so parts of it may be picked if you don't like all of them:
1. Reduce the weight of blessed loadstones to 100 units, 1/5 of their normal weight.  They are magically heavy items, so it makes sense for a blessing to change how the magic acts.
2. Allow putting a loadstone in a bag without it autocursing, if it does not leave your inventory.  This allows players to put a loadstone away to protect from the curse items spell when they don't need steadfastness.  The loadstone autocurse condition is changed to be the same as crysknife reversion condition, so dropping a bag containing a loadstone will degrade the loadstone's beatitude.
3. Prevent loadstones (and bags containing them) from being put in bags of holding, so that they will still be a decent load to the hero when not being used for steadfastness.  Loadstones will not explode bags of holding since we don't need more bag-exploding things; they will just refuse to be put in.  This can be explained by loadstones being smart enough to not want to go in a bag of holding (because they would rather be a load to the hero), just like they are smart enough to not want to leave the main inventory when they are cursed.  As a side effect, attempting to put a sack containing a loadstone and a wand of cancellation into a bag of holding will not explode the bag of holding, because the loadstone will prevent it before the explosion.

Loadstones are still an early-game trap for players without a means of uncursing them.  Players who accidentally pick up a loadstone and then dump it as soon as they uncurse it should not notice any difference.  But this pull request does give players a reason to carry a loadstone if they are able to bless it.  A blessed loadstone is also heavier than Giantslayer, has the risk of causing encumbrance if it gets cursed, and can't be put in a bag of holding to reduce weight, so it doesn't make Giantslayer useless either.

### DIFF
--- a/include/extern.h
+++ b/include/extern.h
@@ -1516,6 +1516,7 @@ extern int add_to_minv(struct monst *, struct obj *);
 extern struct obj *add_to_container(struct obj *, struct obj *);
 extern void add_to_migration(struct obj *);
 extern void add_to_buried(struct obj *);
+extern void container_weight(struct obj *);
 extern void dealloc_obj(struct obj *);
 extern void obj_ice_effects(coordxy, coordxy, boolean);
 extern long peek_at_iced_corpse_age(struct obj *);
--- a/src/do.c
+++ b/src/do.c
@@ -858,6 +858,17 @@ obj_no_longer_held(struct obj *obj)
             obj->oerodeproof = 0;
         }
         break;
+    case LOADSTONE:
+        if (obj->blessed) {
+            unbless(obj);
+        } else {
+            curse(obj);
+        }
+        /* loadstone may change weight */
+        if (obj->where == OBJ_CONTAINED) {
+            container_weight(obj->ocontainer);
+        }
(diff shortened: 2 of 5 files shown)


copperwater:
Any efforts to make loadstones "better" or more attractive to use, while still keeping them loadstones, are putting lipstick on the proverbial pig. No one is going to go around using up 100 aum of their carrying capacity just to get steadfastness - 100 aum which may suddenly inflate to an undroppable 500 if you accidentally set your bag down for a moment while forgetting it's in there.

Ultimately they'll remain a pointless gotcha for new players and a spoiler tax for those who want to learn how to informally identify and use gray stones properly.

Loadstones should be removed entirely, rather than adding features to them which enshrine them as an integral part of the game. I don't object to the steadfastness property or having sources of it besides Giantslayer, but loadstones aren't the answer.

PR #1043, opened 2023-05-31
silvergrey

duwling:
Monsters, objects, and menu colors
+ curses: swaps gray and foreground color
+ tty: hardcodes gray to the same as foreground color, and ignores terminal setting

It turned out that the fix actually adds gray to the colors.


duwling:
Here are some screenshots from my workstation. For the show I've set gray to a bright color.

curses swaps gray/fg
tty ignores gray

After the patch.

ok after patch

PR #1051, opened 2023-06-07
Provide escape item in water-surrounded vaults

entrez:
Water vaults are one of the few places that can/will generate completely
sealed off in a normal level.  Other such spots are designed to provide
a guaranteed means of escape (vault guard, scrolls of teleportation in
niches, etc) -- water vaults were an exception that didn't do this, so a
hero who fell into one from above could have ended up in a position
where she had no choice but to wait to starve to death or #quit. Provide
an escape item in one of the vault's chests to give a hero more options
in that position.

Also fix a minor mistake (I'm pretty sure, though I'm not a Lua expert
enough to be certain) in an nhlib.c comment describing how to use
obj.addcontent() -- when called as box.addcontent(contents) as the
comment suggested it produces an error, but works OK when called as
box:addcontent(contents) or obj.addcontent(box, contents).

### DIFF
--- a/dat/themerms.lua
+++ b/dat/themerms.lua
@@ -628,10 +628,25 @@ xx|.....|xx
 }----}
 }}}}}}]], contents = function(m) des.region({ region={3,3,3,3}, type="themed", irregular=true, filled=0, joined=false });
      local nasty_undead = { "giant zombie", "ettin zombie", "vampire lord" };
-     des.object("chest", 2, 2);
-     des.object("chest", 3, 2);
-     des.object("chest", 2, 3);
-     des.object("chest", 3, 3);
+     local chest_spots = { { 2, 2 }, { 3, 2 }, { 2, 3 }, { 3, 3 } };
+
+     shuffle(chest_spots)
+     -- Guarantee an escape item inside one of the chests, to prevent the hero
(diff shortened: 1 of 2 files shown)


entrez:
xNetHack solves this a slightly different way (the escape item is placed
directly on the floor of the vault, not inside one of the chests, via a
helper function way_out_method(themerms.lua)) but I think it's sort of
neat that the item is inside one of the vaults since it will likely
require a bit of ingenuity and/or experimentation to figure out which 
object it is.

The xNetHack solution is nice though since it can be applied to more
sealed theme rooms (which do exist in xNetHack), while this patch is
much more specific to the water vault with the chests.  Adapting the 
xNetHack helper function-> object on the ground (or trapdoor/hole) 
approach is another possible option.

pat-rankin:
I've just put this in, but if the relevant chest is locked and the randomly chosen item is a wand which happens to be made of glass, that wand could be destroyed if the only way to access the chest's contents is by kicking it until its lock breaks.

entrez:
That's true, I didn't think about that as a factor in comparing this to the xnethack solution (of dropping an item directly onto the floor).  If it's a problem I guess the list of escape items could be restricted further, or the chest with the escape item could be generated unlocked.

PR #1054, opened 2023-06-09
color interface

duwling:
In my opinion, these would make interface more friendly.


FredrIQ:
Personally I disagree. Reverse merges bright+dark colors resulting in far more overall ambiguity. In general I feel like vanilla makes poor use of background coloration to give more information, but it's something I never felt strongly enough about to push for changes.

And besides, isn't there a check for fg==bg? I seem to recall making such a thing, is it possible that it never made its way into mainline curses? The "black is blue" transformation should only happen for darkgray, not true black. Or am I mistaken? Meaning that it would be translated to blue, then my code detects fg==bg and turns foreground black.

duwling:
Currently tty interfaces reverse colors for hilite_pile. But the frame coloring (window ports?) starts using dark only, though bright may be completly different&#8202;&mdash;&#8202;i.&thinsp;e. glass is bright-cyan and not associated with iron.

Not sure if that check is necessary. Perhaps background is lightish&#8202;&mdash;&#8202;don't care, nocolor is both (16 + 1).

Blue for black is a relict from old times and should only be a fallback for systems with 16 colors or less. Even 8 color systems have a good change to display bright colors (so darkgray), but can't differ black from either background or foreground.
See https://github.com/duwling/NetHack/commit/b34e2823b4c585740748f838e267200766ff4f19

PR #1058, opened 2023-06-10
Allow wish for "high altar" in wizard mode

entrez:
This would make it more convenient to test of things that treat or
affect normal altars and high altars differently, e.g. altar coloration,
something that can destroy a normal altar but not a high altar, etc: the
developer can wish for a "high altar" or "lawful high altar" and so on
to interact with one without levelporting to the Sanctum or Astral and
dealing with the monsters there.

The syntax is pretty rigid ("high lawful altar" doesn't work) -- I do
have another commit locally that is much more flexible in that regard
but since this is strictly a debug command and that adds more complexity
to the code I figured it's probably better to keep it simple.

### DIFF
--- a/src/objnam.c
+++ b/src/objnam.c
@@ -3329,18 +3329,22 @@ wizterrainwish(struct _readobjnam_data *d)
         aligntyp al;
 
         lev->typ = ALTAR;
-        if (!strncmpi(bp, "chaotic ", 8))
-            al = A_CHAOTIC;
-        else if (!strncmpi(bp, "neutral ", 8))
-            al = A_NEUTRAL;
-        else if (!strncmpi(bp, "lawful ", 7))
-            al = A_LAWFUL;
-        else if (!strncmpi(bp, "unaligned ", 10))
-            al = A_NONE;
-        else /* -1 - A_CHAOTIC, 0 - A_NEUTRAL, 1 - A_LAWFUL */
+        if (!strncmpi(bp, "chaotic ", 8)) {
+            al = A_CHAOTIC, bp += 8;
+        } else if (!strncmpi(bp, "neutral ", 8)) {
+            al = A_NEUTRAL, bp += 8;
+        } else if (!strncmpi(bp, "lawful ", 7)) {
+            al = A_LAWFUL, bp += 7;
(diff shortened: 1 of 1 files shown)


copperwater:
It's not particularly difficult to have a blessed scroll of genocide in your wizkit.txt and genocide * to get rid of all the troublesome monsters on the Sanctum or Astral, though.

entrez:
Yeah, I know you don't have to fight the monsters in Astral or the Sanctum one by one when testing something in wizmode, but it'd still be more convenient to be able to easily wish one up anywhere, and I don't really see any downside to allowing it.

pat-rankin:
High altars only occur inside temples.

It used to be possible to ascend in wizard mode by wishing for an altar as soon as you arrived on the Astral Plane level.  Requiring the altar to be inside a temple deliberately put a stop to that.  With this proposed change, you wouldn't even need to visit that level; just step off the stairs and wish for a high altar and the Amulet and ascend on turn 2.  It isn't strictly necessary to prevent that from happening but allowing it still seems pretty cheesy.

entrez:
I don't think it's a big deal that it could be abused to gain a fake ascension if "playing" in debug mode.  It's already possible to ascend at T:2 (or T:4 at the latest) by `#wizlevelport`ing to Astral, using `#wizmap` to reveal the map, `#wizintrinsics` to give yourself monster detection, then if necessary `#wizmakemap` and `#wizmap` until all the altars are clear of monsters (all free actions), then teleporting to the altars until you hit the right one.  It's a debug mode, so I don't see the reason to limit debugging powers for the sake of making fake ascensions a bit less trivial for someone misusing wizard mode.

But if you think the ability to more easily test actions with a special/different effect on high altars isn't useful enough to be worth it, well, it is definitely a niche use case.  It's probably not hard to guess that this patch was inspired by some mild annoyances while testing #1057, but I admit I don't really know how useful it would be otherwise -- just seemed to me like it wouldn't hurt anything to have it as a debug/development capability.

entrez:
Closing this since Pat isn't a fan, and it's not so useful that I feel an urge to keep lobbying for it (though I still think it wouldn't really do any harm).

PR #1064, opened 2023-06-21
fix MENUCOLOR non-attribute overriding

Xdminsy:
One may make their wearing and wielding items underlined.
`MENUCOLOR="(being worn|wielded)"=gray&underline`

But the latter matching rules only setting a color will change
the line's attribute none. It forces players to manually copy
their rules and add underlined versions.
`MENUCOLOR="(robe)"=bright magenta`
`MENUCOLOR="(robe).*worn"=bright magenta&underline`

So if the matching rule's attr is ATR_NONE, we keep searching
back previous rules to find the first matching rule which provides
an attribute.

This may slightly influence the performance due to more regex_match.
But makes the config more concise if one has many MENUCOLORs.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -7728,13 +7728,19 @@ boolean
 get_menu_coloring(const char *str, int *color, int *attr)
 {
     struct menucoloring *tmpmc;
+    boolean first_match = TRUE;
 
     if (iflags.use_menu_color)
         for (tmpmc = gm.menu_colorings; tmpmc; tmpmc = tmpmc->next)
             if (regex_match(str, tmpmc->match)) {
-                *color = tmpmc->color;
+                if (first_match) {
+                    *color = tmpmc->color;
+                    first_match = FALSE;
+                }
                 *attr = tmpmc->attr;
-                return TRUE;
+                if (*attr != ATR_NONE) {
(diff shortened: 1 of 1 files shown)


pat-rankin:
I think the menu-color data will need to have flags that distinguish between _no color specified_ and _explicitly specified NO_COLOR_;similarly for attributes.  A second rule should only be found and used if the first one leaves either of the two fields unspecified and the second rule fills in the gap.  (Probably when the second rule _only_ fills in the gap rather than specify both color and attribute.)

Also, this now falls off the end of the loop and returns False if a pattern specifying color&no-attr matches but no other pattern which also matches is found when looking for an attribute one.  (I haven't looked to see whether the called really cares about the return value.)

Xdminsy:
Thanks for your reply!
I think I should return `!first_match`, my code did wrong :(

Yes I agree the MENUCOLOR rules may need more design, I find it cannot specify only attributes.

I heard that there's a MENUCOLOR patch but I don't know what it does, for me cuz I wrote maybe hundreds of menucolors in my .nethackrc, the main problem is the rules don't fall back to check attributes, so I need to add one more underline version to every color rule. Somewhat painful lol.

PR #1067, opened 2023-06-29
Sit to trigger squeaky board even if flying

Ardub23:
If the hero deliberately sits on the floor while flying over a squeaky board, then either they're trying to squeak it on purpose or they haven't noticed it. Either way, sitting should trigger it.


pat-rankin:
This commit should be redone with a log message that is manually wrapped so that lines fit within 80 columns wide.  Since git indents log messages by 4 spaces, the lines themselves should fit within 76 columns.  Even a bit less than that so there will be some right margin when viewed in a basic 80-column display.

I think that instead of
    if (Levitation || Flying) && !forcetrap && !viasitting) { ...
it would be clearer to use
    if (Flying && viasitting)
        forcetrap = TRUE;
    if (Levitation || Flying) && !forcetrap) { ...
because we can't tell from looking just at this that it won't be reached when levitating (controlled via the sit code).

Ardub23:
Rather than initializing forcetrap and then immediately updating it based on viasitting, I've cut out the viasitting variable and folded the VIASITTING flag check into forcetrap's initialization. I initially had the variable just because that's how the flag is used in trapeffect_web and trapeffect_poly_trap, but here I realized I don't expect a need for behavior distinct from forectrap, so I scrapped it.

(Also, hopefully I've done the log message properly now? I'm not a Git expert yet.)

pat-rankin:
That routine is used for monsters triggering traps as well as for the hero, but testing 'Flying' is only appropriate when the hero is the reason the trap is triggered.  (That routine won't ever be called with the VIASITTING flag for a monster, but it's still a good idea to avoid testing 'Flying' if the circumstance isn't to check for a monster's activity while hero is flying.)

The new log message is fine but 'patch-1' still has the previous commit with the inconveniently formatted one.
[You can see with 'git log -2' (log messages only) or 'git show -2' (log messages plus corresponding diffs).]
Merging it into NetHack-3.7 would bring both commits along.
[You could use 'git reset HEAD^^' (or 'git reset HEAD^' twice) to go back to before those two commits, then 'git rebase NetHack-3.7' to bring in changes that have gone into the 3.7 branch, then reapply your revised change as a single new commit as if from scratch.  That can get tricky if steps don't get performed in the optimal order though, and it might also need 'git push -f' to make sure that the pull request gets updated.]

If reset goes awry, recovering can be quite a hassle.  If recovery turns out to be needed, it will be less daunting to delete this pull request, make a new branch containing the final update as one commit, then open a new pull request for that new branch.  (Multiple commits aren't an actual problem but are most useful when maintaining access to intermediate stages is worthwhile.  I don't think that's the case here, particularly when it would propagate the first misformatted log message.)

PR #1079, opened 2023-07-12
Fix: wrong glyphs for clouds and water on Planes

entrez:
ff727e9 introduced an issue where unseen water and cloud on the
respective planes were shown with open and closed drawbridge glyphs
instead of the appropriate glyphs.  This is because they fall into cmap
section B, but the translation from symbols/cmap index to glyphs was
being done as though they were in cmap section A (ff727e9 manually used
that particular cmap section macro instead of the general cmap_to_glyph
to work around some compiler warning).

### DIFF
--- a/src/mkmaze.c
+++ b/src/mkmaze.c
@@ -1465,9 +1465,9 @@ static void mv_bubble(struct bubble *, coordxy, coordxy, boolean);
 void
 movebubbles(void)
 {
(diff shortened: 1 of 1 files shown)


entrez:
Screen Shot 2023-07-12 at 3 26 23 PM
Screen Shot 2023-07-12 at 3 26 09 PM

nhmall:
Your fix is in the pipeline.

PR #1086, opened 2023-08-15
Release obufs used by sortloot

entrez:
Some further application of e43ec0c logic, which was intended to fix odd
messages produced by obufs clobbered by inventory updates (like "the
ogre lord yanks Cleaver from your corpses!").  That issue was still
lurking around because sortloot(), via sortloot_cmp(), was continuing to
call for obufs via loot_xname() without releasing them immediately.  It
was going through the entire inventory doing that, much like
display_pickinv() was prior to Pat's fix in e43ec0ce, so could cause
the contents of obufs to still be clobbered by perm_invent updates.
This changes sortloot_cmp() to releases the obufs it calls for as soon
as possible so that won't happen any more.

### DIFF
--- a/src/invent.c
+++ b/src/invent.c
@@ -320,7 +320,7 @@ sortloot_cmp(const genericptr vptr1, const genericptr vptr2)
                          *sli2 = (struct sortloot_item *) vptr2;
     struct obj *obj1 = sli1->obj,
                *obj2 = sli2->obj;
-    char *nam1, *nam2;
+    char *nam1, *nam2, *tmpstr;
     int val1, val2, namcmp;
 
     /* order by object class unless we're doing by-invlet without sortpack */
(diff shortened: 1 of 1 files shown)


entrez:
I should say, I haven't actually heard anything about this in 3.7 since
e43ec0c, but multiple people have reported it in EvilHack which cherry-picked
e43ec0c, and this is apparently the cause of the issue there.  So I believe
vanilla needs this fix as well, unless there's some other intervening commit I
overlooked which makes it moot.

[Actually, I just went back and tried it, and I do see it churning through the
obufs, so I think this does apply to vanilla too.  But note that it only
happens if sortloot is set a certain way; I think it happens if sortloot is
'full' but not entirely sure if there are other conditions, it took a bit of
experimenting with the options before I actually saw it.]

pat-rankin:
I've only glanced at this so far.

> I think it happens if sortloot is 'full'

You probably also 11 or 12 items which have the same class, sub-class (for armor and weapons, one or two others), bless/curse state, and enchantment (when applicable).

[Edit:  Actually you'll need 11 or 12 items overall which happen to pair up with other items that also need name to disambiguate, not necessarily all of them within the same ultimate grouping.]

[Edit again:  Actually-too, both 'loot' and 'full' should be subject to this so I've deleted the first sentence of the first paragraph.]

PR #1089, opened 2023-08-28
Adjust seenres when observed attack succeeds

entrez:
If a monster sees an elemental attack succeed from some other creature
or environmental danger, it will be willing to try those attacks again.


paxed:
merged

nhmall:
The commit was 8fa53d6ac57021e1a0875b1601a1c6e127be0419  .

PR #1091, opened 2023-08-28
Fix: rising from the grave as a zombie

entrez:
When the hero is killed by a zombie, she is supposed to arise from the
grave as a zombie of a type matching her race.  Commit 580c5a6 broke
this inadvertently -- done_in_by was passing gy.youmonst.data to
zombie_form() and relying on the old behavior where it attempted to
include the hero's race in is_elf, is_orc, etc if the permonst matched
the hero's role.  Because this no longer works, zombie_form was only
returning PM_HUMAN_ZOMBIE when passed the role permonst of an
unpolymorphed hero.  Use gu.urace.zombienum instead.  This would have to
be a little more sophisticated (falling back to zombie_form() if Upolyd)
if a polymorphed hero with unchanging could die via done_in_by but
that's not currently possible as far as I can tell.


entrez:
1dab134fce2a851df

PR #1093, opened 2023-09-18
Allow quest leader to catch and return unique item

entrez:
It's been possible for a while to throw your quest artifact to your
quest leader, completing the quest (he will then throw the object back
to you).  The quest leader will also react to being brought the Amulet
of Yendor (i.e. having it in inventory when you approach or #chat), but
tossing it over like the quest artifact was treated as a hostile act and
would cause the quest leader and guardians to become hostile and attack
the hero.

Make throwing the Amulet of Yendor at the quest leader work like
throwing the quest artifact, and treat the invocation items similarly
just in case people try throwing those (maybe because they are lost,
unsure what to do, and have already seen their quest leader catch and
react to the quest artifact -- regardless of the reason, throwing over
the Book of the Dead is almost certainly not intended to be an attack).
Doing so for invocation items will ID them but not do much more than
that.  The Amulet of Yendor triggers the existing quest pager message.

### DIFF
--- a/src/dothrow.c
+++ b/src/dothrow.c
@@ -1920,8 +1920,8 @@ tmiss(struct obj *obj, struct monst *mon, boolean maybe_wakeup)
     return;
 }
 
-#define quest_arti_hits_leader(obj, mon)      \
-    (obj->oartifact && is_quest_artifact(obj) \
+#define special_obj_hits_leader(obj, mon)      \
+    ((is_quest_artifact(obj) || objects[obj->otyp].oc_unique) \
      && mon->m_id == gq.quest_status.leader_m_id)
 
 /*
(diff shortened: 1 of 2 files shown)


entrez:
This is maybe a somewhat silly patch and it's doubtful it will be encountered
by many people, but I tried tossing my quest leader the AoY the other day to
see the "good job getting the Amulet" message and was surprised when everyone
got mad at me.  The specific implementation here could be reworked but I think
generally it shouldn't hurt anything to have handling for the player tossing
one of those special items at the quest leader (maybe in hopes of getting some
kind of clue about what to do next -- I guess the thought process for adding
the "hasamulet" pager message in the first place was probably similar).

PR #1099, opened 2023-09-23
Updating FMOD channels and FMOD System Updater

MrEveryDay98:
This fixes an issue where sounds would stop playing over each other after a few seconds and instead interrupt the currently playing sound. This was done by changing `static FMOD_CHANNEL channel1` to an array called `static FMOD_CHANNEL channelsA[256]` and having the sound player increment in the array by 1 every time a sound is played. In addition, `FMOD_System_Update(systemvar)` which enabled this fix.

### DIFF
--- a/doc/fixes3-7-0.txt
+++ b/doc/fixes3-7-0.txt
@@ -1251,6 +1251,7 @@ use #monster to make dragon steed breathe
 wand of probing reveals tin contents
 steam vortices and fog clouds leave steam clouds behind
 fog clouds maintain any clouds they are in, even poisonous ones
+bone devils summon skeletons
 
 
 Fixes to 3.7.0-x General Problems Exposed Via git Repository
--- a/src/minion.c
+++ b/src/minion.c
@@ -91,6 +91,9 @@ msummon(struct monst *mon)
                                                         : ndemon(atyp);
         cnt = ((dtype != NON_PM)
                && !rn2(4) && is_ndemon(&mons[dtype])) ? 2 : 1;
+    } else if (ptr == &mons[PM_BONE_DEVIL]) {
+        dtype = PM_SKELETON;
+        cnt = 1;
     } else if (is_ndemon(ptr)) {
         dtype = (!rn2(20)) ? dlord(atyp) : (!rn2(6)) ? ndemon(atyp)
                                                      : monsndx(ptr);


nhmall:
In d2f9f13 "Added fmod documentation" @MrEveryDay98 states:
```
3. In Visual Studio, right click the NetHackW solution, click properties, and do the following:
	I.	Go to C/C++>General>Additional Include Directories, add the path to the FMOD inc folder (*lib\fmod\api\core\inc)
	II.	Go to C/C++>Preprocessor>Preprocessor Definitions, change 'SND_LIB_WINDSOUND' to 'SND_LIB_FMOD'
	III.	Go to Linker>General>Additional Library Directories, add the path to the FMOD lib folder (*lib\fmod\api\core\lib\<target-CPU>)
	IV.	Go to Linker>Input>Additional Dependencies, add 'fmod_vc.lib' to the list
```

Recent NetHack.vcxproj and NetHackW.vcxproj file updates (44ecbb2a1bfc4a0bb81da82b10d09d4a1d88dfff) added conditional code  that should make some of those manual changes you are describing unnecessary. 

There are tests for the presence of all of the following:

1. *lib\fmod\api\core\inc\fmod.h 
2. *lib\fmod-test\api\core\lib\x86\fmod_vc.lib or *lib\fmod-test\api\core\lib\x64\fmod_vc.lib
3. *lib\fmod-test\api\core\lib\x86\fmod.dll or *lib\fmod-test\api\core\lib\x64\fmod.dll

If those tests pass, meaning that they all exist in the expected place in the NetHack lib subfolder, then the build should assume that their presence indicates that you intend to have the fmod support included during the build phase.
Note: You can rename the *lib\fmod folder to some other name to prevent the autodetection tests from passing, and thus not using the fmod integration during the build.

A passed detection test should mean that the following project changes happen without the manual edits that you are describing:

1. *lib\fmod\api\core\inc is added to the additional include directories of the project.
2.  Several C preprocessor macros are defined for the project, SND_LIB_FMOD USER_SOUNDS SND_SOUNDEFFECTS_AUTOMAP RCWAV
3. The appropriate *lib\fmod-test\api\core\lib\x86\fmod_vc.lib or *lib\fmod-test\api\core\lib\x64\fmod_vc.lib gets added to the AdditionalLibraryDirectories and dependencies.

Side note: The tests use the Visual Studio Platform designation for the Build phase (Win32 or x64, by way of the _$(PlatformShortName)_ macro) to determine whether to check the lib\fmod-test\api\core\lib\ **x86** subfolder or the **x64** subfolder for the components.

Further tinkering may happen, but that's where things currently stand.

PR #1100, opened 2023-09-25
Disambiguate b_trapped null bodypart value

entrez:
b_trapped was treating 0 as a null value for its bodypart parameter, but
0 is actually the value of ARM, so b_trapped(..., ARM) would be treated
as intending no A_CON abuse.  Add NO_PART = -1 to the bodypart_types
enum, and use that instead of 0 as the "no body part" value in
b_trapped, so that ARM can be passed to it without any ambiguity.

aosdict identified this issue in xNetHack and handled it differently; he
added NO_PART with a value of 0, incremented the existing bodypart_types
values, and padded the body part arrays so the actual body parts would
start at index 1.  I think using NO_PART = -1 is simpler, but that's an
alternative approach that can be used instead -- it is advantageous in
that it automatically fixes any other places where 0 is assumed to be a
non-body-part value that I may have overlooked.

### DIFF
--- a/include/hack.h
+++ b/include/hack.h
@@ -132,6 +132,7 @@ enum bhit_call_types {
 
 /* Macros for messages referring to hands, eyes, feet, etc... */
 enum bodypart_types {
+    NO_PART   = -1,
     ARM       =  0,
     EYE       =  1,
     FACE      =  2,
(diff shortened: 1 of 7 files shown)


paxed:
Merged

nhmall:
The commit was fa1f1134c8c06567522f6db3843e067921cbf79c.

PR #1101, opened 2023-09-27
Fix: clipping on horizontal window resize

entrez:
Clipping mode was not activated when a comfortably-sized game window was
resized horizontally to become too narrow to display the entire map,
causing various display errors and bugs if the window was resized like
that.  I think the horizontal resize check was removed by mistake in
d1dade164ef220a011bf89fa445a8b0012bf38df.


paxed:
Merged.

nhmall:
The commit was f64c2070df302575fd358f2510f418dbde3551af

PR #1107, opened 2023-10-03
Flipping coins.

NullCGT:
Applying one or more gold pieces now flips one of them, which will cause it to come up heads or tails. This is NetHack, so there are special cases for flipping a coin underwater or while fumbling or greasy.

I've tried to future-proof this commit so that the code will not need to be modified if other items are eventually added to the coin class.

If memory serves, there was a patch for this on the bilious patch database, but I was unable to locate it or who the original author was. If someone is able to find the original patch, please let me know. In any case, the code is entirely original.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -4336,14 +4336,16 @@ flip_coin(struct obj *obj)
         if (otmp->quan > 1L) {
             otmp = splitobj(otmp, 1L);
         }
-        if (carried(otmp))
-            dropx(otmp);
-        else
-            stackobj(otmp);
+        dropx(otmp);
         return ECMD_TIME;
     }
-    pline_The("%s comes up %s.", xname(obj), 
-                                rn2(2) ? "heads" : "tails");
+    if (Hallucination) {
+        pline(rn2(100) ? "Wow, a double header!"
+                       : "The coin miraculously lands on its edge!");
+    } else {
+        pline_The("%s comes up %s.", xname(obj), 
+                                    rn2(2) ? "heads" : "tails");
+    }
     return ECMD_TIME;
 }
 
--- a/src/apply.c
+++ b/src/apply.c
@@ -4333,14 +4333,14 @@ flip_coin(struct obj *obj)
     }
 
     if (lose_coin) {
-        if (otmp->quan > 1L) {
+        if (otmp->quan > 1L)
             otmp = splitobj(otmp, 1L);
-        }
         dropx(otmp);
         return ECMD_TIME;
     }
     if (Hallucination) {
         pline(rn2(100) ? "Wow, a double header!"
+                        /* edge case */
                        : "The coin miraculously lands on its edge!");
     } else {
         pline_The("%s comes up %s.", xname(obj), 
--- a/src/apply.c
+++ b/src/apply.c
@@ -3964,8 +3964,13 @@ apply_ok(struct obj *obj)
     /* all tools, all wands (breaking), all spellbooks (flipping through -
        including blank/novel/Book of the Dead) */
     if (obj->oclass == TOOL_CLASS || obj->oclass == WAND_CLASS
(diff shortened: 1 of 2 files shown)


entrez:
I like the xnethack method of flipping a coin (maybe based on the Bilious patch?  Not sure), which is to throw 1 coin into the air ('t1$<').

NullCGT:
I had no idea about xnethack's t1$<. That's really clever!

pat-rankin:
I thought that that old coin flip patch had already been incorporated.  Apparently not.

[Adding different coin types wouldn't enhance game play and is quite unlikely to happen.  There was some groundwork for it many years ago but no particular reason to follow that up.]

I don't think stackobj() will work without first place_object() but admit that I've never bothered to try that.  I don't see how you could flip a coin that isn't carried though.

There ought to be a hallucinatory variation for the result but I don't have any reasonable suggestions aside from maybe seeming to land on coin's the edge, neither heads nor tails.  Or perhaps use the appropriate body part from a randomly chosen type of monster?

NullCGT:
Makes sense! PR modified.

I'm not terribly satisfied with the hallucinatory messages, but I also had a hard time thinking of any funny coin-related puns. There's a 1% chance for the coin to land on its edge when hallucinating, because I figured that even if it's not real, it should be an *edge case.*

src/apply.c
```diff
@@ -4312,4 +4316,37 @@ flip_through_book(struct obj *obj)
...
+            otmp = splitobj(otmp, 1L);
+        dropx(otmp);
+        return ECMD_TIME;
+    }
+    if (Hallucination) {
+        pline(rn2(100) ? "Wow, a double header!"
+                        /* edge case */
+                       : "The coin miraculously lands on its edge!");
+    } else {
+        pline_The("%s comes up %s.", xname(obj),
```
copperwater: Most of the time the messages produced will be "You flip a gold piece. The gold piece comes up heads/tails." Changing it to "It comes up heads/tails" would flow better.

(Also, if you apply multiple gold pieces, wouldn't this produce "The 247 gold pieces comes up heads/tails"?)

copperwater (commented):
The xNetHack implementation is not based on the bilious patch (per its commit message) and is very simple: 

```diff
         if (u.dz < 0 && !Is_airlevel(&u.uz) && !Underwater
             && !Is_waterlevel(&u.uz)) {
+            if (obj->quan == 1L && !nohands(youmonst.data)) {
+                pline("You flip the coin...  It is %s.",
+                      rn2(2) ? "heads" : "tails");
+                pick_obj(obj);
+                return 1;
+            }
             pline_The("gold hits the %s, then falls back on top of your %s.",
                       ceiling(u.ux, u.uy), body_part(HEAD));
```

I am still in favor of having this trigger by throwing the gold piece rather than applying it. Having gold suggested as an option to apply seems weird. (Note that the nohands() check is important - it isn't needed in this apply implementation because there's a blanket check for that already.)

src/apply.c
NullCGT: Applying multiple gold pieces does not result in odd messaging. Tested. 

Agreed on the use of a pronoun.

NullCGT: As far as throwing a gold piece upwards goes, I think xnethack's solution is very elegant, as it usually is. In my mind though, flipping a coin is a distinct action involving manipulating a coin with one's fingers, not throwing a coin upward. That may be a meaningless distinction, however.

pat-rankin:
I think apply rather than throw is suitable.  Gold pieces could become an acceptable choice rather than a suggested choice so flipping a coin would be a minor Easter egg.  It's not as if anyone is going to have a game ruined because they weren't aware of it and didn't know how to have someone else resolve a 50:50 decision for them.  ;-)

I don't think it would matter much if context-sensitive inventory actions revealed the Easter egg by including "a - flip {it|one}" in the menu for inventory item '$'.

paulwinner:
Do we want the same behavior for (d)rop as well?

PR #1112, opened 2023-10-24
Enable hero to wash hands in fountains and pools (and sinks!)

entrez:
Dipping hands (with '-') or currently-worn gloves in a fountain or pool
will cause the hero to wash her hands, washing away any oil and clearing
the Glib intrinsic timeout.  This does mean bare hands can be used to
fish for fountain effects, without having to pick up a stray arrow and
carry it around as a dipping item, but having your hands be the only
thing that does not activate those effects felt weird.  If that would be
too unbalancing this could be scrapped.

With further commits this now also adds a system for #dipping items into
a sink, which can aid in potion identification (much like dropping a
ring into the sink can).  This is based on code from xNetHack written by
@copperwater, with additions and amendments.  It also means you can wash
your hands in a sink.

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -29,6 +29,7 @@ extern NEARDATA const struct c_color_names c_color_names;
 /* common_strings */
 extern const struct c_common_strings c_common_strings;
 #define nothing_happens c_common_strings.c_nothing_happens
+#define nothing_seems_to_happen c_common_strings.c_nothing_seems_to_happen
 #define thats_enough_tries c_common_strings.c_thats_enough_tries
 #define silly_thing_to c_common_strings.c_silly_thing_to
 #define shudder_for_moment c_common_strings.c_shudder_for_moment
--- a/include/extern.h
+++ b/include/extern.h
@@ -970,6 +970,7 @@ extern void dogushforth(int);
 extern void dryup(coordxy, coordxy, boolean);
 extern void drinkfountain(void);
 extern void dipfountain(struct obj *);
+extern int wash_hands(void);
 extern void breaksink(coordxy, coordxy);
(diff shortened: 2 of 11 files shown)


entrez:
This started with me thinking it would be cool to be able to dip your hands
into a sink to wash them, but that would involve making sinks a possible \#dip
target and having to figure out how to handle potions and other items, so I
changed tack to fountains and pools.  I do think it'd be nice to be able to
wash your hands in a sink though, since that's practically their whole purpose
in real life.

entrez:
I realized that xNetHack actually allowed dipping potions (and only potions) into a sink, so I brought that over in a slightly modified form (added the hand-washing, made sinks cause water damage to non-potion items, and some other minor changes).  If the ability to blank scrolls at a sink is too powerful, the water damage effect for non-potion items could be reconsidered.

entrez:
Incidentally, it was sort of a headache to add that new verbose-aware message, especially since each individual message (not just each type or category of message) has its own bit.  I had to make a choice here between rearranging everything in bitfield 3 (which would break the configuration of anyone who actually is using the verbose bitfield options), or making the dodip enum values nonconsecutive with dodip4 ~25 bits away from the other three (which would probably be confusing for anyone trying to configure the verbose bitfield options).  The Verbose bitfields introduce choices along those lines (or at least additional work/annoyance) when a developer wants to add a new verbose-aware message, and I suspect that's having a chilling effect where few new messages actually have !verbose variations.  I wanted to mention my experience here in case there is a discussion among the devteam about whether it should be reconsidered -- I would probably fall on the side of "yes", especially given the complexity of configuring it means the number of players using it must be low to none.

PR #1114, opened 2023-10-27
doc: add note that USE_XPM can be defined otherwise

bernhardreiter:
* Add an additional hint to win/X11/NetHack.ad that `USE_XPM` can be defined outsite of `config.h` as it is the case with `hints/linux-x11`-

### DIFF
--- a/win/X11/NetHack.ad
+++ b/win/X11/NetHack.ad
@@ -30,7 +30,8 @@ NetHack*text*borderWidth:  0
 ! tile_file names a file containing full-color tiles for the map.
 ! If you use a 100dpi (or greater) monitor you may wish to double the
 ! tile size so you can see the figures.  If NetHack was compiled to
-! use XPM (USE_XPM in config.h), the tile_file is a standard XPM file.
+! use XPM (USE_XPM in config.h or via the Makefile),
+! the tile_file is a standard XPM file.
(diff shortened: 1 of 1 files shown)


entrez:
Maybe it should mention that USE_XPM appears to be the default when
WANT_WIN_X11 is defined if using a hints file that supports it, so users don't think they need to edit the Makefiles themselves to achieve this.  Something like "or via the Makefile, as is the default with certain hints files when WANT_WIN_X11 is defined"?  That's pretty clunky, but there are also hints files that don't do anything with WIN_XPM.  I don't know, maybe it doesn't need to go into that much detail and most people can be trusted to realize the connection between hints and Makefiles and then grep the hints files for USE_XPM.

bernhardreiter:
For me the hint that there are other methods would be been good enough.

I had looked in the config.h files and did not find USE_XPM defined and thus the comment was not correct. I inspected the x11tiles file itself with `file` and saw that is was indeed an `xpm` file. 

The important part is to mention that there are other methods. It would be fine to get into more detail, the problem is that the default of WANT_WIN_X11 could change in the future and then this remark may get forgotten, so using a shorter more general text seemed more future proof to me.

entrez:
Yeah, I think you're right that this is probably not the place to go into intricate detail.

PR #1117, opened 2023-11-02
Fix: hurtling into wall of water

entrez:
A hero hurtled into a wall of water while levitating, flying, or wearing
water-walking boots wouldn't be stopped by it unless it was on the Plane
of Water.  Make it stop hurtling heroes immediately no matter the
location.

I also noticed that once I was hurtled safely into the wall of water, it
was described as a "pool" when I examined it with ':'.  Fix that, too,
even though I think it shouldn't really be encountered in-game.

### DIFF
--- a/src/dothrow.c
+++ b/src/dothrow.c
@@ -924,8 +924,7 @@ hurtle_step(genericptr_t arg, coordxy x, coordxy y)
     check_special_room(FALSE);
 
     if (is_pool(x, y) && !u.uinwater) {
-        if ((Is_waterlevel(&u.uz) && is_waterwall(x,y))
-            || !(Levitation || Flying || Wwalking)) {
+        if (is_waterwall(x, y) || !(Levitation || Flying || Wwalking)) {
             /* couldn't move while hurtling; allow movement now so that
                drown() will give a chance to crawl out of pool and survive */
             gm.multi = 0;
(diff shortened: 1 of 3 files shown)


entrez:
I noticed this because of a Reddit post that Pat replied to a couple
days ago.  I figured a fix would be coming down the pipeline soon after
that comment, but since it hasn't yet, I'm submitting this thinking it
may help save a couple minutes finding and fixing it.

pat-rankin:
I started looking at that but got diverted.  One thing I noticed was that if I was levitating adjacent to a wall of water and threw an item away from it, I saw the animation of it traversing the throw's path, but after I recoiled into the water and escaped, I got the message "Splash!" and the thrown item ended up in the water instead of where it should have landed.  I haven't attempted to figure this one out but it is repeatable.  (It was just a priest throwing his wielded mace, not an aklys or anything else special.)

entrez:
Oh, that's interesting.  I didn't notice that because I was hurtling myself with #wiztelekinesis.  Looks like water_damage_chain is overwriting bhitpos, so when control returns to throwit after hurtle it thinks the item landed in the water location.  I guess you could probably fix it by caching bhitpos before the hurtle call and restoring it afterwards, but I haven't tested it yet.

entrez:
I don't know if that's the right place to cache/restore it.  Maybe it should happen in water_damage_chain instead, since that seems to be the most direct culprit in this case.

entrez:
This is more directly connected with the overwriting on bhitpos and seems to work just as well as doing it in hurtle.

entrez:
This was merged as ccf997b and 5ba7939

PR #1123, opened 2023-11-06
Some mention_decor-related fixes

entrez:
- Revert part of "Fix: hurtling into wall of water"
- pooleffects messaging vs mention_decor
- Fix: mention_decor preposition vs unusual terrain


entrez:
There still is an inconsistency here in that the describe_decor message is a lot more detailed than the similar pooleffects message, but it didn't seem right to just increase the complexity of the pooleffects message until it matched describe_decor.  Maybe the describe_decor message can be extracted into its own function and used both places? (So they wouldn't be phrased slightly different any more, and they would match in handling of rare terrain, etc).

PR #1126, opened 2023-11-08
Limit bugle notes to the actual bugle scale

entrez:
Bugles have a much more limited range than other instruments due to
the way they are constructed and played.  Make bugles in NetHack reflect
this.

I think it is very likely that tooled horns would behave the same way,
but I let them be for now because it's not entirely clear what a "tooled
horn" is meant to be.  Giving both tonal scaring instruments limited
ability to open the Castle might be interesting, though, since it would
give players a reason to actually hang on to wooden flutes and harps.

This also has a additional effect of explicitly rejecting characters
entered by the player that aren't associated with the musical scale.
Previously it would pass those to Hero_playnotes along with the real
notes.

### DIFF
--- a/src/music.c
+++ b/src/music.c
@@ -36,7 +36,8 @@ static void charm_monsters(int);
 static void do_earthquake(int);
 static const char *generic_lvl_desc(void);
 static int do_improvisation(struct obj *);
-static char *improvised_notes(boolean *);
+static char *improvised_notes(struct obj *, boolean *);
+static boolean select_tune(struct obj *, char *);
 
 
 /*
@@ -576,7 +577,7 @@ do_improvisation(struct obj* instr)
 #undef PLAY_CONFUSED
 #undef PLAY_HALLU
 
-    improvisation = improvised_notes(&same_old_song);
+    improvisation = improvised_notes(instr, &same_old_song);
 
     switch (itmp.otyp) { /* note: itmp.otyp might differ from instr->otyp */
     case MAGIC_FLUTE: /* Make monster fall asleep */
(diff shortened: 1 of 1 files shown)


entrez:
This is another idea which is much more debatable than the straightforward
fixes I normally submit (similar to #1112 in that way -- I guess I am
receiving creative inspiration from some dubious Muse lately ;-).

g-branden-robinson:
I like the idea of restricting the tooled horn in the same manner as the bugle; your rationale seems fine to me.

However, according to the squeaky board logic, all 12 notes of the chromatic scale are recognized in the Mazes of Menace.  There is consequently no reason for whoever constructs the tonal instruments of NetHack to limit themselves to bugles and horns in C.

Here are some ideas:

* If you improvise with one of these wind instruments, you discover what key it's in.  "You blow some triads in C."

(If hallucinating: "You won't be taking giant steps on this instrument.")

* There only need be enough different horn tunings to cover the chromatic scale once the perfect fifth and major third overtones are accounted for.  There is an infinite number of enharmonic spellings of any given note.  Double-sharps and double-flats need not be dealt with.  So...

1. Horn in A♭: A♭ C E♭
1. Horn in E♭: E♭ G B♭
1. Horn in B♭: B♭ D♭ F
1. Horn in F: F A C
1. Horn in C: C E G
1. Horn in G: G B D
1. Horn in D: D F♯ A
1. Horn in A: A C♯ E
1. Horn in E: E G♯ B

So that's only nine kinds of horn to cover all 12 notes.  Does the castle passtune code already handle enharmonicity?

This does presume that the equal-temperament tuning system is rigidly applied in the NetHack universe.  If it doesn't, the problems get worse instead of better...

paulwinner:
I have no issues with the discussion regarding instruments and notes, they
are out of my wheelhouse.

However, I would like to suggest that the email be submitted to the OED as
an example of usage for the word "enharmonicity". I thought Branden, that
you must have made it up. But I looked it up, and it is a real word in both
the musical and chemical realms. Kudos.

entrez:
Haha, @g-branden-robinson, I didn't realize what a can of music theory worms this would invite opening...

Maybe the devteam can also license a short snippet of a Peter Brötzmann tune for the sound library integration on the scaring horn improvisation.

g-branden-robinson:
I checked `do_play_instrument` and only notes in the C major scale are accepted.  This permits some simplification.

You would then only need four horn tunings: F, C, G, and D.  That seems more practical.  And since that D horn will blow an F#, there's room for a jokey Easter egg if the castle passtune requires an F natural.

```
You_hear("gears grinding in protest.");
```

entrez:
I'm not going to work on resolving the merge conflicts in this pull request for now, since there's apparently little to no interest and anyway it was always more of a "that could be interesting, but it would need a lot of testing and discussion to see if it makes sense" idea than something I was convinced would improve the game.  I'm happy to fix it up if someone comes along and wants to try it out/playtest it, though -- just let me know.

PR #1132, opened 2023-11-11
remove unnecessary condition on optfn_pickup_types()

argrath:
`opts` here is always non-null, otherwise it leads segv at earlier code.


pat-rankin:
Many of the optfn_foo() routines have that same check.  Why single out pickup_types?

argrath:
optfn_pickup_types() uses `opts` as an argument of strlen() at the top of the function, so `opts` cannot be null.
Other optfn_foo() don't.

PR #1135, opened 2023-11-14
Tweak encumbrance messages

AndrioCelos:
This branch proposes a minor change to the messages related to encumbrance in `lift_object` and `pickup_object`:

- In the `pickup_burden` prompt, keep the message _'You have a little trouble lifting...'_ for becoming burdened, but change it to _'You have trouble lifting...'_ for becoming stressed. This better informs the player how badly encumbered you are when it's set to `unencumbered`, since those two encumberance levels are the only ones that shared a message. Furthermore, the speed reduction and other effects of being stressed are more than what I could call 'a little trouble'.
- Use the prefix _'You have trouble lifting/removing...'_ etc. for picking up each object at any encumbrance level. I'm not sure why this was used only when burdened but not worse, and would consider either showing this at any encumbrance level or never. I chose to show it because it informs the player which objects crossed encumbrance thresholds without using `pickup_burden`.

### DIFF
--- a/src/pickup.c
+++ b/src/pickup.c
@@ -62,7 +62,8 @@ static void tipcontainer(struct obj *);
 #define Icebox (gc.current_container->otyp == ICE_BOX)
 
 static const char
-        moderateloadmsg[] = "You have a little trouble lifting",
+        slightloadmsg[] = "You have a little trouble lifting",
+        moderateloadmsg[] = "You have trouble lifting",
         nearloadmsg[] = "You have much trouble lifting",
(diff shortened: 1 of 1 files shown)


entrez:
FYI, NetHack code style is `char *pfx` rather than `char* pfx`.

entrez:
I was going to put this in an edit but I guess you sniped me: The way you formatted the chains of ternaries is also not really typical (usually they're indented a bit more each time).  Personally I would probably change these to if/else-if/else since they're involved enough and since it's a standalone assignment you don't really have to do anything else to make that change work, but that's your call.

PR #1140, opened 2023-11-16
Add option to exclude dropped items from autopick, and to autopick stolen items

entrez:
This is based on a feature in UnNetHack (and I think some other variants
as well).  If the hero intentionally drops an item with 'pickup_dropped'
disabled, don't autopick it back up when walking over that square again.

Typically when the player drops an item, it's because she doesn't want
it in her inventory any more, and this option stops autopickup from
defeating that goal (especially useful for tasks like stash management
without a container).  Players have come up with workarounds to this
problem like toggling autopickup when approaching their stash pile or
adding name-based autopickup exceptions to allow them to exclude
individual items from autopickup, but this behavior should reduce the
need for those things.

I think 'pickup_dropped' is a little unfortunate because it suggests
equivalence to 'pickup_thrown' (i.e. any dropped items will be
automatically picked up regardless of autopickup exceptions).  Calling
it something like 'nopick_dropped' might be better, but as far as I can
tell options cannot start with the word 'no' because it's interpreted as
a negation of the rest of the option name.

### DIFF
--- a/dat/opthelp
+++ b/dat/opthelp
@@ -17,7 +17,7 @@ cmdassist      give help for errors on direction & other commands  [True]
 color          use different colors for objects on screen  [True for micros]
 confirm        ask before hitting tame or peaceful monsters        [True]
 dark_room      show floor not in sight in different color          [True]
-dropped_nopick exclude dropped objects from autopickup            [False]
+dropped_nopick exclude dropped objects from autopickup             [True]
 eight_bit_tty  send 8-bit characters straight to terminal         [False]
 extmenu        tty, curses: use menu for # (extended commands)    [False]
                X11: menu has all commands (T) or traditional subset (F)
--- a/dat/opthelp
+++ b/dat/opthelp
@@ -50,6 +50,7 @@ null           allow nulls to be sent to your terminal             [True]
                try turning this option off (forcing NetHack to use its own
                delay code) if moving objects seem to teleport across rooms
 perm_invent    keep inventory in a permanent window               [False]
+pickup_stolen  override pickup_types for stolen objects            [True]
 pickup_thrown  override pickup_types for thrown objects            [True]
 pushweapon     when wielding a new weapon, put your previously    [False]
                wielded weapon into the secondary weapon slot
--- a/dat/opthelp
+++ b/dat/opthelp
@@ -50,6 +50,7 @@ null           allow nulls to be sent to your terminal             [True]
                delay code) if moving objects seem to teleport across rooms
 perm_invent    keep inventory in a permanent window               [False]
 pickup_thrown  override pickup_types for thrown objects            [True]
+pickup_dropped evaluate pickup_types for dropped objects           [True]
 pushweapon     when wielding a new weapon, put your previously    [False]
                wielded weapon into the secondary weapon slot
 quick_farsight usually skip the chance to browse the map when     [False]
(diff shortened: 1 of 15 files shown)


pat-rankin:
The confusion of opposite behavior between similarly named 'pickup_thrown' and 'pickup_dropped' concerns me since hardly anybody reads the Guidebook.  'dropped_no_autopickup' is more accurate but ridiculously verbose.  Maybe 'pickup_dropped_not'?  Too silly or idiomatic?  Tossing in "not!" after the fact is annoying but intended to be, I think.

Does stack merging check pickup_dropped and reject or override if the two objects or stacks differ?  I didn't see anything like that in the commit's diff.  I don't recall whether merging pays any attention to pickup_thrown.

Object sanity checking should check for and complain if both pickup_thrown and pickup_dropped are set at the same time.

Just musing:  I've always thought that there ought to be a pickup_stolen that would kick in after dispatching nymphs and monkeys.  It has the virtue of matching the behavior of pickup_thrown, and players in the habit of filling up inventory slots with useless junk for pickpocket fodder could leave it set to False if they didn't want to pick that stuff back up.

entrez:
Yeah, I agree with you about the name (which is from UnNetHack), but I also had a hard time coming up with a concise name that was a negation of pickup_dropped but didn't start with "no" (like I think I mentioned in the commit message).  The best I could come up with that didn't seem extremely long was "dont_pick_dropped".  I'll keep brainstorming about it...

Maybe it could be changed to something that distances it from pickup_thrown without having to tie ourselves into knots negating it without the use of "no" -- like "may_pick_dropped" or "dropped_autopick", which are clearly related to item pickup but don't necessarily sound equivalent to pickup_thrown.

Another option is to just modify the options parser so it won't treat a "no" prefix as a negation if the rest of the option matches 'nopick_dropped', and just use that.

Your other points make sense too, I'll take a deeper look tomorrow.

entrez:
Here's one idea that sounds OK to me, "dropped_nopick".  I'm just throwing that up there now and will return to this tomorrow.

paxed:
Wouldn't it make sense to fold these into a single compound option with parameters what to pick/drop?

entrez:
I was thinking about that earlier today too, paxed.  Basically make it like paranoid_confirm, right?

But thinking about the issue where similar names between "pickup_thrown" and "pickup_dropped" suggested their behavior is equivalent, don't you think putting them all into one option would probably cause this kind of problem too?  That was the reason I held off on trying to change it earlier.  Though bhaak did say earlier (on IRC) that he doesn't remember anyone complaining about being confused by the pickup_dropped option, in all the years it's been in UnNetHack.

entrez:
I rebased this since 86c29f5 added some more was_thrown uses that needed
updating.  Still happy to explore changing the options into a compound option
if the consensus is that it makes more sense and the potential confusion over
'OPTIONS=pickup_special:thrown' vs 'OPTIONS=pickup_special:dropped' (or
whatever the option may be) wouldn't be too confusing.

entrez:
Rebased and force-pushed only to revert the change to EDITLEVEL per discussion
on the erinyes PR.

nhmall:
This was put in the pipeline. The increment to editlevel lags by approximately 10 minutes, so have a coffee or tea when the first commit comes in, if they don't arrive together.

nhmall:
Pull request commit: c08304f8
Bump of EDITLEVEL: c459630b

PR #1143, opened 2023-11-17
Ignore loot/multidrop 'A' if no filter specified

entrez:
The 'A' option in the #loot or 'D'rop menu selecting every selectable
item when used on its own has been the cause of many bag of holding
explosions and other typo-related frustration.  Now that it operates on
other filters rather than overriding them, actually require some other
filter be selected for it to have any effect.  This means that 'A' on
its own will do nothing, but 'A'+'a' will still act like 'A' alone
previously did.  I think this will reduce the rate of serious typo
accidents in games, without being intrusive and while still maintaining
'A' as a useful option (which I think it is, in it's 3.7 incarnation --
I use it all the time in combination with other filters, especially
justpicked).

### DIFF
--- a/src/do.c
+++ b/src/do.c
@@ -914,7 +914,7 @@ menu_drop(int retry)
     int n, i, n_dropped = 0;
     struct obj *otmp, *otmp2;
     menu_item *pick_list;
-    boolean all_categories = TRUE, drop_everything = FALSE, autopick = FALSE;
+    boolean all_categories = TRUE, autopick = FALSE;
     boolean drop_justpicked = FALSE;
     long justpicked_quan = 0;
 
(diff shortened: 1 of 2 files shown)


pat-rankin:
The current behavior is useful.  For players who don't like it, they can set _paranoid_confirmation:A_.  These examples use D but \#loot in or out should behave similarly once the container action has been established.

Without confirmation:
drop everything:  D A \<space or return\>

With confirmation:
drop everything: D A \<space or return\> y
drop by menu: D A \<space or return\> n
never mind: D A \<space or return\> q [or ESC]

In all these cases, explicit 'A a' in either order produces the same result as just 'A', without requiring both.  The 'n' confirmation case is the same as
D a \<space or return\>
or as _menustyle:partial_ plus just 'D', leading to a menu of all items.

After this pull request (based on its description, not actual testing), the intentionally drop everything case would be
D A a \<space or return\> . \<space(s) or return\>
_paranoid_confirmation:A_ seems superior to me.  The pull request doesn't remove it but assumes that players won't bother using it.

entrez:
Hi Pat, thanks for the feedback.  Yeah, I guess this could be considered as an alternative to the new paranoid_confirmation option, that applies to everyone and is more integrated into the existing loot/multidrop UI rather than invoking a separate prompt.  But I don't think it's necessarily either/or: I think they can both have utility for different uses of 'A' (or lack thereof).

My thinking was that I use the new 'A' behavior a lot in combination with other filters like justpicked, and I don't want to be slapped with a paranoid prompt constantly for doing that, but it's still a significant hazard to hit 'A' instead of 'a', or hit 'A' deliberately but not fully hit the other filter key so that instead of "justpicked items" it applies to everything in my inventory.

So this was my idea to address that issue, and require some demonstrated intentionality for "autopick everything in inventory" without getting in the way of deliberate use of 'A'.  I get the sense that many of the people who are going to use 'paranoid_Autoall' never use 'A' at all and want the prompt as a next-best alternative to removing it altogether.  But I also get the sense that may be most players, and my frequent use of it in combination with other filters is somewhat unusual even among people who are playing the 3.7-dev version, so maybe this is appealing to a crowd of basically just myself.

entrez:
It is true that this is somewhat incompatible with "paranoid_Autoall" in the sense that some behaviors there (in particular, prompting even if 'A' is the only option selected) should be changed if this patch were to coexist with paranoid_Autoall.

PR #1148, opened 2023-11-20
delete ancient commented out codes on wall_angle()

argrath:
These codes have been commented out since 3.4.3, at least.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -3426,22 +3426,6 @@ wall_angle(struct rm *lev)
             }
             break;
         case WM_T_BL:
-#if 0  /* older method, fixed */
-            if (only(seenv, SV4 | SV5)) {
-                col = T_tlcorn;
-            } else if ((seenv & (SV0 | SV1 | SV2))
-                       && only(seenv, SV0 | SV1 | SV2 | SV6 | SV7)) {
-                col = T_hwall;
-            } else if ((seenv & SV3)
-                       || ((seenv & (SV0 | SV1 | SV2))
-                           && (seenv & (SV4 | SV5)))) {
-                col = T_tdwall;
-            } else {
-                if (seenv != SV6)
-                    t_warn(lev);
-                col = T_stone;
-            }
-#endif /* 0 */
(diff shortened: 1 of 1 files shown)


nhmall:
I'm curious about this one - is your objection to it the fact that it is deactivated by `#if 0 ... #endif` block, instead of an actual comment?

If it was commented out with `/*...*/` would it still have caught your attention and still led to the submission of a pull request?

argrath:
I think `/*` ... `*/` is better, because it is less affected by global code changes like reformatting.
Also, if these codes are worth preserving in the source codes (maybe forever), meaningful comments should be added.

nhmall:
> Also, if these codes are worth preserving in the source codes (maybe forever), meaningful comments should be added.

A general comment, not specific to this instance:

Adding meaningful comments would most efficiently be done by the original author of the change. The original author understood the deficiency in the original code enough, at the time, to add the improved code.  There's a lot of deactivated and commented-out code in NetHack.  I doubt that anyone else is going to expend time on comparing the old (commented-out, or de-activated) code, and the new, corrected code, to reverse engineer the change enough to add meaningful comments to them all.

I suppose that probably means that the only realistic option left is to delete the  `#if 0...#endif` code. 

I, speaking for myself, would not look forward to a lengthy series of individual pull requests for every instance of commented-out or deactivated code, especially in unreleased work that is considered very much still under-development.  That sounds like something that could be done more efficiently by looking into them all, prior to commencing preparation for an upcoming release.  I agree that there isn't much value in leaving uncommented, deactivated code blocks in there for a new release.

I can say, with absolute certainty, that I don't want to discourage pull requests in general though. They are all welcomed, and this one made think about the motivation behind it.

EDIT:  In `I agree that there isn't much value in leaving them in there for a new release.`
Changed "**them**" to "**uncommented, deactivated code blocks**"

argrath:
Thank you for your comment.

It is no problem if these codes are to be deleted by some means at some time, even if it is not by this PR.

PR #1150, opened 2023-11-21
Rework erinyes into avenging Furies

entrez:
Based on conversations on IRC with paxed, bhaak, K2, and Umbire.  This is an idea based on the role of the erinyes in Greek myth: punishers of oathbreakers, hosts who harm their guests, murderers, and others who have violated important ancient social mores or customs.  This is reflected in NetHack by the erinyes becoming more dangerous as you abuse your alignment -- something like killing a peaceful monster does seem consistent to me with things that bring down the Furies in mythology.  The alignment abuse metric that determines the difficulty of the erinyes is tracked separately from alignment record, and only counts negative adjustments to alignment: 'adjalign(-5); adjalign(10); adjalign(-5);' would leave you with an alignment record of 0, but alignment abuse value of 10.

If extinction allows, one or more erinyes are also (edit: potentially, based on a roll against abuse value) summoned when you don a helm of opposite alignment (since you're breaking your oath to your god, so it especially incites them).  The helm of opposite alignment also causes brief confusion, and #offering is made impossible while confused, in order to make this a potential challenge on Astral rather than something which can easily be blown past by #offering the amulet on the next turn.

This could use more testing and conversation about possible changes or extensions (or alternatives), but I think it's at a point now where I am comfortable submitting the patch in its current state and starting some discussion around it.

### DIFF
--- a/dat/data.base
+++ b/dat/data.base
@@ -1730,8 +1730,18 @@ A.S*
 		[ Wikipedia, the free encyclopedia ]
 erinys
 erinyes
-	These female-seeming devils named after the Furies of mythology
-	attack hand to hand and poison their unwary victims as well.
+	Holy and pure, from Zeus Khthonios born
+	and Persephone, whom lovely locks adorn:
+	Whose piercing sight, with vision unconfin'd,
+	surveys the deeds of all the impious kind:
+	On Fate attendant, punishing the race
+	(with wrath severe) of deeds unjust and base.
+	Dark-colour'd queens, whose glittering eyes are bright
+	with dreadful, radiant, life-destroying, light:
+	Eternal rulers, terrible and strong,
+	to whom revenge, and tortures dire belong.
+		[ LXIX. To the Furies, The Orphic Hymns,
+		  translated by Thomas Taylor ]
 ettin
 	The two-headed giant, or ettin, is a vicious and unpredictable
 	hunter that stalks by night and eats any meat it can catch.
--- a/dat/rumors.tru
+++ b/dat/rumors.tru
@@ -332,6 +332,7 @@ They say that you should pray for divine inspiration.
 They say that you should religiously give your gold away.
 They say that you will never get healthy by eating geckos.
 They say that zapping yourself with a wand of undead turning is stupid.
+They say the Furies are more furious if you've been sinning.
 They say the Wizard's castle is booby-trapped!
 They say the gods get angry if you kill your dog.
 They say the gods get angry if you pray too much.
--- a/include/align.h
+++ b/include/align.h
@@ -60,4 +60,11 @@ typedef struct align { /* alignment & record */
 #define Msa2amask(x) (((x) == 3) ? 4 : (x))
 #define MSA_NONE    0  /* unaligned or multiple alignments */
 
+/* alignment change reasons for uchangealign(attrib.c) */
+enum uchangealign_reasons {
+    A_CG_CONVERT  = 0, /* permanently converted */
+    A_CG_HELM_ON  = 1, /* donned helm of opposite alignment */
+    A_CG_HELM_OFF = 2, /* doffed helm of opposite alignment */
+};
+
 #endif /* ALIGN_H */
--- a/include/align.h
+++ b/include/align.h
@@ -10,6 +10,7 @@ typedef schar aligntyp; /* basic alignment type */
 typedef struct align { /* alignment & record */
     aligntyp type;
     int record;
+    unsigned abuse;
 } align;
 
 /* bounds for "record" -- respect initial alignments of 10 */
--- a/include/extern.h
+++ b/include/extern.h
@@ -1341,6 +1341,7 @@ extern int mbirth_limit(int);
 extern void mkmonmoney(struct monst *, long);
 extern int bagotricks(struct obj *, boolean, int *);
 extern boolean propagate(int, boolean, boolean);
+extern void summon_furies(int);
 
 /* ### mcastu.c ### */
 
--- a/include/extern.h
+++ b/include/extern.h
@@ -1663,6 +1663,7 @@ extern void copy_mextra(struct monst *, struct monst *);
 extern void dealloc_mextra(struct monst *);
 extern boolean usmellmon(struct permonst *);
 extern void mimic_hit_msg(struct monst *, short);
+extern void adj_erinys(unsigned);
 
 /* ### mondata.c ### */
 
--- a/include/monsters.h
+++ b/include/monsters.h
@@ -2489,12 +2489,14 @@
        and spelled this way */
     MON("erinys", S_DEMON, LVL(7, 12, 2, 30, 10),
         (G_HELL | G_NOCORPSE | G_SGROUP | 2),
+        /* erinys attacks (among other things) are variable depending on your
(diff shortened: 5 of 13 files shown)


entrez:
This would break saves.  I have another PR open that breaks saves which
increments EDITLEVEL in the PR itself, #1140, but I didn't change it
this time because I didn't want to be continually changing it if some
other EDITLEVEL bumps come down the pipe while the PR is open.

pat-rankin:
Doesn't this let someone summon an erinys as soon as they find a helm of opposite alignment?  And assuming they survive, repeat two more times.

Survival is far from guaranteed though.  It also turns a very early helm of opposite alignment that the hero tries on become almost certain death.

nhmall:
This is not a comment on the pull request itself, just my opinion about what would be optimal procedure.
> I have another PR open that breaks saves which
> increments EDITLEVEL in the PR itself, #1140, but I didn't change it
> this time because I didn't want to be continually changing it if some
> other EDITLEVEL bumps come down the pipe while the PR is open.

I think taking steps to draw attention to the requirement to bump EDITLEVEL in include/patchlevel.h is the smoothest way to go, and leave out include/patchlevel.h from any pull requests. It has a high probability of being outdated by the time the pull request is merged.

I've also added a **_++EDITLEVEL when merging_** label that can be applied, although I cannot guarantee it will be applied in every case, or applied by all developers.

entrez:
@pat-rankin Yes, it would allow the hero to summon an erinys at will (which could be a strategic decision to avoid their summoning on Astral).  Extremely early-game erinyes may not be a huge problem because they are generated peaceful if you haven't yet committed any alignment abuse, but it could definitely be dangerous.  The abuse-based scaling could be increased further so that their base level, speed, etc, pre-abuse is additionally reduced, so that it's possible to flee from them if you don a HoOA too early.

The erinyes summoning could also be restricted entirely to Astral, of course -- I thought it could be an interesting decision though, whether you want to deliberately don an HoOA repeatedly to clear out the erinyes in preparation for using one on Astral, but also lose all your protection etc.  If donning an HoOA significantly abused alignment it could be even riskier and more interesting, because each time you put it on you'd make the erinyes summons left to go more dangerous!  Maybe some would opt to only eliminate one of the three, etc.  I didn't add this out of concern for its effects on early-game accidental HoOA use -- but it could be something that only abuses your alignment after the erinys is summoned, so only affects subsequent erinys generation, and only increments 'abuse' without decrementing 'record' (since it wouldn't make much sense that wearing the helm harms your alignment record with the god you've just converted to).  Basically just spitballing there.

entrez:
I just force-pushed to restore a comment in uchangealign(attrib.c) that
I had apparently deleted by mistake.  I didn't make any other changes.

entrez:
Pat, to your point about early-game HoOA use, maybe it could be changed so putting on a helm of opposite alignment just has a (alignment abuse-based?) chance of summoning an erinys, if you're not on Astral.  That would make it more challenging to eliminate all erinyes and make it less "guaranteed insta-death" for the early game.  It could also work with the HoOA alignment abuse idea I mentioned in my previous comment (i.e. it would mean subsequent uses of a helm would have increased likelihood of summoning an erinyes each time if the chance is based on abuse, but also make them more dangerous).

NullCGT:
This is a really interesting and flavorful change. Is there any way to communicate the changes in Erinyes strength to the player? It seems like it could be fairly unintuitive for unspoiled players to have Erinyes be much stronger or weaker with no indication of why.

k21971:
> Is there any way to communicate the changes in Erinyes strength to the player? It seems like it could be fairly unintuitive for unspoiled players to have Erinyes be much stronger or weaker with no indication of why.

I don't believe that would be necessary. Players will learn via interaction. There's no in-game indication that elementals are stronger on their home plane; players figured this out as they played. I think the same logic would apply here.

entrez:
I'd figure that if you've played enough NetHack to be surprised at the variation in the difficulty, a savvy encyclopedia-using player could likely come up with a hypothesis that their "wrath" is somehow based on how many "deeds unjust and base" the hero has committed, based on the encyclopedia entry in the PR.  If you're at that point as an unspoiled player you are probably very familiar with the encyclopedia and how to use it to make guesses about the game, since I think significant variation in difficulty that is clearly not tied to xlvl would require reaching the lategame across multiple different games.  And I think puzzling out the quote encyclopedia entries and how they might inform you about what you're looking at is a really fun part of being unspoiled about some aspect of the game.

nhmall:
A clever true rumor can't hurt though, if you can come up with one.

entrez:
Oh yeah, that's a great idea.  I'll try to think of something.  I guess I really only think about the encyclopedia since my recent "unspoiled" experience is with changes in variants, and trying to get relevant true rumors there is tough because they're massively outnumbered by vanilla rumors, even if the variant dev keeps the rumors files updated.  So I basically rely on the encyclopedia.

entrez:
Last force-push there was just to resolve merge conflicts from the recent extern.h changes.  (The previous force-push was to remove some formatting fixes that were already merged in 275713b, and didn't introduce any material changes to the patch either.)

paxed:
the HoOA thing doesn't really feel right to me, not sure ... I'd also want to lean towards the shadow and darkness aspects of Erinyes from mythology

entrez:
What are the shadow and darkness aspects of the Furies in mythology?  I'm not familiar with that.

paxed:
> What are the shadow and darkness aspects of the Furies in mythology? I'm not familiar with that.

https://en.wikipedia.org/wiki/Erinyes#Description

entrez:
Interesting.  I'm sure that could be cool -- right now there is basically nothing about them other than their number, so anything at all would be a flavor improvement.  There's also no reason these two concepts would have to be mutually exclusive.

The reason I think the HoOA change could be good is that it introduces an additional thing to deal with on Astral if going for an altar other than your own, one that can be tackled in a number of different ways -- if you are planning to use a HoOA on Astral, do you want to risk making the erinyes more dangerous for the chance to potentially eliminate some of them early, do you instead want to try to be careful to avoid alignment abuse so that they aren't particularly strong when you fight them (and/or are weak enough that it will be trivial to use a unicorn horn and #offer the amulet without fighting them), or do you want to spend a significant amount of time trying to generate them normally without buffing them the way HoOA use would?  Do you have a sense of your alignment abuse so far, to know which options are viable?  If you end up on the wrong altar on Astral and have a wish, do you want to risk fighting them or continue to the other altars?  Those are interesting decisions that can add layers of short- and long-term strategy to the end game, which often feels lacking in strategic choices like that by the time you've assembled an ascension kit.  And I think it does make flavor sense that HoOA use can trigger that kind of divine vengeance, given the way conversion is already treated as an especially serious betrayal of your deity in the game.  But it's not necessarily an integral part of the patch (though the reason I wrote all this out is in its defense is because I think it's would add dynamism and interesting decisions to make).

entrez:
I will say, though, I think this is a significant enough change and with enough of a delicate sort of balance that it would probably be better implemented in a variant and tested that way first rather than sticking it directly into NetHack "raw".  I think K2 was saying he may implement it in EvilHack -- that's not very equivalent to vanilla in terms of power levels of the hero and monsters, so probably it would settle on different numbers, but it could be interesting to see how well the general concept works there.

PR #1154, opened 2023-11-29
Don't grant tame djinni from lamps/potions if player is petless.

elunna:
From xNetHack git commit 9ee7a0e:

'This is similar to the Astral angel not appearing if the player is petless. Strictly speaking, players attempting to play petless could control this simply by not rubbing any lamps or quaffing any smoky potions, but that has a rather large impact on the game just for maintaining a conduct.

It doesn't change the overall probability of a hostile or wish-granting djinni; a petless player will simply get the neutral "You freed me!" djinni twice as often.

This could technically deny a djinn pet from players who actually wanted one, but since such players would have to have the pet option turned off and acquire no other pets along the way to making the djinn come out, it's probably an extreme edge case.'

### DIFF
--- a/src/potion.c
+++ b/src/potion.c
@@ -2778,9 +2778,14 @@ djinni_from_bottle(struct obj *obj)
         mongrantswish(&mtmp);
         break;
     case 1:
-        verbalize("Thank you for freeing me!");
-        (void) tamedog(mtmp, (struct obj *) 0);
-        break;
+        /* if the player is trying to play petless, make it safe
+           for them to rub lamps */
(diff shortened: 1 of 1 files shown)


pat-rankin:
Conducts are not supposed to change the way the program works.  There are a couple of exceptions (the one that comes to mind is requiring confirmation to read a scroll of mail if doing so will break illiterate conduct).  In this case, I think that attempting to adhere to petless conduct is more interesting if it results in making using magic lamps and smoky potions be risky.

elunna:
Ok that makes sense - thanks for the explanation!

PR #1155, opened 2023-11-30
Fix: msg_window:combination and TTY getlin() ^P

entrez:
The ability to use ^P from within a getlin prompt didn't work with
msg_window:combination: it's a combination (unsurprisingly) of 'full'
and 'single', but getlin() was treating it like 'full': i.e.  expecting
everything to be displayed at once, doprev_message() to return only once
the user was finished reading.  I didn't even know ^P was supposed to be
accessible from a getlin prompt until Pat's recent commit 5120764,
because I have always used msg_window:combination -- I had come up with
a system where I'd name a potion thrown at me "xx", check the previous
messages with ^P, and then rename it from the discoveries list.  This
will be a lot easier!

### DIFF
--- a/win/tty/getline.c
+++ b/win/tty/getline.c
@@ -48,7 +48,7 @@ hooked_tty_getlin(
     register char *obufp = bufp;
     register int c;
     struct WinDesc *cw = wins[WIN_MESSAGE];
-    boolean doprev = 0;
+    boolean doprev = FALSE;
 
     if (ttyDisplay->toplin == TOPLINE_NEED_MORE && !(cw->flags & WIN_STOP))
         more();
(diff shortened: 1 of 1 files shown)


entrez:
By the way, I noticed that because of the introduction of bot_disabled,
the use of ^P from a getlin prompt with msg_window:full, reverse, or
combination (after this patch and once you get to the point that it's
acting like msg_window:full) can cause the status line to be
overwritten/partially cleared by the message window, and then it won't
be redrawn until the prompt is finished.  Maybe that is known already
and accepted as worth it for the benefits of not having the status line
updates happen during a prompt (which seems reasonable, it's not a very
big deal) but I thought I'd mention it.

nhmall:
This is now in the pipeline.

PR #1160, opened 2023-12-04
remove unnecessary null-check on bhitm()

argrath:
`otmp` here is always non-null, otherwise it leads segv at earlier code.


nhmall:
In the pipeline

PR #1161, opened 2023-12-04
Ice and fumbling-related tweaks

entrez:
This came out of an IRC discussion about how ice uses fumbling internally, and it can cause some unexpected effects.

- Fix: fumbling vs losing footing in earthquake
- Refine ice fumbling effects vs mounted hero
- Check whether steed will slip on ice if mounted


nhmall:
In the pipeline

PR #1163, opened 2023-12-05
sp_lvl: fortify selection_getbounds

mkuoppal:
All callsites for selection_getbounds use NhRect from stack. If selection is not set, remember to set all rect values (to zero) as the callsites do not check for selection before iterating thru NhRect.

This prevents random values from stack ending up being used in temporary NhRects.

### DIFF
--- a/src/sp_lev.c
+++ b/src/sp_lev.c
@@ -4547,8 +4547,12 @@ selection_clone(struct selectionvar *sel)
 void
 selection_getbounds(struct selectionvar *sel, NhRect *b)
 {
-    if (!sel || !b)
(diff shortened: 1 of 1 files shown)


mkuoppal:
NhRect rect = { 0, }; would have worked also?

But thanks for taking a look.
-Mika

nhmall:
Thanks for drawing attention to the potential issue.

nhmall:
> NhRect rect = { 0, }; would have worked also?

Yes, on the initializers, as in this case. 
cg.zeroNhRect can now be used on initializers, and also in assignment statements.

PR #1168, opened 2023-12-07
split 'nothing_special' of arti_invoke() into separate function

argrath:


pat-rankin:
This could be simplified by making nothing_special() return 'ECMD_TIME' and its callers use 'return nothing_special()' instead of having separate return after each call.

argrath:
Although it can reduce code size, I don't think it is 'simplify.'
nothing_special() is not relevant to the return value of arti_invoke(), so nothing_special() should not have any information about that.

entrez:
I don't see the problem with the goto here.  It consolidates the early return behavior without this massive proliferation of single-purpose functions that just print a message, are used within one other function, and do nothing except provide an excuse not to use gotos.  A lot of these PRs feel to me like a sort of cargo cult goto removal because "gotos are bad" rather than fixing some serious readability issue this particular goto is causing.

argrath:
> I don't see the problem with the goto here.

No.
These gotos prevent further refactoring.

> It consolidates the early return behavior without this massive proliferation of single-purpose functions that just print a message, are used within one other function, and do nothing except provide an excuse not to use gotos.

No.
After this PR is applied, we can split arti_invoke() easily.
I'll submit some following PRs when this PR is merged.

> A lot of these PRs feel to me like a sort of cargo cult goto removal because "gotos are bad" rather than fixing some serious readability issue this particular goto is causing.

No.
My intention is not "gotos are bad", but "gotos (or anything) preventing further refactoring are bad."

argrath:
Resolved conflicts and force-pushed.

PR #1175, opened 2023-12-13
Corpsenm fixes

mkuoppal:
This pull requests fixes handful of callsites where corpsenm was used
into the permanent monster array. As corpsenm for TINs and EGGs can be -1
 (reported by ubsan), we need to adjust with >= LOW_PM check as adviced
by the obj.h:

    /* note: sometimes eggs and tins have special corpsenm values that
       shouldn't be used as an index into mons[]                       */

Please consider all or cherrypicking one-by-one basis.

### DIFF
--- a/src/eat.c
+++ b/src/eat.c
@@ -57,7 +57,8 @@ static int tin_ok(struct obj *);
    they can be sacrificed regardless of age which implies that they never
    become rotten */
 #define nonrotting_corpse(mnum) \
-    ((mnum) == PM_LIZARD || (mnum) == PM_LICHEN || is_rider(&mons[mnum]) \
+    ((mnum) == PM_LIZARD || (mnum) == PM_LICHEN \
+     || ((mnum) >= LOW_PM && is_rider(&mons[mnum])) \
      || (mnum) == PM_ACID_BLOB)
 
 /* non-rotting non-corpses; unlike lizard corpses, these items will behave
--- a/src/mon.c
+++ b/src/mon.c
@@ -11,6 +11,7 @@ static void sanity_check_single_mon(struct monst *, boolean, const char *);
 static struct obj *make_corpse(struct monst *, unsigned);
 static int minliquid_core(struct monst *);
 static void m_calcdistress(struct monst *);
+static boolean mstoning(const struct obj *obj);
 static boolean monlineu(struct monst *, int, int);
 static long mm_2way_aggression(struct monst *, struct monst *);
 static long mm_aggression(struct monst *, struct monst *);
@@ -1156,9 +1157,19 @@ meatbox(struct monst *mon, struct obj *otmp)
     }
 }
 
-#define mstoning(obj)                                       \
-    (ofood(obj) && (touch_petrifies(&mons[(obj)->corpsenm]) \
-                    || (obj)->corpsenm == PM_MEDUSA))
+static boolean
+mstoning(const struct obj *obj)
+{
+    const int nm = obj->corpsenm;
+
+    if (nm < LOW_PM)
+        return FALSE;
+
+    if (nm == PM_MEDUSA)
+        return TRUE;
+
+    return ofood(obj) && touch_petrifies(&mons[nm]);
+}
 
 /* monster consumes an object.
 
@@ -1417,8 +1428,6 @@ meatobj(struct monst* mtmp) /* for gelatinous cubes */
     }
     return (count > 0 || ecount > 0) ? 1 : 0;
 }
-
-#undef mstoning
 
 /* Monster eats a corpse off the ground.
  * Return value is 0 = nothing eaten, 1 = ate a corpse, 2 = died */
--- a/src/uhitm.c
+++ b/src/uhitm.c
@@ -1156,7 +1156,8 @@ hmon_hitmon_misc_obj(
                 change_luck(-5);
         }
 
-        if (touch_petrifies(&mons[obj->corpsenm])) {
+        if (obj->corpsenm >= LOW_PM
+            && touch_petrifies(&mons[obj->corpsenm])) {
             /*learn_egg_type(obj->corpsenm);*/
             pline("Splat!  You hit %s with %s %s egg%s!",
                   mon_nam(mon),
--- a/src/weapon.c
+++ b/src/weapon.c
@@ -469,17 +469,24 @@ static struct obj *oselect(struct monst *, int);
     } while (0)
 
 static struct obj *
-oselect(struct monst *mtmp, int x)
+oselect(struct monst *mtmp, int type)
 {
     struct obj *otmp;
 
     for (otmp = mtmp->minvent; otmp; otmp = otmp->nobj) {
-        if (otmp->otyp == x
-            /* never select non-cockatrice corpses */
-            && !((x == CORPSE || x == EGG)
-                 && !touch_petrifies(&mons[otmp->corpsenm]))
-            && (!otmp->oartifact || touch_artifact(otmp, mtmp)))
-            return otmp;
+        if (otmp->otyp != type)
+            continue;
+
+        /* never select non-cockatrice corpses */
+        if ((type == CORPSE || type == EGG)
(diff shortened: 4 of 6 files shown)


src/uhitm.c
```diff
@@ -1150,7 +1150,8 @@ hmon_hitmon_misc_obj(
                 change_luck(-5);
         }
 
-        if (touch_petrifies(&mons[obj->corpsenm])) {
+        if (obj->corpsenm >= LOW_PM &&
+            touch_petrifies(&mons[obj->corpsenm])) {
```
entrez: This is another case where the wrapping should happen before the operator:

```c
if (obj->corpsenm >= LOW_PM
    && touch_petrifies(&mons[obj->corpsenm]))
```

src/weapon.c
```diff
@@ -475,8 +475,8 @@ oselect(struct monst *mtmp, int x)
     for (otmp = mtmp->minvent; otmp; otmp = otmp->nobj) {
         if (otmp->otyp == x
             /* never select non-cockatrice corpses */
-            && !((x == CORPSE || x == EGG)
-                 && !touch_petrifies(&mons[otmp->corpsenm]))
+            && !((x == CORPSE || x == EGG) &&
+		 !(otmp->corpsenm >= LOW_PM && touch_petrifies(&mons[otmp->corpsenm])))
```
entrez: Same formatting issue here with the trailing '&&'.

src/mon.c
```diff
@@ -1156,9 +1156,18 @@ meatbox(struct monst *mon, struct obj *otmp)
     }
 }
 
-#define mstoning(obj)                                       \
-    (ofood(obj) && (touch_petrifies(&mons[(obj)->corpsenm]) \
-                    || (obj)->corpsenm == PM_MEDUSA))
+static boolean mstoning(const struct obj *obj)
```
entrez: Normally this would be formatted with the type on its own line, like this:

```c
static boolean
mstoning(const struct obj *obj)
{
```

You'd normally also add a function prototype in the appropriate position (between whatever functions come before and after it) at the top of the file.

src/eat.c
```diff
@@ -58,7 +58,8 @@ static int tin_ok(struct obj *);
    they can be sacrificed regardless of age which implies that they never
    become rotten */
 #define nonrotting_corpse(mnum) \
-    ((mnum) == PM_LIZARD || (mnum) == PM_LICHEN || is_rider(&mons[mnum]) \
+    ((mnum) == PM_LIZARD || (mnum) == PM_LICHEN || \
+     (mnum >= LOW_PM && is_rider(&mons[mnum])) \
```
entrez: NetHack code style would be like this, with the wrapping happening before the operator:

```c
#define nonrotting_corpse(mnum) \
    ((mnum) == PM_LIZARD || (mnum) == PM_LICHEN \
     || (mnum >= LOW_PM && is_rider(&mons[mnum])) \
     || (mnum) == PM_ACID_BLOB)
```

entrez (commented):
There are a few deviations from the normal code style described in DEVEL/code_style.txt I noticed and wanted to point out.  (Apparently they're all out of order because I edited them at different times, but you get the gist I'm sure :-)

src/eat.c
entrez: Oh, also, mnum should be in parentheses in case the macro is called on some more complicated thing that could theoretically complicate the order of operations I guess.

```c
     || ((mnum) >= LOW_PM && is_rider(&mons[mnum])) \
```

src/weapon.c
entrez: Also, I think this line also ends up being too long, and should be split up:

```c
            && !((x == CORPSE || x == EGG)
                 && !(otmp->corpsenm >= LOW_PM
                      && touch_petrifies(&mons[otmp->corpsenm])))
```

Actually, I think it might make more sense to organize it like this, since it makes it easier to understand it's going through a list of reasons to exclude objects rather than actual positive requirements for the object being selected, but I understand that may be out of scope for this PR:
```c
    for (otmp = mtmp->minvent; otmp; otmp = otmp->nobj) {
        if (otmp->otyp != x)
            continue;
        /* never select non-cockatrice corpses */
        if ((x == CORPSE || x == EGG)
            && (otmp->corpsenm == NON_PM
                || !touch_petrifies(&mons[otmp->corpsenm])))
            continue;
        if (otmp->oartifact && !touch_artifact(otmp, mtmp))
            continue;
        return otmp;
    }
```

src/uhitm.c
mkuoppal: Thanks entrez for reading and advice. I will redo the pull request with changes. I have done bulk of code for completely different style so should have been more careful to investigate the style.

entrez: Cool, these seem like good catches to me.  I'm surprised it hasn't caused any kind of warnings or problems before -- makes me wonder if there's already some reason that it never reaches the point where a normal chicken egg gets evaluated in these functions.  Maybe something totally incidental like "monsters never pick up chicken eggs or generate with them in inventory".  (Even if that's the case, it still seems worth fixing it to make it more robust against possible future changes.)

src/mon.c
mkuoppal: As this is static, what is the benefit of having early prototype? Regarless added and noted that quite amount of prototypes are missing here in general.

src/uhitm.c
mkuoppal: I did add asan+ubsan and then did quite amount of debugfuzzing. ubsan noticed -1 index access and started to look it through.

entrez: Oh, interesting, I've never heard of ubsan before.  I will have to try that.

src/weapon.c
mkuoppal: Thinking about this, should it be?
                || touch_petrifies(&mons[otmp->corpsenm])))
instead of
                || !touch_petrifies(&mons[otmp->corpsenm])))

entrez: The second one with the '||'?  Based on the comment, the idea is that it's skipping all corpses and eggs except those which can be used to petrify an enemy, so it should skip any '!touch_petrifies' items.  So I think ` if (!touch_petrifies) continue;` would be right.  The original code is difficult to parse because it has the two negations at different depths like `if (!(!touch_petrifies)) return otmp;` but I think it does the same thing.

mkuoppal: Oh yes, double negative in original. The comment and your explanation makes sense. So the monster wants to grab something of value =). Thanks!

src/uhitm.c
mkuoppal: -fsanitize=undefined -lubsan
linking util/makedefs just needs a extra  massaging

src/mon.c
entrez: I don't know.  It may not be necessary, I have just noted it seems to be usually done and figured it may be some kind of compatibility thing (and/or a holdover from ages long past, maybe).  Maybe a DevTeam member might know if it's actually something that is held to be generally required (stylistically or otherwise).

src/weapon.c
```diff
@@ -468,17 +468,21 @@ static struct obj *oselect(struct monst *, int);
...
-            && (!otmp->oartifact || touch_artifact(otmp, mtmp)))
-            return otmp;
+        if (otmp->otyp != type)
+            continue;
+        /* never select non-cockatrice corpses */
+        if ((type == CORPSE || type == EGG)
+            && (otmp->corpsenm == NON_PM
+                || !touch_petrifies(&mons[otmp->corpsenm])))
+            continue;
+        if (otmp->oartifact || !touch_artifact(otmp, mtmp))
```
entrez: This should be

```c
        if (otmp->oartifact && !touch_artifact(otmp, mtmp))
```

(Actually, maybe it should really be `if (!can_touch_safely(mtmp, otmp))` since that handles other cases like silver-haters vs silver weapons, in addition to `touch_artifact`... Not sure if there's a reason that isn't used or not.)

mkuoppal: Yup, can_touch_safely seems to contain the artifact check so included that one.

PR #1177, opened 2023-12-14
Saving a bones file should not free memory; the function really_done …

janne-hmp:
…will be using that information after the call to savebones, resulting in a heap-use-after-free error (and possibly later in a double-free in nh_terminate if things get that far).

### DIFF
--- a/src/bones.c
+++ b/src/bones.c
@@ -575,7 +575,7 @@ savebones(int how, time_t when, struct obj *corpse)
     }
     c = (char) (strlen(bonesid) + 1);
 
-    nhfp->mode = WRITING | FREEING;
+    nhfp->mode = WRITING;
     store_version(nhfp);
     store_savefileinfo(nhfp);
     if (nhfp->structlevel) {


entrez:
What data does really_done use that's freed there?  `WRITING | FREEING` doesn't actually free as much as `FREEING` does (the relevant functions distinguish between the two with `mode == FREEING` vs `mode & FREEING != 0`), and I thought at that point the stuff that gets freed is pretty peripheral.  And I've never encountered an asan error from saving bones as far as I can remember (but maybe that's due to some kind of limitation of the version of asan I'm using?).

janne-hmp:
I didn't actually check the 3.7 implementation of FREEING in detail, since ASAN captures the heap-use-after-free in 3.6.2-based variant, which I have been using, and I assumed that the implementation in 3.7 is analogous. In any case, savelev saves (and frees) the entire current level, including in the items on the floor. These may have been used while making the dumplog later, although I realize now that this may in fact apply only to EvilHack-style HTML Dumplogs that use floor items in map tooltips, which 3.7 is not using currently. So perhaps this is just an advance warning in the case somebody wants to add the HTML dumplog later on. Currently elsewhere in really_done just the character inventory items are being used, and thus there is no problem, but I would assume that freeing dynamic data at the very end would still result in a double-free? But perhaps this is accounted for somehow in the 3.7 implementation, e.g. by setting the relevant pointers to zero. In any case, it still seems to me that it would make more sense to free all the data just at the end in nh_terminate, rather than some of it in the middle of the process in savebones.

entrez:
Oh yeah, that makes sense that it would be a problem for HTML dumplogs.

PR #1179, opened 2023-12-15
add sanity check on choose_classes_menu()

argrath:
If class_list contains an illegal char for mon/obj class (even if it should not happen), it might cause out-of-bound access.


entrez:
There should be a space between "if" and the opening parenthesis.

argrath:
Thank you for pointing it out.
Fixed it and force-pushed.

PR #1184, opened 2023-12-20
Make pets unwilling to eat all polymorphing corpses

### LINKED ISSUE #1183: Pets eating genetic engineers
https://github.com/NetHack/NetHack/blob/0713b91beb6fd29234cfb06ef48ce7adbd0526cb/src/dog.c#L991-L995

The check for pets eating corpses that polymorph them doesn't seem to include genetic engineers at all - possibly an oversight, and a reasonable enough one considering the genetic engineer isn't a shapeshifter itself. A yellow dragon of mine just became a lichen from doing so (which is infinitely more annoying than it eating the corpse to begin with because isn't _that_ stereotypical), which prompted me to make this issue.

entrez:
Don't limit the hestitation to eat a corpse to shapeshifter corpses; apply it
to all corpses that will polymorph the creature eating them.  Fixes #1183.

- Rename polyfodder() to polyfood()
- Make pets unwilling to eat all polyfood() corpses

### DIFF
--- a/include/obj.h
+++ b/include/obj.h
@@ -303,7 +303,7 @@ struct obj {
 #define ofood(o) ((o)->otyp == CORPSE || (o)->otyp == EGG || (o)->otyp == TIN)
     /* note: sometimes eggs and tins have special corpsenm values that
        shouldn't be used as an index into mons[]                       */
-#define polyfodder(obj) \
+#define polyfood(obj) \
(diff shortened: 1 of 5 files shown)


entrez:
This has been ignored for a while but I just noticed that it included a typo (hence the force push).  Maybe that was the issue?

PR #1188, opened 2023-12-31
Fix out of bound access in uhpinc and ueninc.

mkuoppal:
uhpinc and ueninc are indexed by u.ulevel. ulevel can reach MAXULEVEL (30), which means that the 0-29 range of uhpinc and ueninc arrays is not enough and out of bounds by one will result:

exper.c:245:25: runtime error: index 30 out of bounds for type 'xint16 [30]'
exper.c:263:25: runtime error: index 30 out of bounds for type 'xint16 [30]'

Fix this by increasing both arrays by one, which allows u.ulevel indexing (and leaving index 0 as unused)

### DIFF
--- a/include/you.h
+++ b/include/you.h
@@ -463,8 +463,8 @@ struct you {
         uhppeak;             /* highest value of uhpmax so far */
     int uen, uenmax,         /* magical energy, aka spell power */
         uenpeak;             /* highest value of uenmax so far */
-    xint16 uhpinc[MAXULEV],  /* increases to uhpmax for each level gain */
-          ueninc[MAXULEV];   /* increases to uenmax for each level gain */
+    xint16 uhpinc[MAXULEV + 1],  /* increases to uhpmax for each level gain */
+          ueninc[MAXULEV + 1];   /* increases to uenmax for each level gain */
     int ugangr;              /* if the gods are angry at you */
     int ugifts;              /* number of artifacts bestowed */
     int ublessed, ublesscnt; /* blessing/duration from #pray */


nhmall:
I’m not sure that’s the right, or most comprehensive, fix, even though it prevents the out-of-bounds. Aren’t there some for-loops that search those arrays beginning with index 0?

nhmall:
These look like they may be outstanding issues post-PR:

https://github.com/NetHack/NetHack/blob/cd91d0630b4aac8b502713715b29a696ece5d765/src/polyself.c#L392

https://github.com/NetHack/NetHack/blob/cd91d0630b4aac8b502713715b29a696ece5d765/src/polyself.c#L377

mkuoppal:
src/polyself:newman() will loop using these arrays:

for (i = 0; i < oldlvl; i++)

where oldlvl = u.ulevel;

Assert could be added there to make sure it is <= MAXULEVEL, as it should.

mkuoppal:
It seems that the code is designed around [0-29] everywhere else but exper.c.
So yes, this is nothing but a out of bounds fix. exper.c should be molded around [0-29] indexing.

nhmall:
We don’t want to hide an indexing error by moving the boundary to include the errant index.

mkuoppal:
Yes, I will try to figure out how the oob was possible in the first place.

nhmall:
[...]
Thank you for bringing the matter to our attention.

nhmall:
I have opened issue #1189

PR #1196, opened 2024-01-10
early return when kicking a door on dokick()

argrath:


argrath:
It seems that this PR has served its purpose.  Closing.

PR #1199, opened 2024-01-12
Guard against out of bounds access to level locations

mkuoppal:
Replace levl macro and direct access to the room locations with a loc(x, y) macro which will direct 
access into a static inline function with type and optionally, if DEBUG is enabled, boundary checking.

This will reveal out of bounds access to the gl.level.locations

I have found multiple callsites that do access rm[] with y >= 21  (ROWNO)  and seen also -1 as index.
Obviously future work could be to extend this kind of hardening to objects and mon array accesses.

I dont know if this kind of large patch causing churn is desirable in this stage of development. In that case this can be used as out of tree debugging aid if someone is interested of hunting the all callsites down. And to handle the varying callsites in 'Issues', one backtrace per issue.

debugfuzzer will usually find out of bound access within a minute or two with this applied.

Thoughts?

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -216,8 +216,8 @@
  * "cover" any objects or traps that might be there.
  */
 #define covers_objects(xx, yy) \
-    ((is_pool(xx, yy) && !Underwater) || (levl[xx][yy].typ == LAVAPOOL) \
-      || (levl[xx][yy].typ == LAVAWALL))
+    ((is_pool(xx, yy) && !Underwater) || (loc(xx, yy)->typ == LAVAPOOL)  \
+     || (loc(xx, yy)->typ == LAVAWALL))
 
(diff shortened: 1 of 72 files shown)


mkuoppal:
debugfuzzer will usually find out of bound access within a minute or two with this applied.

paxed:
I'm not opposed to this change, although it is a very big one - perhaps it should be split into smaller chunks?

Also, there should be fixes (in a separate PR) at least for the most common problems exposed by this change; otherwise the game could be unplayable due to the asserts.

mkuoppal:
UBSAN makes this pull request redundant atleast what comes to guarding array accesses to map locations:

https://github.com/NetHack/NetHack/commit/20f8af224b536e8e883269639c47f8792952980b

PR #1216, opened 2024-03-01
Fix buffer overflow in crashreport_init

liferooter:


liferooter:
Just realized that this bug leads to almost no consequences, because number of characters written is always less that buffer size, so calling `snprinft` with wrong buffer size actually doesn't lead to real buffer overflow in that case. It leads to skip of two characters in binary ID when passed buffer size is zero, leaving garbadge in their place, but not buffer overflow. It explains why this bug wasn't yet found.

liferooter:
I've occasionally built NetHack with sanitizers or something like that, so this bug crashed the game.

Issue #1157, opened 2023-11-30
Monsters zapping wand of striking doesn't identify the wand


elunna:
To reproduce: Summon a large kobold and give it a wand of striking. When it hits you it doesn't identify as most wands normally do. Other ray wands don't seem to have this problem, I tested wands of fire and magic missile and both id'd on seeing the ray shot.

I also noticed that with magic resistance, monsters keep zapping you with the wand of striking even though M_SEEN_MAGR should have been set. I also checked if this happened with wands of magic missile, but that behavior was fine.

I also tested the wand of undead turning, since it's another beam wand, but that one did auto-identify when a monster zapped me with it and it revived a corpse.

PR #1223, opened 2024-03-11
followup fix for github issue #1157.

elunna:
For some reason the cansee check wasn't working to see the monster zapping the wand. Changed it to canseemon and tested getting zapped and it auto-identifies now.

### DIFF
--- a/src/muse.c
+++ b/src/muse.c
@@ -1562,7 +1562,7 @@ mbhitm(struct monst *mtmp, struct obj *otmp)
         } else {
             miss("wand", mtmp);
         }
-        if (learnit && cansee(mtmp->mx, mtmp->my) && gz.zap_oseen)
+        if (learnit && gz.zap_oseen)
             makeknown(WAN_STRIKING);
         break;
     case WAN_TELEPORTATION:


entrez:
Under what circumstances was it not identifying?  It never worked?

entrez:
Looking into it a bit further because "for some reason" sounded worth investigating in more depth, the issue here is actually that 'mtmp' in mbhitm() is the zapee, not the zapper (so it is &gy.youmonst in this scenario).  So this patch retains the underlying problem.

mbhit does have access to the zapper, and mbhitm currently always returns 0, so maybe one solution could be to change mbhitm to return the value of learn_it and then move the cansee check to the calling function which has access to the zapper.  Or else to just make the pointer to the zapper available within mbhitm.

Actually isn't this the reason for gz.zap_oseen?  It already checks whether you can see the zapper when it's set.  I think maybe the cansee() call should just be eliminated entirely and zap_oseen should be used on its own.

entrez:
So I think the fix should really be 

```diff
diff --git a/src/muse.c b/src/muse.c
index 5f5edc250..4b71ff360 100644
--- a/src/muse.c
+++ b/src/muse.c
@@ -1562,7 +1562,7 @@ mbhitm(struct monst *mtmp, struct obj *otmp)
         } else {
             miss("wand", mtmp);
         }
-        if (learnit && cansee(mtmp->mx, mtmp->my) && gz.zap_oseen)
+        if (learnit && gz.zap_oseen)
             makeknown(WAN_STRIKING);
         break;
     case WAN_TELEPORTATION:
```

unless there's some reason there needs to be a secondary visibility check beyond the one cached in zap_oseen.

elunna:
> Under what circumstances was it not identifying? It never worked?

I had noticed it recently just when being zapped by monsters.  Thanks for looking further into this, I had assumed mtmp was the monster (zapper) since the original code used it (my bad - never assume!), but yes that does look like the reason for zap_oseen existing.
```

  /* muse.c */
    boolean zap_oseen; /* for wands which use mbhitm and are zapped at
                        * players.  We usually want an oseen local to
                        * the function, but this is impossible since the
                        * function mbhitm has to be compatible with the
                        * normal zap routines, and those routines don't
                        * remember who zapped the wand. */
```

entrez:
IMO inserting the == 1 makes it read like you're testing an integer that can have a range of meaningful values when it's actually a boolean, so it's potentially a little more confusing than the existing check.  I would leave it off if it were me.

elunna:
Oh that should not have made it in, will fix.

PR #1224, opened 2024-03-13
Make the curses windowport able to interpret glyph color adjustment

FredrIQ:
The ncurses library only has 256 color support, but the curses windowport is limited even beyond this, to 16 colors. This commit allows curses to do a limited interpretation of the glyph user color adjustment to the basic NetHack colors. Two examples:

This works:

    OPTIONS=glyph:G_pet_female_kitten:U+0066/red

This does not since it tries to invoke a 24bit color:

    OPTIONS=glyph:G_pet_female_kitten:U+0066/255-0-0

Something I was also contemplating, unrelated to implementation of this support in curses, would be the ability for the following:

* allow defining colors if other symbol handling modes are used (possibly limited to the standard 16 colors)
* allow defining attributes (for example: glyph:G_pet_female_kitten:U+0066/red/underline)
* allow specifying glyphs as wildcards for defining global color/attribute changes

### DIFF
--- a/win/curses/cursmain.c
+++ b/win/curses/cursmain.c
@@ -913,6 +913,16 @@ curses_print_glyph(
     special = glyphinfo->gm.glyphflags;
     ch = glyphinfo->ttychar;
     color = glyphinfo->gm.sym.color;
+
+#ifdef ENHANCED_SYMBOLS
+    /* The curses library does not support truecolor, only the more limited 256
+       color mode. On top of this, the windowport only supports 16 color mode.
+       Thus, we only allow users to customize glyph colors to the basic NetHack
+       colors. */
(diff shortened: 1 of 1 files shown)


FredrIQ:
Something I also want to see are keywords for "don't change the current defined data". If this were to be added, you could for example do this:

    OPTIONS=glyph:G_*_fox:U+0064/blue
    OPTIONS=glyph:G_statue_*:basechar/gray/underline

for "make all foxes use a blue color, make all statues gray with underline" without needing to specify the relevant character for every statue.

This ("basechar", "basefg", etc) should perhaps also be added for MENUCOLORS and statushilites, so that you can, for example, underline all items being worn without needing to specify a bunch of near-duplicate rules for combining BUC colors + underline worn items. as per #1064

PR #1232, opened 2024-04-11
Un-defer shimmering dragons.

NullCGT:
A comment in the codebase indicates that shimmering dragons are deferred because monster displacement has not yet been implemented. Seeing as that is no longer the case I believe they should be added back in, particularly because shimmering dragon scale mail presents another interesting ascension kit item.

I added a bit of code to ensure that like displacer beasts, shimmering dragon and baby shimmering dragon corpses grant temporary displacement upon consumption. Corpses of full grown dragons grant a large number of turns of displacement in order to reward a player that deals with the additional challenge of one spawning in the castle, and also to present an interesting option for something to wish for before entering the planes.

### DIFF
--- a/include/mondata.h
+++ b/include/mondata.h
@@ -154,6 +154,9 @@
 #define infravision(ptr) (((ptr)->mflags3 & M3_INFRAVISION))
 #define infravisible(ptr) (((ptr)->mflags3 & M3_INFRAVISIBLE))
 #define is_displacer(ptr) (((ptr)->mflags3 & M3_DISPLACES) != 0L)
+#define is_displaced(ptr) ((ptr) == &mons[PM_DISPLACER_BEAST] \
+                            || (ptr) == &mons[PM_SHIMMERING_DRAGON] \
+                            || (ptr) == &mons[PM_BABY_SHIMMERING_DRAGON])
 #define is_mplayer(ptr) \
     (((ptr) >= &mons[PM_ARCHEOLOGIST]) && ((ptr) <= &mons[PM_WIZARD]))
 #define is_watch(ptr) \
--- a/include/monsters.h
+++ b/include/monsters.h
@@ -1360,8 +1360,6 @@
         M1_FLY | M1_THICK_HIDE | M1_NOHANDS | M1_CARNIVORE,
         M2_HOSTILE | M2_STRONG | M2_GREEDY | M2_JEWELS, 0,
         13, DRAGON_SILVER, BABY_SILVER_DRAGON),
-#if 0 /* DEFERRED */
-    /* [see "shimmering dragon" below] */
     MON(NAM("baby shimmering dragon"), S_DRAGON,
         LVL(12, 9, 2, 10, 0), G_GENO,
         A(ATTK(AT_BITE, AD_PHYS, 2, 6),
@@ -1370,7 +1368,6 @@
         M1_FLY | M1_THICK_HIDE | M1_NOHANDS | M1_CARNIVORE,
         M2_HOSTILE | M2_STRONG | M2_GREEDY | M2_JEWELS, 0,
         13, CLR_CYAN, BABY_SHIMMERING_DRAGON),
-#endif
     MON(NAM("baby red dragon"), S_DRAGON,
         LVL(12, 9, 2, 10, 0), G_GENO,
         A(ATTK(AT_BITE, AD_PHYS, 2, 6),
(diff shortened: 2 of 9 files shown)


loggersviii:
is there a secondary extrinsic for the DSM? displacement is cool but would be underpowered by itself methinks

NullCGT:
I was hesitant to add a secondary extrinsic in this PR. Displacement is fairly useful, and I wasn't sure what extrinsic I could add that fit thematically and did not make shimmering dragon scale mail very powerful.

Umbire:
I've been thinking on it a bit and similarly drawing a blank, wish I could come up with something flavorful and applicable - I think me and Loggers kinda stuck on stun resistance, but he feels that'd be hard to justify flavorwise.

vultur-cadens:
I don't think shimmering DSM necessarily needs a secondary property, though it could reasonably have one, if one could be found that is flavorful and not too strong.

The other DSMs with secondary intrinsics are have primary extrinsic effects that can be mostly (even if imperfectly) easily replicated via other means, so a secondary extrinsic is needed to give players incentive to use them.  (The other DSMs' primary extrinsics are mostly resistances that can be permanently gained intrinsically by eating the dragon corpse; exceptions are yellow DSM's acid resistance, which isn't very useful by the time you're able to deal with dragons, and gold DSM's light source, which is moderately useful but also has other sources that are common, such as lamps.)  The extrinsic item-protection buff in 3.7 does mean that the extrinsic resistances are slightly better than the intrinsics, but not by that much since non-metallic gloves and bags are effective defenses against item destruction.

Permanent displacement currently only has one source, the cloak of displacement, which is somewhat rare and takes up the cloak slot.  Since displacement is a property that some players like to have, I don't think shimmering DSM needs a secondary property to incentivize usage.

On the other hand, the cloak of magic resistance is rarer than the cloak of displacement when randomly generated (by a 1-to-5 ratio), and it is almost forced to be part of a player's kit if shimmering DSM is chosen (unless they're using Magicbane -- magic resistance from quest artifacts practically doesn't count due to Rodney's theft attack), so it may make some sense for the combination of shimmering DSM+cloak of magic resistance to be slightly better than gray DSM+cloak of displacement.  Another point could be made that displacement isn't always a desirable property to have on (angry bribable demons and such), and so displacement from cloak is better than displacement from DSM, because a cloak is much faster and safer to put on and take off than DSM.  A secondary extrinsic for the DSM could compensate for these factors.  I haven't decided whether I think stun resistance might be too strong when considered as compensation for CoMR rarity and armor inconvenience though.

loggersviii:
Honestly, it's as simple as shimmering scale mail in its current state not being competitive with gray or silver. This is the entire reason secondary extrinsics exist in the first place. If a dragon scale mail's primary extrinsic isn't on par with magic resistance/reflection, it should have a secondary extrinsic. 

If shimmering scale mail was released in its current state, you would not see it being used outside of novelty builds. Logically, no one would choose it over blue, gray, or silver mail. Even gold and orange are arguably more useful.

ratanvarghese:
Phasing would be nifty in Gehennom and the Plane of Earth. "You displace the wall."

loggersviii:
phasing is both badass and underrepresented. I second phasing

copperwater:
Phasing might bring shimmering dragon scale mail up to the level of other dragon scale mails... or surpass them. It's far too powerful for a permanent extrinsic on equipment - you'd be able to plow straight through levels designed to make you fight through them (skip the entire Valley, Castle, etc) and avoid fights by hiding in solid rock in complete safety to heal up (or merely fleeing to space on the other side of a wall while the monsters have to go around).

Currently the only way to phase involves putting up with the downsides of polyselfing and not being able to wear body armor or a cloak. Let's not add a way that removes those downsides and additionally makes it freely switchable on and off.

loggersviii:
it's funny how even though phasing lets you do all that, it still wouldn't make shimmering mail better than gray mail :P

ars3niy:
But comparing directly to GDSM is unfair because that's not what the choice really is. The choice is between an alternative DSM and a cloak or amulet you now cannot wear, or off-hand weapon you cannot use, due to having to get your MR or reflection elsewhere, or spending a wish on cloak of magic resistance vs. doing something else.

Is ~~intrinsic~~ phasing better than a robe or amulet of life saving or +6 silver saber?

NullCGT:
Not to derail the conversation about balance, but someone on the DevTeam reached out to me last month, and this PR may be moot. If I'm understanding the conversation I had correctly, it sounds like displacement as implemented for displacer beasts is different from the generalized monster displacement that would need to be implemented for shimmering dragons to be added.

loggersviii:
Arseniy: there's already DSM that have that problem, namely silver and blue. Coincidentally, they're also two of the three DSM that are frequently used. I'm approaching this from the angle of knowing the problem exists, and that it will not be fixed because to fix it the devteam would have to change gray and silver mail, which will not happen.

PR #1240, opened 2024-05-03
YANI: Brides of Dracula

NullCGT:
In the original Bram Stoker novel, Dracula has three brides, colloquially referred to as the [Brides of Dracula](https://en.wikipedia.org/wiki/Brides_of_Dracula). I thought it would be a nice touch for NetHack to quietly reflect this by ensuring three of the vampires of his court are vampire ladies. While this technically increases the difficulty of Vlad's Tower, the vampires in Vlad's court are not necessary to fight, and the experience level of a hero at this point in the game is likely to cause them to spawn as vampire lords or ladies in any case.

No established names exist for these characters, so I have left them unnamed.

### DIFF
--- a/dat/tower1.lua
+++ b/dat/tower1.lua
@@ -30,9 +30,10 @@ des.monster("Vlad the Impaler", 06, 05)
 des.monster("V",niches[1])
 des.monster("V",niches[2])
 des.monster("V",niches[3])
-des.monster("V",niches[4])
-des.monster("V",niches[5])
-des.monster("V",niches[6])
+-- The brides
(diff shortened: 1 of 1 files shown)


copperwater:
Is it ever established that the in-game Vlad == Count Dracula? (The historical figure's surname is Dracula, but it's not clear whether he's intended to be the same figure in NetHack.)

Worth noting that if vampire mages are ever undeferred then this will actually be a minor difficulty decrease for that level, since there are fewer opportunities to spawn them as a random V.

PR #1242, opened 2024-05-06
When making an invalid move (stepping diagonally out of doorway) still warned about traps in destination square

disperse:
Steps to reproduce the issue:
1. Place a trap diagonally from a doorway
2. Stand in the doorway
3. Move toward the trap

Expected results:
You can't move diagonally so nothing will happen.

Actual results:
You get the "Really step onto that fire trap/hole/etc." message


entrez:
Commit titles should be no more than 50 characters long, FYI (and lines should be wrapped at something like 72 characters). NetHack is very consistent about this.  I think this is a pretty good post explaining the reasoning behind sticking to it: https://cbea.ms/git-commit/

PR #1271, opened 2024-08-19
YANI: Add option to start without inventory, skills and spells

ars3niy:
You know how it goes – you have been heralded from birth to be a hero who saves the world and your hour of destiny has come, but you haven't prepared very well. You come from a poor family and could not afford any training or equipment, so the only thing you bring with you into this dungeon is your potential. Empty-handed and with only your underwear on you, go bravely with your god!

The proposed option (tentatively called "broke" here) makes you not only start with no inventory (for which you could just drop everything on turn 1) but also with every skill at Unskilled and no spells. You do, however, start with knowledge of all the objects you would normally be guaranteed to know, which is actually mostly because of the healer: without healing potions at least discovered they would have too little reason to call themselves a healer at all. A rogue still knows what a sack is, a wizard knows all level 1 spellbooks.

You may think that this makes all roles too similar since starting inventory is a big part of what makes each role what they are. However, even with such start every role still has its defining features: archeologist is good with touchstones (admittedly they may never find one) and can become expert in pick-axe, barbarian is strong and gets Cleaver, caveperson can eat cats and is good with spears, and so on.

Best used in conjunction with !bones, lest you find early bones of your own role. I've played several roles like this and found it quite refreshing.

This patch as-is has a problem: starting with no inventory seems to interfere with rendering when using curses, so one has to do something on turn 1 then hit Ctrl+R on turn 2 and then play normally. But because at this point I do not know if the developers or any variant developers are interested in including this, I'll leave it as it is but can fix should the need arise.

### DIFF
--- a/include/optlist.h
+++ b/include/optlist.h
@@ -213,6 +213,9 @@ static int optfn_##a(int, int, boolean, char *, char *);
                 No, Yes, No, No, NoAlias,
                 "deprecated (use S_boulder in sym file instead)")
 #endif
+    NHOPTB(broke, Advanced, 0, opt_in, set_in_config,
+           Off, Yes, No, No, NoAlias, &u.uroleplay.broke, Term_False,
+           "start game with no inventory or spells")
     NHOPTC(catname, Advanced, PL_PSIZ, opt_in, set_gameview,
                 No, Yes, No, No, NoAlias,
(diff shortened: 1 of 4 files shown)


paxed:
Implemented as "pauper" in commit e645c0b4bb38595e95990930e207d68118a96d1f

PR #1293, opened 2024-10-02
Gluten free conduct

mlehotay:
https://www.reddit.com/r/nethack/comments/1fsajaa/gluten_free_conduct/


vultur-cadens:
We already have three official food conducts, and this conduct seems too trivial compared to the other ones.  Vegan conduct is already two conducts in one, and this will make it three for vegan characters who simply add a "no cram ration" restriction (a trivial restriction for everyone but human and gnomish Rangers).

PR #1301, opened 2024-10-23
Remove magic members from instance globals

mkuoppal:
We ought to trust compiler and we have other
tools to check oob access.

### DIFF
--- a/include/decl.h
+++ b/include/decl.h
@@ -181,7 +181,6 @@ struct instance_globals_a {
     struct h2o_ctx acid_ctx;
 
     boolean havestate;
-    unsigned long magic; /* validate that structure layout is preserved */
 };
 
 struct instance_globals_b {
@@ -219,7 +218,6 @@ struct instance_globals_b {
     boolean bot_disabled;
 
     boolean havestate;
-    unsigned long magic; /* validate that structure layout is preserved */
 };
 
 struct instance_globals_c {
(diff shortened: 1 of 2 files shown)


nhmall:
I **think** the original intent by the author, back when there was only one very large struct, was as a prelude to the implementation of "_play again?_" functionality mentioned in the README.

The structs initialized by the compiler are not the ones that are actually used within the NetHack core code (instance_globals_f vs gf, for example).  If/when "_play again?_" were to get implemented, that field would have provided a way to check that the destination (gX) structs had been reinitialized with the pristine compiler-initialized values.

That is, if all the gX struct variations get cleared to zeros at the start of initiating "_play again?_", then the gX fields could be confirmed to having been properly reinitialized from the corresponding instance_globals_X struct by checking that the magic value had been reinstated in the destination gX struct.

To date, the "_play again?_" code has not yet been added to the NetHack-3.7 branch, and I suppose the potential implementor could always include the addition of the confirmation mechanism back in as part of the implementation, if they still wanted to.

On the other hand, is their presence actually harming anything beyond consuming the additional memory? I'm not opposed to removing them.

mkuoppal:
The memory impact is miniscule. I added this pull request with the sentiment of dead code is bad code. Feel free to close if there is no value added in here.

argrath:
> The memory impact is miniscule. I added this pull request with the sentiment of dead code is bad code. Feel free to close if there is no value added in here.

I agree.
This PR should be merged.

PR #1302, opened 2024-10-24
util: match fopen_datafile function prototype

mkuoppal:
The prototype for fopen_datafile() is different if it is used for util/dlb. Add the missing integer parameter and assert that it is only used for dataprefixes.

### DIFF
--- a/util/dlb_main.c
+++ b/util/dlb_main.c
@@ -18,7 +18,7 @@
 
 ATTRNORETURN static void xexit(int) NORETURN;
 ATTRNORETURN extern void panic(const char *, ...) NORETURN;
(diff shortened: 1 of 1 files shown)


nhmall:
Adding `#include "hack.h"`, as this pull request appears to do, would also require a dependency update for dlb_main.o in sys/unix/Makefile.utl (and other platform Makefiles and so forth, that have a build recipe involving util/dlb_main.c).

Also, it's unlikely that `#include "config.h"` is required once  `#include "hack.h"` has been added.

mkuoppal:
I omitted including "hack.h" as it was only for assert

PR #1311, opened 2024-11-04
early return on dosacrifice()

argrath:
... to reduce indents and prepare further refactorings.

### DIFF
--- a/src/pray.c
+++ b/src/pray.c
@@ -1827,7 +1827,9 @@ dosacrifice(void)
     struct obj *otmp;
     boolean highaltar;
     aligntyp altaralign = a_align(u.ux, u.uy);
-    int value = 0;
+    int value;
+    struct permonst *ptr;
+    struct monst *mtmp;
 
     if (!on_altar() || u.uswallow) {
         You("are not %s an altar.",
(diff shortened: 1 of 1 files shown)


nhmall:
Regarding this:
```
-    if (otmp->otyp == CORPSE) {
-        struct permonst *ptr = &mons[otmp->corpsenm];
-        struct monst *mtmp;
+    ptr = &mons[otmp->corpsenm];
```
Using _otmp->corpsenm_  as an index into mons[] without a guard can be problematic and/or draw complaints from static analyzers.

That's because the corpsenm field is overloaded for a number of uses:
```
../include/obj.h:#define leashmon corpsenm /* gets m_id of attached pet */
../include/obj.h:#define fromsink corpsenm /* a potion from a sink */
../include/obj.h:#define novelidx corpsenm /* 3.6 tribute - the index of the novel title */
../include/obj.h:#define migr_species corpsenm /* species to endow for MIGR_TO_SPECIES */
../include/obj.h:#define next_boulder corpsenm /* flag for xname() when pushing a pile of
```

Edit:
Your
```
+    if (otmp->otyp != CORPSE) {
+        pline1(nothing_happens);
+        return ECMD_TIME;
+    }
+
```
block at the top should be enough to appease an analyzer.  I haven't tested it yet.

nhmall:
I've tested it now, and there were no new analyzer complaints.
The pull request is in the pipeline.

PR #1331, opened 2024-11-30
Various fixes and additions for cross compilation to WebAssembly

guillaumebrunerie:
Hello,

I have been working on a browser/mobile port of NetHack 3.7 using cross-compilation to WebAssembly (it is very playable already, you can try it at https://guillaumebrunerie.github.io/nethack/).

Screenshot_20241130-183417

The existing code for compiling to WebAssembly was a great help, although it wasn’t fully up to date and was missing a number of things in order to be able to create a proper window port (for instance there was no way to set `iflags.window_inited` to true, `print_glyph` was not working properly as its signature changed, and various other things).

This pull request contains various fixes and additions that I found were needed/helpful.

Changes:
- export more constants/pointers/globals,
- fix various types that are incorrect,
- disable compression of save files, as it uses fork which isn’t supported in WebAssembly,
- include `genl_player_setup` when `SHIM_GRAPHICS` is defined, in order to make it possible to reuse the existing player selection code,
- move initialization of JavaScript global constants up, as I was running in some issue when `raw_print` was being called before initialization,
- change various compilation options for Emscripten, in particular it now generates an ES6 module (easier to use in a modern Javascript project) and exports more methods from Emscripten (for instance to be able to use the virtual file system when saving).
- simplify the implementation of the main loop to avoid `setTimeout` which can be pretty slow in the browser when called many times,
- change the way pointer arguments are being sent to JavaScript in `getArg` (they were being sent as a pointer to the pointer itself on the stack, which doesn’t really make sense, now the pointer itself is sent)

Ping @apowers313 who first implemented cross compilation to WebAssembly (#385), and @Anhijkt who I see recently fixed an issue that I ran into as well (#1328).

### DIFF
--- a/include/global.h
+++ b/include/global.h
@@ -482,6 +482,9 @@ extern struct nomakedefs_s nomakedefs;
 #if !defined(CROSS_TO_WASM) /* no popen in WASM */
 #define PANICTRACE_GDB
 #endif
+#ifdef CROSS_TO_WASM
+#undef COMPRESS
+#endif
 #endif
 
 /* Supply nethack_enter macro if not supplied by port */
--- a/src/role.c
+++ b/src/role.c
@@ -2175,7 +2175,7 @@ genl_player_selection(void)
     /*NOTREACHED*/
 }
 
-#if defined(TTY_GRAPHICS) || defined(CURSES_GRAPHICS)
+#if defined(TTY_GRAPHICS) || defined(CURSES_GRAPHICS) || defined(SHIM_GRAPHICS)
 /* ['#else' far below] */
 
 staticfn boolean reset_role_filtering(void);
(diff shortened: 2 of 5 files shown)


apowers313:
I'm super excited that you fixed this, thanks for putting the effort in.

nhmall:
This pull request is in the pipeline.

apowers313:
@guillaumebrunerie I know this was already merged but I was just thinking -- do we need to update any of the [API documentation](https://github.com/NetHack/NetHack/blob/NetHack-3.7/sys/libnh/README.md) with the changes that were made?

guillaumebrunerie:
> @guillaumebrunerie I know this was already merged but I was just thinking -- do we need to update any of the [API documentation](https://github.com/NetHack/NetHack/blob/NetHack-3.7/sys/libnh/README.md) with the changes that were made?

I don’t think so, I did not change the general way libnh and shim work.

PR #1353, opened 2025-01-01
Fix some bigroom wall corners to how they display in-game

copperwater:
Noticed a few corners in some bigroom maps were | instead of -, which doesn't have any gameplay effect but was mildly annoying for what I was doing at the time (copying the maps out into documentation that is supposed to show what the maps look like in-game).

There are other special levels out there that still use | for corners; this doesn't address those, only the bigrooms.


copperwater:
This was merged as 41d95d4.

PR #1355, opened 2025-01-01
Fix: brides of Dracula not generating in their niches

copperwater:
When the des.monster() statements for the vampire ladies were changed to use the lua-table form, the coordinate argument was not given the coord= name in the table, so the lua loader was ignoring it and the vampire ladies were placed on random spaces on the level. Fix this by supplying the coord=; testing shows that they now appear back in the niches.

Also lowercase the monster species id "Vampire Lady" to "vampire lady". The uppercase didn't affect the species being generated but having the id be the same case as in monsters.h is consistent with how it's done everywhere else.


copperwater:
This was merged as 171d48c.

PR #1366, opened 2025-01-15
use blood for "Vlad was here" engravings

vultur-cadens:
A hero in vampire form using a finger to engrave scrawls in blood; so should Vlad.

### DIFF
--- a/src/mklev.c
+++ b/src/mklev.c
@@ -673,9 +673,16 @@ makeniche(int trap_type)
                     if (trap_type != ROCKTRAP)
                         ttmp->once = 1;
                     if (trap_engravings[trap_type]) {
+                        int engr_type = DUST;
+                        /* "Vlad was here" might be in blood instead
+                           of dust, more likely at deep levels */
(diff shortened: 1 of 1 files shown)


copperwater:
Suggestion: the messages only have a chance of becoming scrawled in blood after a certain depth. They can appear very early, after all, and I think it increases the creepiness factor if an engraving you've seen before in dust, and thought little of it, is now scrawled in blood.

Vlad doesn't have to make blood engravings if he doesn't want to; writing with a soft item would do the trick of making a dust one.

vultur-cadens:
Updated.  The chance of "Vlad was here" being written in blood now increases with depth.  0% chance of a blood engraving at levels 6-9, increasing linearly from level 10 until it's 100% at level 21 and deeper.

NullCGT:
I really like that. It helps sell the idea that the dungeon is becoming more threatening the deeper you go, and that perhaps the powers that be are beginning to take you more seriously.

PR #1370, opened 2025-01-29
Remove useless call to `wakeup` in `bhitm` routine

mathewhodson:
`wakeup(mtmp, TRUE)` is called at the end since `wake` is set.

### DIFF
--- a/src/zap.c
+++ b/src/zap.c
@@ -406,7 +406,6 @@ bhitm(struct monst *mtmp, struct obj *otmp)
                     pline("%s doesn't budge.", Monnam(mtmp));
             }
             if (!DEADMONSTER(mtmp)) {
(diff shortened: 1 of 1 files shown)


nhmall:
Are you sure it's _useless_? Have you looked into all the side-effects around message sequencing?

I haven't looked in great detail, but a cursory look seems to show that:
wakeup() (the call that you are suggesting removing) calls wake_msg() which delivers a "Grumpy wakes up!" message.

The line immediately after the line you suggest removing is a call to abuse_dog(), and abuse_dog() delivers pet sounds, which should follow the "wakes up!" message:
```
        if (mtmp->mtame && rn2(mtmp->mtame))
            yelp(mtmp);
        else
            growl(mtmp); /* give them a moment's worry */
```

nhmall:
No response, closing.

PR #1372, opened 2025-02-06
Fix: ghoul and ettin zombie mis-ordered in monster list

copperwater:
An assumption of monster generating code is that monsters within a class appear in increasing order of difficulty. This wasn't the case in the Z class which went: human zombie (diff 5) -> ettin zombie (diff 7) -> ghoul (diff 5).

As a result, ghouls would not randomly generate when they should have at certain level difficulties (where a ghoul is weak enough to generate but an ettin zombie would be too strong).

Fix this by swapping the two monsters, which makes the difficulties in the Z class monotonically increase again.


paxed:
These monsters are in the wrong order: little dog, wolf,  minotaur,  ghoul,  elf,  guard,  Oracle,  soldier,  lieutenant,  watchman,  Croesus,  sandestin,  Death,  mail daemon,  djinni,  giant eel. This obviously doesn't matter for some of those (either NOGEN, or uniques).

One of the things that bothers me about this though: If we were to change a monster (eg. by making it do more damage, or whatever), adjusting the difficulty might also require reordering.

copperwater:
This pull request is obsoleted by 027bf78; withdrawing it. (Someone should maybe update the "Assumption 2: monsters of a given class are presented in increasing order of strength." comment though. Even the term "strength" is several years out of date.)

PR #1374, opened 2025-02-07
Livelog when a player breaks petless conduct

copperwater:
If a player initially goes petless then later obtains a pet, it's absent from the game chronicle. Fix that by adding a livelog for it.

This required a bit of restructuring in create_familiar() that I wanted to do anyway: removing the kludge of decrementing u.uconduct.pets when a figurine has been deployed but isn't actually going to come out tame. Calling initedog() /after/ deciding whether it's needed or not prevents a first-pet livelog being produced for a figurine that didn't come out tame.

The guardian angel code, which avoids calling initedog(), can never be the hero's first pet anyway because it only appears tame when the hero has already broken petless conduct. But while checking, I noticed a duplicate comment, so I removed that.

### DIFF
--- a/src/dog.c
+++ b/src/dog.c
@@ -63,6 +63,16 @@ initedog(struct monst *mtmp, boolean everything)
         EDOG(mtmp)->mhpmax_penalty = 0;
         EDOG(mtmp)->killed_by_u = 0;
     }
+    /* livelog first pet, but only if you didn't start with one (the starting
+     * pet will be initialized before in_moveloop is true) */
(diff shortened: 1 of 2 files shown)


paxed:
In "obtained %s first pet (a %s)" - shouldn't that "a" use an()?

copperwater:
You are correct - changed it to use an().

nhmall:
71e562def8ec365da6a99e28c702a4a9f1dba669

PR #1375, opened 2025-02-07
Clean up hardcoded material constant in water vault themerm

copperwater:
This implements a TODO to return an object's material as text rather than as an int when a Lua file requests all the details about an object's objclass. That is as simple as looking it up in materialnm[].

With that done, it's possible to clean up the one use where a Lua file looks up the material of an object it generated, in the "water-surrounded vault" themed room, previously an inflexible 19 but which can now be compared directly to "glass". It also enables shortening the comments that follow since the branches of the if statement are now obvious.


nhmall:
f228a790

PR #1376, opened 2025-02-07
Fix: gremlin cry of pain could be heard while deaf

copperwater:
Main problem was there was no condition applied to this message, so anyone would hear it even if they were deaf. Even assuming a cry of pain is something that could be seen, the message was still printed when the hero couldn't see the gremlin.

This puts both a deafness check and a range check on that cry (if a gremlin somehow takes light damage on the other side of the map behind many walls, it doesn't make much sense to hear its cry), and provides an alternate message if the hero can't hear it, but can see it. The alternate message does rely on the hero being able to /see/, not just spot, the gremlin and the light it's shying away from -- if you can only sense it, there is no special message.

### DIFF
--- a/src/uhitm.c
+++ b/src/uhitm.c
@@ -6264,8 +6264,15 @@ flash_hits_mon(
 void
 light_hits_gremlin(struct monst *mon, int dmg)
 {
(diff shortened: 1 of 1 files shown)


nhmall:
6c9d4df4

PR #1381, opened 2025-02-19
Fix: sitting on bidirectional teleportation traps

copperwater:
It is possible to create a bidirectional teleportation trap by making a pair of teleportation traps with a fixed destination of each other's coordinate. Moving or hurtling onto such a trap correctly materializes the hero on top of the other trap without triggering it, but for some reason I didn't dig into, sitting down to trigger the first trap does also trigger the second one at the destination end, causing you to counterintuitively teleport twice and end up back where you started.

Fix this by stopping tele_trap() from doing anything if it's called recursively, using a static variable like spoteffects() does for the same purpose. I had to adjust a bit of other tele_trap code to remove its sole early return.

### DIFF
--- a/src/teleport.c
+++ b/src/teleport.c
@@ -1458,6 +1458,14 @@ domagicportal(struct trap *ttmp)
 void
 tele_trap(struct trap *trap)
 {
+    /* a fixed-destination teleport trap could theoretically place hero onto a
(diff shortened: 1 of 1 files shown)


nhmall:
55c3a7c6c5b299d6e7528c0cb568d14d69c904e2

PR #1388, opened 2025-02-28
Fix: map random y-placement used its width instead of height 

copperwater:
Not sure how long this has existed without triggering any issues, but
when I was testing out a themed room wider than it was tall, I ran into
rn2-of-a-negative-number impossibles. Traced it to here, where it was
trying to subtract the width of the mapfrag from ROWNO to figure out
which y-value it should place the map on. The correct behavior is to
subtract the height of the mapfrag.

This pull request also contains a small unrelated lua.adoc update.

### DIFF
--- a/doc/lua.adoc
+++ b/doc/lua.adoc
@@ -1319,6 +1319,8 @@ Example:
 === room
 
 Create a selection of locations inside the (current) room.
+Does not include the edges of the room, such as its walls.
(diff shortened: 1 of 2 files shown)


nhmall:
50d80b5183a89e17cd0d68fce0bbeaaf3d6cb577 Fix: map random y-placement used its width instead of height

01cdf519935b01fc62b5191b67e3c49bc1365af4 Add a bit of detail in selection.room() documentation

PR #1405, opened 2025-04-01
Silver maces

disperse:
Added the silver mace to be the base weapon type of Demonbane.  It is appropriate that an artifact weapon designed to slay Demons would be made of (or plated with) silver.  This helps to offset the damage reduction when Demonbane was changed from a longsword to a mace and makes it more specialized against silver-haters.

Set the probability to 2, equal to that of a silver spear. Increased the weight of the silver mace by 120% -- equal to the weight increase from a normal spear to a silver spear. (Assuming the weapon is silver plated rather than made entirely out of silver.)

Increased the base cost to 60, a similar increase as spear to silver spear, to be an even number between silver spear and silver saber.
 
Monsters will prefer silver maces over regular maces.

Otherwise, identical in function to a normal mace.


paxed:
Missing the tile for the silver mace.

ars3niy:
Expert mace is 2 better against small than a basic long sword while 1 worse against large (better still if hit chance is still relevant), so there was only slight reduction against major demons (all but a couple of whom are large) and not a reduction overall. Silver damage would be a very comfortable buff.

PR #1426, opened 2025-06-20
Add gender preference (same or opposite) and check it in could_seduce()

greg-kennedy:
This PR adds two bits to monsters that track gender preferences - a 90% chance for "opposite" and a separate 10% for "same" - and these preference bits are checked in `could_seduce()` during a "seduction attack" against the defender's gender to determine whether the monster will seduce or not.

From the player's perspective, this simply means that 9/10 opposite gender foocubus will make a pass and 1/10 of the same will - same overall odds, just not 10/10 to 0/10.

**Gay incubus?**
Yes.  It makes any foocubus (or nymph) 81% hetero, 9% homo, 9% bi and 1% asexual.  Happy Pride!

**But I'm not into men?**
Grow up.

### DIFF
--- a/include/monst.h
+++ b/include/monst.h
@@ -160,8 +160,10 @@ struct monst {
     Bitfield(wormno, 5);    /* at most 31 worms on any level */
     Bitfield(mtemplit, 1);  /* temporarily seen; only valid during bhit() */
     Bitfield(meverseen, 1); /* mon has been seen at some point */
-
     Bitfield(mspotted, 1);  /* mon is currently seen by hero */
+
+    Bitfield(mprefopp, 1);  /* mon has preference for opposite sex */
+    Bitfield(mprefsame, 1); /* mon has preference for same sex */
 
     unsigned long mstrategy; /* for monsters with mflag3: current strategy */
 #ifdef NHSTDC
--- a/src/makemon.c
+++ b/src/makemon.c
@@ -1279,6 +1279,11 @@ makemon(
     else
         mtmp->female = femaleok ? rn2(2) : 0;
 
+    /* 90% chance of preferring opposite sex
+       and 10% chance of preferring same */
+    mtmp->mprefopp = (rn2(10) < 9);
+    mtmp->mprefsame = (rn2(10) < 1);
+
     if (In_sokoban(&u.uz) && !mindless(ptr)) { /* know about traps here */
         mon_learns_traps(mtmp, PIT);
         mon_learns_traps(mtmp, HOLE);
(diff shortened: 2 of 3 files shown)


copperwater:
The issue I see with this is that a monster's ability to seduce you is determined by the *player character's* sexuality, not the foocubus's. Currently, in-game, the hero is always heterosexual relative to their current sex, and is therefore vulnerable to seduction from the opposite sex while being immune to the same sex because that's not their orientation. I imagine that the foocubi themselves are always down for seduction and draining, and would gladly seduce a same-sex hero if the hero could ever be homosexual.

Several years ago, I and a few others came up with a solution that I implemented into xNetHack. I believe the impetus was to avoid player discomfort from having seductions that don't match them in real life (or otherwise: let players play as a sexuality that matches their own), and the solution was to make sexuality a setting in the config file.

The broad strokes: 

- There's an "orientation" option that accepts "straight", "gay", "heterosexual", "homosexual", or "bisexual". Configuring any of these in the file sets it for your character when you start a new game. It's not presented as a choice during character creation.
- Default orientation is heterosexual, not because of heteronormativity but so the game behaves the same to players who never set the option.
- Sexuality remains relative to the hero's current sex. I.e. a gay male hero who uses an amulet of change will now be seducable by succubi but not incubi.
- The foocubus summoned from a sink will always be compatible with your orientation (same as before).
- To compensate for bisexual characters being seducable by all foocubi, they roll twice for the chance of a severe headache.
- Asexuality doesn't exist mainly for gameplay reasons: it would allow a player who doesn't intend to take advantage of foocubi for levels or other benefits to turn off foocubi as a hazard. (For instance, in tournaments or for speedrunning this would probably be an optimal choice.) There's no change to the 'seduce' sysopt that removes seduction attacks from the install.

The commit, with some more design notes: https://github.com/copperwater/xNetHack/commit/e41489c

greg-kennedy:
> Several years ago, I and a few others came up with a solution that I implemented into xNetHack. I believe the impetus was to avoid player discomfort from having seductions that don't match them in real life (or otherwise: let players play as a sexuality that matches their own), and the solution was to make sexuality a setting in the config file.

I did consider this actually, but decided against it, because the idea of magical demonic seduction seemed (to me) *deliberately* not respective of the player's wishes - i.e. it's not about what the player wants but what the demon wants :)  certainly imagining myself (a straight guy) faced with an ancient male sex demon using magical hypnotism powers in a dungeon, let's be real, I'm not getting out of that with my clothes on if that's his preference.

So what this does is simply makes the *demons* not 100% heterosexual, as they were before.

But I'm not opposed to the xNethack solution either.

Issue #1429, opened 2025-07-24
Request: Deterministic Progressive Arti Wishing


chappg:
Posting after a discussion on #hardfought. 

3.7 introduces nerfs to artifact sacrificing and wands of wishing. Either nerf in isolation has limited effect on overall balance, but together they throw off the balance of obtaining the better artifact weapons on an ascension run.

My recent example: I received Sting, Ogresmasher, and Trollsbane as sacrifice gifts. I then used two wishes for Grayswandir, unsuccessfully. The lack of surplus wishes from the Castle means this is much more difficult to remediate and leads to frustration in 3.7. 

Demo's suggested resolution in the chat was to rework artifact wishing: "deterministic progressive arti wishing with reasonable blacklists plays so much better than vanilla behaviors".

copperwater:
This hardly works as a feature request because I have no clue what "deterministic progressive artifact wishing" even is. How is it supposed to work?

chappg:
Demo would be best to elaborate as he proposed this as the `obvious' solution; I believe a variant already has a similar mechanic that they were referring to.

In my ideation, I think a simple and satisfactory solution would be: artifact wishes are calculated using total artifacts minus the number of unsuccessful artifact wishes. Perhaps reset the count of unsuccessful artifact wishes to zero after any successful artifact wish.

copperwater:
I read it as working as intended - now that wishes are spread out, using both Castle wishes on a non-guaranteed chance of an artifact is a high-risk high-reward strategy. Getting frustrated when such a strategy fails is understandable, but I don't think the solution is to reduce the risk.

In the system you describe, it wouldn't have helped very much either - you had a 1/4 chance of missing on both wishes under current rules, and under these rules only the second wish would have different odds, for an overall 1/6 chance of missing on both wishes. It would only be guaranteed on the third wish. Ultimately I'm not a fan of steering the player towards getting the items in their late-game kit via wishes, so I don't agree with this being a good change.

chappg:
I don't think it steers the player towards late-game kit via wishes; the proposal still causes lost wishes and that's a huge penalty. This is for players who have already filled their kit except for a satisfactory artifact weapon and are willing to invest in multiple wishes to get one.

[...]

PR #1443, opened 2025-09-07
Make weaker artifacts less likely for better sacrifices

NDusN1:
It's weird that, with the current code, two sacrifice gifts meeting the same criteria has an equal chance of generating from a high-value sacrifice, which is noticeable from this short quote from #1429: 
> test your RNG with high-difficulty corpses that can still vend more early-game artifacts.

This PR makes it so that the chance of an artifact generating would be affected by its innate gift value versus max_giftvalue, the bigger the difference, the less the chance: a difference 10 or above would mean extremely unlikely.

### DIFF
--- a/src/artifact.c
+++ b/src/artifact.c
@@ -175,7 +175,7 @@ mk_artifact(
     boolean adjust_spe)  /* whether to add spe to situational artifacts */
 {
     const struct artifact *a;
-    int m, n, altn;
+    int m, n, altn, chance;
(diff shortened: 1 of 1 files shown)


NullCGT:
Seconded. This is something I've run into frequently in my recent 3.7 games, and it's very frustrating.

chappg:
I like the change and also have frustrations at always using early-game artifacts for most of my ascension runs.

PR #1456, opened 2025-11-05
Qt: Fix QTCXXFLAGS and WINLIB when compiling with WANT_WINT_QT6

horlogeislux:
The previous code made incorrect assumptions about Qt6's file hierarchy.
Reuse automatic configuration (with pkg-config) for Qt5, and replace '5' by '6'.

Works on macOS Tahoe 26.1 (25B78), when following instructions from `sys/unix/NewInstall.unx`.
This is especially important, as Qt6 is now the default for Homebrew (`brew`).

Also fix a typo in multiw-2.370.

### DIFF
--- a/sys/unix/hints/include/multiw-2.370
+++ b/sys/unix/hints/include/multiw-2.370
@@ -15,7 +15,7 @@
 #     - WINOBJ0
 #---
 # User selections could be specified as combinations of any of the following:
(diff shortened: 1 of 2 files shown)


nhmall:
4e29db079b6ae75609c9580bae14dc283ee9f0e6

Issue #1474, opened 2026-01-05
Confirm blessed scrolls of genocide when they could be typos


NateChoe1:
There's a potential YASD where a player accidentally genocides "master mind flayers" with a blessed scroll of genocide while playing as a dwarvish creature, as seen at 6:55:50 of [this](https://youtu.be/DzD9PSB4T7o) streak-ending run by Spazm. I wonder if this could be prevented by adding a confirmation whenever someone types in a specific monster name rather than a single letter for class genocides. Something like this:

```
What class of monsters do you want to genocide? master mind flayer
Are you sure you would like to eliminate all 'h' monsters? [yes|no] no
What class of monsters do you want to genocide? [enter '?' to see previous genocides] L
```

This probably wouldn't be necessary if the player types in a specific letter, as then it's clear that they knew that this was a blessed scroll.

This may also be intended behavior, similar to C343-120 "There is no warning when hitting floating eyes." being labeled as "Not a bug", although reading scrolls feels like a different type of action.

Tomsod:
3.7 does add a few new paranoid_confirmation options to protect from typos, and accidental self-genocide is a rather infamous way to die, although I'm more common with monks offing themselves via "master [anything]" being parsed as their rank title.  The latter is already ameliorated somewhat with the game accepting alternative spellings for master liches, so having a confirmation prompt with whatever the game thinks you've typed would fit the existing trend.

ars3niy:
This not a typo though. When you forget, or don't realise, that the scroll you're reading is blessed, that's distinct from fat-fingering or misspelling.

I suppose you could still have also an uncursed in open inventory on a nearby letter and mistype which one to read. One could also argue that, being blessed, the scroll should be extra forgiving (like it already works while confused). One could even argue that they already added confirmation for taking off gloves with a cockatrice corpse in inventory (non even wielded) and that's not (not primarily) against typos either.

Still, it's a tricky question. Where do you draw the line for when to protect the player against themselves?

PR #1475, opened 2026-01-06
Add paranoid_confirmation:genocide option

NateChoe1:
Addresses issue #1474, adds the `paranoid_confirmation:genocide` config option, which asks to confirm the chosen creature whenever you read a scroll of genocide. Here's a quick excerpt from running with a blessed scroll of genocide. I copy-pasted from `^P` so the messages are read bottom-up:

```
Wiped out all liches.  Wiped out all demiliches.  Wiped out all master liches.  Wiped out all arch-liches.
Are you sure you want to genocide all 'L' monsters? [yes|no] yes
What class of monsters do you want to genocide? [enter the symbol or name representing a class, or '?'] L
Are you sure you want to genocide all '@' monsters? [yes|no] no
What class of monsters do you want to genocide? master mnd flayer # note the typo
As you read the scroll, it disappears.
```

Here's another with an uncursed scroll of genocide:

```
Wiped out all master mind flayers.
Are you sure you want to genocide all master mind flayers? [yes|no] yes
What type of monster do you want to genocide? [enter the name of a type of monster, or '?'] master mind flayer
Are you sure you want to genocide all monks? [yes|no] no
What type of monster do you want to genocide? master mnd flayer
As you read the scroll, it disappears.
```

### DIFF
--- a/include/flag.h
+++ b/include/flag.h
@@ -90,6 +90,7 @@ struct flag {
 #define PARANOID_SWIM       0x0400
 #define PARANOID_TRAP       0x0800
 #define PARANOID_AUTOALL    0x1000
+#define PARANOID_GENOCIDE   0x2000
     unsigned versinfo; /* flag mask for 'showvers' option */
     /* mask bits for 'versinfo'; numeric order does not match display order
        which is "name branch number" */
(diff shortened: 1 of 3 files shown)


copperwater:
One interface consideration here is that tiles players are less likely to know the ASCII character representing a monster class, so a prompt like "Are you sure you want to genocide all 'h' monsters?" probably won't save a dwarven adventurer from genociding themselves even with the confirmation.

Personally, I don't think a paranoid option to handle it is the best solution - the way I'd do it is to remove the fuzzy matching and re-prompt until the player enters something unambiguous (or until hitting a "That's enough tries!" case, whatever). For instance, uncursed and cursed genocide should require the entire string to match a monster species, so "master mnd flayer" won't work. Blessed genocide should require the string to match either a monster class symbol or its string description, so entering a single species won't work.

PR #1482, opened 2026-02-07
get_rnd_text() equal odds for all lines

GorillaSapiens:
Renames rumors.tru to rumors_t.txt and rumors.fal to rumors_f.txt. Splits rumors into rumors_t and rumors_f.
Removes kludgey padding from engrave, epitaph, and bogusmon. Adds sidecar *.idx files for all of the above to guarantee all lines have the exact same odds of appearing, regardless of length. Removes kludgey get_rnd_line().
Adds new files to various Makefiles etc.


copperwater:
While I am in favor of making all lines from data files equal-probability, I think it could be achieved more cleanly and elegantly: 

- Create an instance_globals variable (or one each for true/false) of rumor offsets, type `int*`. This will be an in-memory array of the offsets, rather than an on-disk array in a separate .idx file.
- Create another instance_globals variable (or one each for true/false) for the number of rumors, type `int`.
- The first time in each game session a rumor is requested from the file, scan through the file to find the number of rumors, store it in the `int` variable, allocate an int array with that many entries to the `int*` variable, and scan through the file a second time to find the file offset of each rumor, storing them in the array as you go.
- On all subsequent requests for a rumor, randomly pick an offset and use the offsets array to seek to the appropriate location in the file.

GorillaSapiens:
> I think it could be achieved more cleanly and elegantly:

"clean" and "elegant", like "beauty", are in the eye of the beholder.

your proposed solution is effectively the same; build an index of file offsets.  where this PR stores that index on disk, your proposal stores it in memory.  which of course uses more memory by definition.

further, this PR builds the index once at compile time.  your proposal builds it every time the program is run, which is less efficient.

you also propose reading the file in 2 passes when an index can easily be built in a single pass. (hint: realloc() is your friend)

another possible solution would be "just read the whole thing into memory".  the obvious downside is "uses way more memory".

it's clear from the previous implementation that it was written at a time when memory was much more scarce than disk space.  this PR preserves that design goal by keeping the index on disk.  it's not clear to me how many of these legacy "low on memory" platforms are still supported, but i'm willing to bet there's at least one.

you are, of course, free to write your own implementation and submit it as a PR.

Issue #47, opened 2017-09-14
Make minotaur respect Elbereth?


ostrosablin:
Apparently, it was made non Elbereth-respecting back in '89, when 3.0.5 was released. It was done so to avoid polypiling Mino's wands into wishing. In 3.0.8, polypiling into wishing was disabled. So, there's no longer a reason to make Minotaurs an exception from Elbereth. Especially with nerfed Elbereth in 3.6.x. Perhaps, that change should be undone in vanilla, too?

https://unnethack.wordpress.com/2012/03/07/why-did-the-minotaur-cross-the-elbereth/

copperwater:
copying my comment from the pull request:

I figure that while the original motivation for the minotaur not respecting Elbereth may no longer apply, the game is more interesting when you have to gameplan for a monster that is physically strong, magically weak, and ignores Elbereth completely.

I believe this change would go against the general principle driving many monster changes in 3.7 by making minotaurs more boring instead of more different and interesting.

eakaye:
My thinking on this issue has evolved over the years. Back in the 3.4.3 days, I would have called Elbereth the most broken feature in the game -- a level one Tourist standing on Elbereth can solo a demon prince, assuming no ignoring nasties are summoned. So it makes sense to have at least a few enemies that force you to stand and fight, or employ ranged attacks or escape items. The Yendorian Army ceases to be a concern around the mid-game, and the angelic beings don't really have a presence until the ascension run, so minotaurs slot in nicely for this purpose.

However, minotaurs in practice generate mostly in Gehennom, where Elbereth doesn't function at all nowadays. The most likely scenario where you'd fight a minotaur in the Dungeons of Doom is if a shapeshifter polymorphs into one, or a random monster wanders onto a polytrap. This makes the minotaur an out-of-depth monster, and a hugely dangerous one at that -- faster than any player without speed boots, and often tough enough to survive even after emptying out all your attack wands. Oftentimes the player won't even have any escape items at that point in the game, meaning Elbereth may be their only line of defense. Additionally, Elbereth itself is weaker -- abusing it by attacking from it is devastating to your alignment record.

So my current thinking is, make minotaurs respect Elbereth. This makes no difference in the part of the game where you're expected to be able to handle them, and will prevent more than a few frustrating YASDs.

Tomsod:
[...]

Issue #100, opened 2018-05-19
LOGO - My contribution to netHack


foways:
<b>Hello!</b> I am a <i>graphic designer,</i> for the project "NetHack" I contribute a logo that I have designed is perfect, clean and minimalist.

![web](https://user-images.githubusercontent.com/38687140/40272077-820b8292-5bd1-11e8-8aa9-8aa77e88b509.png)

<b>The concept of logo design,</b> I picked up a very easy symbol to remember that is "n" which is a spelling rather than "nethack", and there is also a block code symbol. Here is the final result:

![last result](https://user-images.githubusercontent.com/38687140/40271847-84fbcc68-5bcd-11e8-9bf1-082e6bc3ac9b.png)

![background versions](https://user-images.githubusercontent.com/38687140/40271849-8a262c42-5bcd-11e8-8b0a-e454ecf4da7f.png)

That's some of my contribution to "netHack" hopefully you are interested and can use on your project. Give Feedback to me, and you can also give suggestions.

<b><i>k-syusteem</i></b>

koolraap:
It's not a bad logo. Much better than I could do.

It's not a NetHack logo, though. At the very least change the left-ear... little circle... thing.. (ear wax?) to an @

pavelbraginskiy:
Courtesy of Nethack.org, the current NH logo:
![image](https://user-images.githubusercontent.com/29113974/49701878-a703ff00-fba6-11e8-9a1f-b0b04afd6301.png)

AlexWithBeard:
Am I the only one seeing a guy drinking form the sink in the first logo?

neuronsong:
I'm pretty sure it's whatever character I'm playing being bent over the sink... or something... and the rng is just about to make me #RAGEQUIT... so from that point of view I suppose it's at least accurate in a certain sense.

PR #1484, opened 2026-02-07
minotaurs now respect Elbereth, fixes issue #47

GorillaSapiens:
fixes issue #47

### DIFF
--- a/src/monmove.c
+++ b/src/monmove.c
@@ -284,7 +284,7 @@ onscary(coordxy x, coordxy y, struct monst *mtmp)
      * Creatures who don't (or can't) fear a written Elbereth:
      * all the above plus shopkeepers (even if poly'd into non-human),
      * vault guards (also even if poly'd), blind or peaceful monsters,
-     * humans and elves, and minotaurs.
+     * humans and elves.
      *
      * If the player isn't actually on the square OR the player's image
      * isn't displaced to the square, no protection is being granted.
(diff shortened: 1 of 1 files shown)


copperwater:
I figure that while the original motivation for the minotaur not respecting Elbereth may no longer apply, the game is more interesting when you have to gameplan for a monster that is physically strong, magically weak, and ignores Elbereth completely.

I believe this change would go against the general principle driving many monster changes in 3.7 by making minotaurs more boring instead of more different and interesting.

GorillaSapiens:
> I believe this change would go against the general principle driving many monster changes in 3.7 by making minotaurs more boring instead of more different and interesting.

This is not the place for that discussion.

This PR fixes issue #47, which was opened Sep 14 2017.  If the devteam felt this was a bad change, they could have closed that issue any time in the last 8.5 years.

copperwater:
I felt it was better to comment on the pull request implementing the change because it's newer and it's less likely that people looking at this thread will go check on the issue to see if there are any dissenting opinions.

Unfortunately I do not think we can take an issue's long lifetime as a sign of general acceptance by the devteam - they do not appear to have any procedure in place to deal with suggestions that fail to achieve consensus or that none of them who looks at github is interested in merging. (case in point, the next oldest open issue #100 which adds something no one asked for or wants... still open after 7 years.)

GorillaSapiens:
> I felt it was better to comment on the pull request implementing the change because it's newer and it's less likely that people looking at this thread will go check on the issue to see if there are any dissenting opinions.

This PR should stay focused on reviewing the implementation and getting the code merged.

Broader discussion about the problem itself, scope changes, or design direction belongs in the linked issue. That keeps the conversation structured the way GitHub intends: issues track the “what” and “why,” PRs review the “how.”

Mixing the two makes the review harder to follow, buries actionable feedback in noise, and muddies the project history for anyone looking back later.

Please move the issue-level discussion to the issue so we can keep this PR clean and reviewable.

Thanks.

PR #1489, opened 2026-02-17
add WASM/Emscripten build support via shim window port

apowers313:
Backport of the WASM/Emscripten cross-compilation support from 3.7
(#385, #1331, #1328) to the 3.6 branch so the current production
release can run in the browser or Node.js.

Same approach as 3.7: a shim window port (win/shim/winshim.c) bridges
the C engine to JavaScript via an async callback, and a library main
(sys/libnh/libnhmain.c) replaces unixmain.c. A hints file
(sys/unix/hints/linux-wasm) wires up the Emscripten flags.

The one wrinkle specific to 3.6 is a cross-compilation data mismatch.
On LP64 hosts (Linux/macOS x86_64) unsigned long is 8 bytes, but in
WASM it's 4. The binary data files (dungeon, levels, quest text) are
written by native utilities and read by the WASM game, so the struct
layouts have to agree. A CROSS_TO_WASM guard in include/global.h
typedefs the relevant types to int/unsigned int when building for WASM
and leaves them as long/unsigned long otherwise. This doesn't come up
in 3.7 because the data file format was reworked there.

Build instructions and API docs are in sys/libnh/README.md.

I've also published this as npm packages for anyone who wants to embed
NetHack in a JS project without setting up the C toolchain:

  @neth4ck/wasm-367 -- https://www.npmjs.com/package/@neth4ck/wasm-367
  @neth4ck/neth4ck  -- https://www.npmjs.com/package/@neth4ck/neth4ck

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,5 @@ Makefile.gcc-orig
 # Win32-specific ignores end
 .DS_Store
 
+# WASM build artifacts
+wasm-data/
--- a/include/config.h
+++ b/include/config.h
@@ -50,6 +50,7 @@
 /* #define QT_GRAPHICS */    /* Qt interface */
 /* #define GNOME_GRAPHICS */ /* Gnome interface */
 /* #define MSWIN_GRAPHICS */ /* Windows NT, CE, Graphics */
+/* #define SHIM_GRAPHICS */  /* WASM/Library shim callback graphics */
 
 /*
  * Define the default window system.  This should be one that is compiled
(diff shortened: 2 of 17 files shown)


nhmall:
Interesting approach, that of using typedefs for related struct fields when building host-side utilities for the cross-compile that causes the same data sizes to be written as those that the target cross-compiler expects.

Beyond data type size matches, I would have been a little concerned about struct padding differences between the two compilers, as well as bitfield layouts between them (perhaps the involved structs for this avoid the use of any bitfields, I didn't actually take a look).

The embedding of the data for retrieval is also noteworthy.

I hope to look into this pull request more deeply at a later date.

EDIT: a couple of typo corrections

nhmall:

PR #1494, opened 2026-03-23
Amiga port adapted to 3.7 and cross compilation

ingpaschke:
This brings back the Amiga Intuition window port (AMII text mode and AMIV tile mode) for NetHack 3.7, updated for the current window_procs API and cross-compiled from Linux using bebbo's m68k-amigaos-gcc toolchain.

The Amiga code only needed minor modifications, the biggest changes are:

- cliparound after vision_recalc: Without this change, at least the Amiga port very noticeably redraws the map twice when the view changes which tests the patience of users of actual hardware. I think this fix applies to all other ports, but probably isn't much of an annoyance there.
- inclusion of Spencer regex: We don't have posix regex on stock AmigaOS installations or the toolchain (at least without resorting to ixemul which would have to be installed on the user's machine and be the right version), so I added the classic BSD licensed implementation for convenience in sys/amiga/regex. I don't know of that's acceptable to you.
-  minor additions to tile support (16 and 32 color tiles directly generated from the bmp file), added portable versions of the .iff utilites, enabled menucolor
-  I cleaned up the sys/amiga directory and left now unused files in outdated


nhmall:
> inclusion of Spencer regex: We don't have posix regex on stock AmigaOS installations or the toolchain (at least without resorting to ixemul which would have to be installed on the user's machine and be the right version), so I added the classic BSD licensed implementation for convenience in sys/amiga/regex. I don't know of that's acceptable to you.

Do you have a definitive location for the original of this? It might be best to have a "fetch" step to pull a copy into NetHack's lib subdirectory during, or ahead of, the build,  and adjust the Makefile build steps to compile it from that location, rather than distributing the 3rd party source in NetHack's source tree.

Rhialto:
As a maintainer of a packaging system (and having packaged NetHack for pkgsrc), I object to fetching stuff during build time. That sort of thing is severely frowned upon.
Fortunately there is unlikely to be an authoritative "upstream" location for this anyway.

ingpaschke:
On Monday, March 23rd, 2026 at 10:08 PM, nhmall ***@***.***> wrote:

> nhmall left a comment (NetHack/NetHack#1494)
> 

> > inclusion of Spencer regex: We don't have posix regex on stock AmigaOS installations or the toolchain (at least without resorting to ixemul which would have to be installed on the user's machine and be the right version), so I added the classic BSD licensed implementation for convenience in sys/amiga/regex. I don't know of that's acceptable to you.
> 

> Do you have a definitive location for the original of this? It might be best to have a "fetch" step to pull a copy into NetHack's lib subdirectory during build, and adjust the Makefile build steps to compile if from that location, rather than distributing the 3rd party source in NetHack's source tree.

Good idea, I took them from bebbo's amiga-gcc sourcetree (projects/ixmul). The most well known source today (this is from the 90s...) is probably https://github.com/garyhouston/regex

I updated the PR with this change if you want to implement it.

ingpaschke:
> As a maintainer of a packaging system (and having packaged NetHack for pkgsrc), I object to fetching stuff during build time. That sort of thing is severely frowned upon. Fortunately there is unlikely to be an authoritative "upstream" location for this anyway.

I agree that this complicated the build process, but we have precedent for this with fetch-lua. And this is an AmigaOS-only change so it won't affect pkgsrc or other packaging systems.

nhmall:
@Rhialto wrote:
> As a maintainer of a packaging system (and having packaged NetHack for pkgsrc), I object to fetching stuff during build time. That sort of thing is severely frowned upon.

Here's what I'm going to suggest:

- We can add a sys/amiga/fetch-amiga-cross-compiler.sh prerequisite step to assist those looking for help installing the cross-compiler itself ahead of the NetHack build (see the sys/msdos/fetch-cross-compiler.sh for a similar precedent). That script can check for, and pick up if necessary, the cross-compiler. It can also check for, and pick up if necessary, the regex source pieces and have them in place prior to commencing the actual NetHack build.

That makes it part of the prerequisite build-tools set-up process, rather than part of the NetHack build.

nhmall:
My concern about distributing 3rd party source code with a potentially conflicting license has been addressed in this latest iteration, so I no longer see any issue with merging this pull request.

Refinement of the best approach for obtaining the prerequisite pieces for an Amiga cross-compiler build may continue.

ingpaschke:
I would prefer to document the process of setting up the cross compile environment rather than automating it here. The bebbo cross compiler repository fetches, compiles and installs the Amiga NDK, binutils, gcc and g++, vbcc (another C-compiler), a C-Library replacement, a posix library, SDL. It would be a massive waste of disk space and compile time to install all of this just to compile nethack and a massive undertaking of slimming this down to just what's neccessary for compiling it.

I've also updated the fetch-commit.

nhmall:
a first build attempt:
https://gist.github.com/nhmall/add6c76d129f511c4c15b5f1046eb440

Observations:

- uudecode not found

1. Either util/uudecode needs to be built from sys/share/uudecode.c, and the Make recipe steps adjusted to use that NetHack-supplied util/uudecode.
OR
2.  the Amiga cross-compile documentation needs to state that uudecode needs to be installed and available on the system prior to building (perhaps from sharutils).

I suggest the former to minimize external prerequisite system changes. We supply it to assist with builds.

ingpaschke:
It now uses the supplied uudecode, thanks!

nhmall:
Cross-compile test results:
https://gist.github.com/nhmall/a6db6d5649d83d83a362bfc925a62c52

nhmall:
@ingpaschke Is the -DLUA_32BITS=1 necessary on the command line? It appears to be getting overridden by the #define in luaconf.h anyway.

```
/opt/amiga/bin/m68k-amigaos-gcc -c -O2 -noixemul -m68000 -include sys/types.h -I../include -I../sys/amiga -I../win/share -I../lib/lua-5.4.8/src -DAMIGA -DNOTTYGRAPHICS -DNO_TERMS -DNO_SIGNAL -DTILES_IN_GLYPHMAP  -DCROSSCOMPILE -DCROSSCOMPILE_TARGET -DCROSS_TO_AMIGA -DAMIGA_VERSION_STRING=\""VER: NetHack 3.7.0"\" -DLUA_32BITS=1 -c  -o../targets/amiga/lapi.o ../lib/lua-5.4.8/src/lapi.c
In file included from ../lib/lua-5.4.8/src/lua.h:16:0,
                 from ../lib/lua-5.4.8/src/lapi.c:17:
../lib/lua-5.4.8/src/luaconf.h:125:0: warning: "LUA_32BITS" redefined
 #define LUA_32BITS 0
```

ingpaschke:
it's not needed, thanks.

PR #1497, opened 2026-04-06
Fix brew CFLAGS for ncurses in macOS.370 hint.

horlogeislux:
This fixes a mismatch between macOS system ncurses headers and the brew ncurses library, which do not export the same symbols.


bhaak:
Thanks. Fixed with 60de227 already.

PR #1500, opened 2026-04-13
Fix Clang detection in unix hints

Feyorsh:
When trying to detect if the compiler is Clang or not, the users of `$(CCISCLANG)` expect it to be the exit status of the `grep` command, when in fact it is just the result of the grep command (an empty string if no match and the version information if it did match). 
```
CCISCLANG := $(shell echo `$(CC) --version` | grep clang)
```


nhmall:
Thank you. I'm going to adjust things in a slightly different way than this pull request.

PR #1510, opened 2026-05-03
Amiga 5.0.0 startup fixes

ingpaschke:
When I tested the release version on an Amiga 1000, I noticed some problems in the startup code:

- the stack setup code wasn't properly setup for libnix, so it relied on the user having a rather large stack-setting in their environment.
- we don't need SYSCF on Amiga and the code caused the startup to fail with "sysconfig not found" when the sysconfig file wasn't in the current directory
- the shipped nethack.cnf wants to use tile mode by default. That failed with an Oops (bad openwin) if the current screenmode doesn't allow for at least 16 colors or not enough chipmem is available. It now prints a warning and falls back to text mode.

The release binary works fine if you make sure that the window-type is set up correctly for your machine in nethack.cnf, you have enough stack space ("stack 262144") and start from the nethack directory:

```
> assign nethack: Games:NH500AMI
> stack 262144
> cd nethack:
> nethack
```

### DIFF
--- a/include/amiconf.h
+++ b/include/amiconf.h
@@ -212,4 +212,7 @@ struct ami_sysflags {
 };
 extern struct ami_sysflags sysflags;
 
+#undef SYSCF
+#undef SYSCF_FILE
+
 #endif /* AMICONF_H */
--- a/sys/amiga/amistack.c
+++ b/sys/amiga/amistack.c
@@ -22,6 +22,10 @@
 #ifdef __SASC_60
 long __stack = 256 * 1024;
 #else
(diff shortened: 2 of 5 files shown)


nhmall:
Acknowledging this pull request. 
It is now in the pipeline as it currently stands. At this point it is best to open a new pull request for further changes, or it will get complicated (for me). :)

nhmall:
An updated Amiga binary is now available
https://www.nethack.org/v500/ports/download-amiga.html

Edit: typo correction

ingpaschke:
Great! Confirmed working on real hardware.

PR #1515, opened 2026-05-04
Amiga: Workbench integration and quality of life improvements

ingpaschke:
The headline change is that Nethack can now be started directly from the workbench without any assigns and has a proper icon.

### Workbench integration

- automatically assign nethack: so the user doesn't have to fiddle with assigns.
- include .icon files for nethack and the user facing text files
- restructure the distfile so we have a NetHack-Folder that's visible in Workbench
- update README.amiga and include it in the distribution

### Cleanup / cosmetic fixes

- support the altmeta-flag again
- version info didn't display the supported windowing systems
- number_pad:1 in nethack.cnf as default is probably more appropriate for most Amiga configurations (except the 600)
- black and white rendered as inverse video which looked bad in amii mode
- the warning when falling back to text mode was backwards
- provide an implementation of get_nhuuid / free_nhuuid for the Amiga port


nhmall:
In the pipeline

Issue #1512, opened 2026-05-04
nethack --showpaths prompts for input if hackdir is unreadable/unwritable


mistydemeo:
In NetHack 5.0.0, if the hackdir can't be read/written by the nethack program, then `nethack --showpaths` will print a warning and also prompt the user to hit enter before continuing. This happens even if the command is being run noninteractively.

```
Warning: cannot write scoreboard file '/opt/homebrew/share/nethack/record'

Hit return to continue:
```

This came up when I was packaging nethack 5.0.0 for Homebrew (https://github.com/Homebrew/homebrew-core/pull/280634). Homebrew's tests run `nethack --showpaths` and reads the output to determine if the hackdir was set correctly, so there's no user present to hit enter and continue to print the output. Homebrew also runs its tests in a sandbox which prevents certain paths from being read/written to by the program, and it happens that the hackdir is one of them. (That'd be a problem for actually playing the game, but since the tests are just checking that the paths were configured all right it's not an issue!)

I bisected the issue to 57f86662fd9d2157e288d544e078d78648dc8818; this doesn't occur before that commit.

mistydemeo:
Thanks! I tried applying 60d59f4d5574c00cbb391cd58da3ed959de1ba2b to stable, and building the latest git HEAD, but I'm still seeing this issue either way. If you're looking to repro and you've got a Mac, you can use this sandbox profile to deny access to the record dir: https://gist.github.com/mistydemeo/2f7e2aa6534fa45e7bed2510a0a7b1b2 (Replace `$USER` with your actual username.)

And run it via: 

```sh
/usr/bin/sandbox-exec -f $PATH_TO_PROFILE/homebrew.sb $PATH_TO_NETHACK/nethack -- --showpaths
```

mistydemeo:
Right, I see what happened - I tested it with stdin being a tty but stdout not directed to my terminal, and in that case it still prompted for input but I couldn't see it because output was being captured by something else. I can confirm that it doesn't block if stdin isn't a tty.

mistydemeo:
Confirmed the second patch in b7735632bfdac6a502f8f2954fbe6d5bbac53e2b has fully fixed this. Thanks!

PR #1516, opened 2026-05-05
wintty: return early if stdout not a tty

mistydemeo:
This was already done if stdin wasn't a tty. If stdout isn't either, the user won't be able to see the message asking them to input and so it makes sense to skip prompting them in that circumstance as well.

This fixes Homebrew's automated tests for NetHack, which calls `nethack --showpaths` to check the configured paths. In Homebrew's case, stdout isn't a tty and stdin is, but it's unattended so nothing will ever be available to hit enter.

This adds on to 60d59f4d5574c00cbb391cd58da3ed959de1ba2b, which added the check for stdin. refs #1512

### DIFF
--- a/win/tty/wintty.c
+++ b/win/tty/wintty.c
@@ -766,7 +766,7 @@ getret(void)
 #if defined(MICRO) || defined(WIN32CON)
     getreturn("to continue");
 #else
-    if (!isatty(STDIN_FILENO))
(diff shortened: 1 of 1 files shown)


mistydemeo:
Closing since b7735632bfdac6a502f8f2954fbe6d5bbac53e2b handled this.

PR #1519, opened 2026-05-06
Unoffical NetBSD 10.1 configuration and compile fixes.

SirWumpus:
* Add a `hints/netbsd.500` for non-pkg install under `/usr/local`.  

* From `man tparm`:
```
       char *tparm(const char *str, ...);
            /* or */
       char *tparm(const char *str, long p1 ... long p9);
```
Added macro `tparm2` to handle the compiler error for second use case (`gcc 14.3.0`).

* NetBSD `termcap` exports `UP`, `BC`, and `PC` capabilities and from the man page:
> Only PC is actually used internally.


nhmall:
I created a NetBSD virtual machine and gave it a shot.
Results are here:
https://gist.github.com/nhmall/a92c105c0d177ed2b3cee55793281c0e

For the mostpart it seemed to work.

- There is an error on the Guidebook build. It isn't clear to me, at first glance, what is causing that.
```
( cd doc ; make Guidebook )
cat Guidebook.mn | ../util/makedefs --grep --input - --output - | tbl tmac.n - | nroff -c -Tascii  | col -bx > Guidebook
tmac.n:755: a tab character is not allowed before an argument
fatal error: cannot locate "tmac.nh" macro package
User Abort.
```
- There isn't support for the less-verbose and repetative compiler command lines by using a response file via resp=1 on the Make command line, which was added to the other .500 series of hints and hints/includes.

SirWumpus:
The focus of my PR is the binary building, not documentation.  I am not familiar with troff et. al in order to fix documentation building.

- - -
From your output...
```
( cd doc ; make Guidebook )
cat Guidebook.mn | ../util/makedefs --grep --input - --output - | tbl tmac.n - | nroff -c -Tascii  | col -bx > Guidebook
tmac.n:755: a tab character is not allowed before an argument
fatal error: cannot locate "tmac.nh" macro package
User Abort.
```

Note how the `( cd doc ; ... )` yet within three files there is reference to `doc/tmac.nh` ...

```
elf$ grep 'doc/tmac.nh' *
Guidebook.mn:.if !\n(nH So doc/tmac.nh
mnh.7:doc/tmac.nh
mnh.txt:       doc/tmac.nh
```

This doc issue with build process and directory position, not tools from what I can.  I did try editing the two files to remove the `doc/` and rebuild the Guidebook and it made no difference, same error occurs.

nhmall:
It doesn’t behave the same way on Linux. There is one expected warning there, but it completes. One of the tools must be behaving differently on NetBSD. There is likely an alternative recipe that works on NetBSD that the hints file can utilize - the trick is identifying it. I’m no *roff expert either. 

I wasn’t pointing out the result to gripe or reject the PR though. Let’s give it a little time to see if a fix arises from the “community eyes.”  That’s one of the advantages of public exposure on GitHub.

SirWumpus:
> I wasn’t pointing out the result to gripe or reject the PR though

Phew.

- - -
One thing of note (or not) is NetBSD (BSD)  `make(1)` is different from GNU Make where conditional syntax is concerned.  I have not seen any build issue concerning `make(1)` yet, but did notice `linux.500` uses some GNU syntax.  In which case on NetBSD if you really need GNU Make syntax:

```
# pkgin install gmake
```

Then use `gmake(1)` instead of `make`.

- - -
Also the current `groff`, as you noted, is 1.19.2 while there there exists a slightly newer package:

```
$ pkgin avail | grep roff
...
groff-1.23.0nb1      GNU roff text processing suite
...
$
```

nhmall:
I am aware of the bsdmake and GNU make differences. We deliberately avoid the use of most GNU syntax in the actual Makefiles for that reason. Only the hints files introduce GNU syntax. 

We do use “?=“ variable assignment in the actual Makefiles, which assign the value only if a value has not already been assigned. That’s very handy. Fortunately bsdmake and GNU make both support that.

SirWumpus:
> We do use “?=“ variable assignment in the actual Makefiles, which assign the value only if a value has not already been assigned. That’s very handy. Fortunately bsdmake and GNU make both support that.

Also `!=` syntax for shell expression supported by both.  I use this sometimes to work around the difference, though a little clunky.

From a personal project:
```
CC	!= if test ${CC} = 'c99'; then echo gcc; else echo ${CC}; fi
LDFLAGS	!= if test ${CC} = 'gcc'; then echo '-fno-ident -flto'; fi
```

nhmall:
It definitely seems to be GNU nroff (groff) version 1.19.2 that is causing the issue.
```
nhmallbsd$ cat Guidebook.mn >../p1
nhmallbsd$ cat ../p1 | ../util/makedefs --grep --input - --output ->../p2
nhmallbsd$ cat ../p2 | tbl tmac.n - >../p3
nhmallbsd$ cat ../p3 | nroff -c -Tascii >../p4
tmac.n:755: a tab character is not allowed before an argument
fatal error: cannot locate "tmac.nh" macro package
User Abort.
```

I've marked line 755 below with '->':
```
    .\" install traps
->  .wh 0i hd                               \" position header trap
    .wh -\n(bmu fo                          \" position footer trap
    .wh \n(.pu+1i fx                        \" put footnote overflow trap here
```

I'll try installing the newer groff next. 
The log of the groff installation can be found here:
https://gist.github.com/nhmall/fb4786d59a785011654330135c24b204

```
nhmallbsd$ groff --version
GNU groff version 1.19.2
Copyright (C) 2004 Free Software Foundation, Inc.
GNU groff comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of groff and its subprograms
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.

called subprograms:

GNU troff (groff) version 1.19.2
GNU grops (groff) version 1.19.2
```
That is the same version as before.
[searches the web...]
Okay, apparently the newer version gets placed in /usr/pkg/bin and isn't invoked by default.
```
nhmallbsd$ /usr/pkg/bin/groff --version
GNU groff version 1.23.0
Copyright (C) 2022 Free Software Foundation, Inc.
GNU groff comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of groff and its subprograms
under the terms of the GNU General Public License.
For more information about these matters, see the file
named COPYING.

called subprograms:

GNU grops (groff) version 1.23.0
GNU troff (groff) version 1.23.0
```

Try that last command again, specifying /usr/pkg/bin/groff this time:
```
nhmallbsd$ cat ../p3 | /usr/pkg/bin/groff -c -Tascii >../p4
gtroff:tmac.n:762: error: cannot load font 'S' for emboldening
```
That error is the same as the one received on Linux, so that looks promising.
Trying the Makefile recipe again:
```
cat Guidebook.mn | ../util/makedefs --grep --input - --output - | tbl tmac.n - | /usr/pkg/bin/groff -c -Tascii | col -bx > Guidebook

```
That's really close. We just need to add the suppression of the color and other escape sequences in the output. The _-c_ argument to groff was supposed to do that, but I have a feeling the wrong grotty is being invoked (not the new one in /usr/pkg/bin/)

This seemed to fix the output when run prior to the command though:
```
export GROFF_NO_SGR=1
cat Guidebook.mn | ../util/makedefs --grep --input - --output - | tbl tmac.n - | /usr/pkg/bin/groff -c -Tascii | col -bx > Guidebook
```

nhmall:
I found another note about the newer groff installation on NetBSD:
`Path: The newer groff will be installed in /usr/pkg/bin/groff. Ensure /usr/pkg/bin is in your $PATH before /usr/bin`

```
echo $PATH
/home/nhmall/bin:/bin:/sbin:/usr/bin:/usr/sbin:/usr/X11R7/bin:/usr/pkg/bin:/usr/pkg/sbin:/usr/games:/usr/local/bin:/usr/local/sbin
```

Okay, I'll update the path and try again.

That didn't help. I had to use this to get the desired output with no escape sequences.
```
export GROFF_NO_SGR=1
cat Guidebook.mn | ../util/makedefs --grep --input - --output - | tbl tmac.n - | /usr/pkg/bin/groff -c -Tascii | col -bx > Guidebook
```

nhmall:
Once commit d62b56e0f2b9bb86d96759528b0df580e78459a9 arrives for sys/unix/Makefile.doc later tonight,
your hints file will work for the documentation by appending this to the bottom of your sys/unix/hints/netbsd.500 file, as long as the newer groff is installed on your netbsd system. If it isn't the results will be the same as they have been:

```
.ifdef MAKEFILE_DOC
MAKEDEFS = ../util/makedefs
NEEDMAKEDEFS = $(MAKEDEFS)
.if exists (/usr/pkg/bin/groff)
NROFF = /usr/pkg/bin/groff
.else
NROFF = groff
.endif
# NROFF_FLAGS comes from a hints file
GUIDEBOOK = Guidebook           # regular ASCII file
COLCMD = col -bx
PSCMD = $(NROFF)
NHGREP = $(MAKEDEFS) --grep --input - --output -
GUIDEBOOK_MN_SRC = Guidebook.mn
GUIDEBOOK_MN = $(GUIDEBOOK_MN_SRC)
GUIDE_PREFORMAT = cat $(GUIDEBOOK_MN) | $(NHGREP) | tbl tmac.n -
ONEPAGE_PREFORMAT = cat Gbk-1pg-pfx.mn $(GUIDEBOOK_MN) Gbk-1pg-sfx.mn \
        | $(NHGREP) | tbl tmac.n -
ONEPAGECMD = $(ONEPAGE_PREFORMAT) | $(NROFF) -c -Tascii $(NROFF_FLAGS) \
        | $(COLCMD)
GUIDECMD = export GROFF_NO_SGR=1; \
        $(GUIDE_PREFORMAT) | $(NROFF) -c -Tascii $(NROFF_FLAGS) \
        | $(COLCMD)
GUIDEBOOK_TEX_SRC = Guidebook.tex
GUIDEBOOK_TEX = $(GUIDEBOOK_TEX_SRC)
.endif  # MAKEFILE_DOC
```

EDIT: replaced nroff with $(NROFF) in the assignment of PSCMD

SirWumpus:
```
PSCMD = groff
```
Shouldn't PSCMD be `PSCMD = $(NROFF)` ?

nhmall:
Yes, correct - it should be.

nhmall:
This pull request contains a contributed file sys/unix/hints/netbsd.500. 
Your name should be on the Copyright notice, not Kenneth Lorber's. Please correct that to get it merged.

bhaak:
@SirWumpus disregard my previous comment. You don't need to do anything. We'll sort it out.

SirWumpus:
> @SirWumpus disregard my previous comment. You don't need to do anything. We'll sort it out.

I can nuke and start again.  Its only 3 files (that I've saved just in case).  I tried to sort it out yesterday when I rebased to pull in Makefile.doc changes, which then included all the stuff in between my two commits.

SirWumpus:
Shows as merged now.  Do I bother?

nhmall:
Don’t bother right now. We are sorting out some things.

nhmall:
@SirWumpus, I think the changes from your pull request have all been included now, although the "merging" of the changes took an unexpected approach.

Something was off about your pull request or your repo that the pull request uses, so you might want to look into that, or scrap it and start it over. I'm not sure what happened there, but perhaps I might jokingly suggest "don't do that in future." :)

Please feel free to open new pull requests if you want to do any maintenance on the NetBSD support going forward.

Thanks for contributing to NetHack.

SirWumpus:
All I did was submit a PR, then a pile of other changes came about, followed by [a request I change the silly copyright](https://github.com/NetHack/NetHack/pull/1519#issuecomment-4432830815) and I think this is where it went wrong and also add some [extra stuff for handling documentation](https://github.com/NetHack/NetHack/pull/1519#issuecomment-4426471236) building which required a change that I had to rebase (with many other commits) into my branch.  It didn't look right to me either.

I think the OPR should have been accepted as it was, THEN a separate PR submitted with the requested additions.  That would have avoided the whole mess for both of use.

Sorry for the mess.

nhmall:
> I think the OPR should have been accepted as it was

It couldn't be accepted as it initially was. While it tested fine
on NetBSD, the original pull request broke the build for Linux,
and possibly other platforms. 

The changing of win/tty/termcap.c to use
tparm2() meant that tparm2() _had to be defined_ for all platforms, and
it wasn't.  The only #define for tparm2() was inside an 
`#if defined(BSD) || defined(ULTRIX)`
block, so the compiler complained on other platforms.

This was the part of the original pull request that I'm referring to: 
```
#if defined(BSD) || defined(ULTRIX)
#if !defined(DGUX) && !defined(SUNOS4)
#if defined(__NetBSD__)
#define tparm2(s, x)	tparm(s,x,0,0,0,0,0,0,0,0);
#else
#define tparm2(s, x)	tparm(s,x);
#endif
#if !defined(DGUX) && !defined(SUNOS4) && !defined(__NetBSD__)
#define memcpy(d, s, n) bcopy(s, d, n)
#define memcmp(s1, s2, n) bcmp(s2, s1, n)
#endif
```

SirWumpus:
Ah.  You didn't mention that before.

nhmall:
It was mentioned in the commit message for afeb2f5, where that issue was fixed.

PR #1524, opened 2026-05-09
Amiga: restore glyph icons in menus

ingpaschke:
Core changes routed transient inventory away from WIN_INVEN, so the port's win == WIN_INVEN gate stopped matching and glyphs vanished. Track has_glyphs on the menu and key rendering and sizing off that.


nhmall:
c1ae9d5 is in the pipeline.

PR #1529, opened 2026-05-12
Amiga: Cleanup followup

ingpaschke:
## More fixing of long-standing bugs in the Amiga UI codebase.

- playing with overview window open could crash the game when you pressed ESC anywhere
- tombstone rendering was broken on RTG screens and the lowres assets would look bad anyway, so falling back to text mode tombstone for now
- When printing messages without spaces, the --more-- logic could get stuck (triggered by very long names for example)
- picking from the extended command menu with the mouse only returned the first command starting with that letter
- many fixes found by automated static analysis ranging from pedantic to valid concerns

### DIFF
--- a/include/amiconf.h
+++ b/include/amiconf.h
@@ -48,9 +48,8 @@ extern void nethack_exit(int);
 extern void amii_setpens(int);
 
 extern void getlind(const char *, char *, const char *);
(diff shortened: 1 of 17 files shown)


nhmall:
In the pipeline, as it is.

PR #1543, opened 2026-05-24
lua: read coordxy fields with the right width (big-endian fix)

ingpaschke:
Add ANY_INT16 to any_types and use it for u.ux/uy/tx/ty and uz dlevel/dnum.  These are coordxy (int16_t) but the Lua bindings were treating them as 1-byte fields, so on big-endian m68k Lua read the high byte (0) instead of the actual value.  Symptom: place_object off map <0,0> in the tutorial.


nhmall:
This is in the pipeline.

PR #1547, opened 2026-05-25
display: fix off-by-one in glyph_is_normal_piletop_obj

ingpaschke:
Change > to >= so the two ranges are inclusive on the same side. --dumpglyphnames now produces (8009) G_piletop_generic_venom.


nhmall:
87a0bdf26c851f5da2a48279c0d647e25dde5e2b

PR #1561, opened 2026-05-30
Add support for 256 colours in curses window port

ars3niy:
Only glyphs that are customised via config can have colours different
from the basic 16. For such glyphs, when curses supports 256 colours,
use them.

### DIFF
--- a/include/wincurs.h
+++ b/include/wincurs.h
@@ -145,7 +145,6 @@ extern void curses_puts(winid wid, int attr, const char *text);
 extern void curses_clear_nhwin(winid wid);
 extern void curses_alert_win_border(winid wid, boolean onoff);
 extern void curses_alert_main_borders(boolean onoff);
-extern int get_framecolor(int nhcolor, int framecolor);
 extern void curses_draw_map(int sx, int sy, int ex, int ey);
 extern boolean curses_map_borders(int *sx, int *sy, int *ex, int *ey,
                                   int ux, int uy);
--- a/include/wincurs.h
+++ b/include/wincurs.h
@@ -128,14 +128,6 @@ extern void curses_refresh_nethack_windows(void);
 extern void curses_del_nhwin(winid wid);
 extern void curses_del_wid(winid wid);
 extern void curs_destroy_all_wins(void);
-#ifdef ENHANCED_SYMBOLS
-extern void curses_putch(winid wid, int x, int y, int ch,
-                         struct unicode_representation *u, int color,
-                         int framecolor, int attrs);
-#else
-extern void curses_putch(winid wid, int x, int y, int ch, int color,
-                         int framecolor, int attrs);
-#endif
 extern void curses_get_window_size(winid wid, int *height, int *width);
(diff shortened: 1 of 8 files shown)


ars3niy:
To show how it works for me (and what existing features I tested so they do not break), here is a screenshot containing:
- An erinys with custom extended colour (OPTIONS=glyph:G_female_erinys:&/light-coral)
- A barbed devil with default colour
- A fox with custom basic colour (OPTIONS=glyph:G_male_fox:d/blue)
- A highlighted pet
- A highlighted item pile
nhcolour

nhmall:
Have the changes in win/curses been tested on pdcurses as well as ncurses?

ars3niy:
ncurses (version 6.5 on Linux) is what I tested it on. pdcurses I have not tried but it would just be a matter of checking for regressions, as these commits in their current form cannot enable 256-colour support when only 256 colour pairs are available (which seems to be the case with pdcurses according to its documentation). 

It would theoretically be possible to work even with 256 colour pairs because right now there are 16 character colours and 8 background colours, leaving 128 colour pairs to be used for a subset of extended colours with the help of some kind of colour pair allocator. I don't know if that would be worth the complexity, however.

ars3niy:
The way 256 colours works in this merge request at the moment is, during initialisation 2048 colour pairs will be created for each of 256 colours and 8 background colours, without any regard for which combinations would be in use.

ars3niy:
This is with pdcurses (SDL2 of Linux). The library indeed reports 256 available colour pairs so the game sticks to 16 colours.

nhcolour_pdcurses

nhmall:
The plan is to review this pull request over the next two weeks (June 14-28). It requires testing on a few different platforms.

nhmall:
During an initial test, this function:
```
boolean
curses_has_256color(void)
{
#ifdef NCURSES_EXT_COLORS
    return (COLORS >= 256) && (COLOR_PAIRS >= 256 * CURSES_NUM_BACKGROUND_COLORS);
#else
    return FALSE;
#endif
}
```
is capping my [pdcursesmod](https://github.com/Bill-Gray/PDCursesMod) maxc at 16. The function returns FALSE, and the next line to execute:
`int maxc = curses_has_256color() ? 256 : (COLORS >= 16) ? 16 : 8;`
sets maxc to 16, but pdcursesmod COLORS holds a value of 768 (wincon version).

ars3niy:
Looking into it again, I now think checking NCURSES_EXT_COLORS was a mistake even back when I thought ncurses was the only curses. Because ncurses compiled without that option will report 256 max colour pairs anyway, which is already checked.

So I removed the macro check, curses_has_256color now returns true with pdcursesmod and 256 colours seem to work for me (only tried SDL2 version, there it reported 2²⁴+256 colours and 2²⁰ colours pairs). Perhaps you can test again with your setups?

bhaak:
That works now for me with pdcursesmod.

Minetown and Fort Ludios have now colored walls (they have RGB values which are mapped to 256 color indexes).

PR #1562, opened 2026-05-31
libnh: export GLYPH_PILETOP_OFF constant

avgas3:
## Summary
Expose GLYPH_PILETOP_OFF in libnh JS constants so embedding apps can distinguish statue glyphs from piletop glyphs.

## Change
- Add GLYPH_PILETOP_OFF to js_constants_init() in sys/libnh/libnhmain.c

## Rationale
This allows downstream libnh consumers to separate statue glyphs from piletop glyphs using public constants.


nhmall:
In the pipeline for later today

PR #1568, opened 2026-06-05
mac: drop the obsolete exit getreturn

ingpaschke:
The revived mac68k port (on the m68k-wip branch) removed it: it runs an event loop after exit_nhwindows() has torn the windowing system down and crashes; the tombstone already pauses.  Removing it here too keeps the block from resurfacing in NetHack-5.0 <-> m68k-wip merges.


nhmall:
This is in the pipeline.

PR #1570, opened 2026-06-06
Don't overflow svp.plname on restore

chasonr:
`get_plname_from_file` requires a buffer of size `PL_NSIZ_PLUS`.  `dorecover()` passes `svp.plname` directly to `get_plname_from_file`.  `svp.plname` has size `PL_NSIZ`. The resulting buffer overflow clobbers `svp.pl_character`.

The bug manifests in the MS-DOS port. Configure for a tiled display, either VGA or VESA. Start a game. The F4 and F5 keys work. Save, and then restore. The F4 and F5 keys stop working, because they check `svp.pl_character` to avoid corrupting the screen in the initial dialog.

The change passes a buffer of the proper size and copies it to `svp.plname`.

A check is added for save files that were affected by the bug so that `svp.pl_character` can be filled.

### DIFF
--- a/src/restore.c
+++ b/src/restore.c
@@ -713,6 +713,14 @@ restgamestate(NHFILE *nhfp)
     restore_oracles(nhfp);
     Sfi_char(nhfp, svp.pl_character,
              "gamestate-pl_character", sizeof svp.pl_character);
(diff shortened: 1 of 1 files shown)


nhmall:
This is in the pipeline.

PR #1575, opened 2026-06-07
Add length checks in `init_dungeon_dungeons`

0x0b1ade:
Fix a possible buffer overflow issue in
`init_dungeon_dungeons` (dungeon.c) by adding length checks. If the
buffer is too short, it will panic directly.

### DIFF
--- a/src/dungeon.c
+++ b/src/dungeon.c
@@ -1058,11 +1058,22 @@ init_dungeon_dungeons(
     pd->tmpdungeon[dngidx].chance = dgn_chance;
     pd->tmpdungeon[dngidx].entry_lev = dgn_entry;
 
-    /* FIXME: these should have length checks */
+    if (Strlen(dgn_fill) >= sizeof svd.dungeons[dngidx].fill_lvl)
+        panic("fill_lvl too long for dungeon %d", dngidx);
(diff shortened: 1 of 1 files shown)


nhmall:
Is this precautionary? Is it correct to state that  you didn't run into any such overflow with the distributed sources to date?
Bullet-proofing isn't a bad thing, but "possible buffer overflow issue" might sound more ominous than the actual situation.

0xb1aded:
> Is this precautionary? Is it correct to state that you didn't run into any such overflow with the distributed sources to date? Bullet-proofing isn't a bad thing, but "possible buffer overflow issue" might sound more ominous than the actual situation.

It's precautionary — I haven't encountered any overflow issue. However, the existing FIXME comment(L1061) shows this was already recognized as a concern. By "possible" I mean the code lacks stricter length checks — there is no safeguard to prevent overflow. The Lua strings are heap-allocated via `dupstr`, and `Strcpy` also performs no bounds checking. If the dungeon data is ever modified or extended in the future, a stack buffer could be silently overwritten, which is undefined behavior. Buffer overflow risks like this can only be prevented proactively.

nhmall:
This is in the pipeline.

PR #1576, opened 2026-06-07
Honor wraptext in curses persistent inventory

avgas3:
## Summary

This updates the curses `perm_invent` display so that when `wraptext` is enabled, long persistent inventory entries wrap within the inventory window instead of being clipped and horizontally panned.

The existing non-wrapped behavior is preserved: `perm_invent` still clips long lines and supports horizontal panning when `wraptext` is disabled.

## Details

- Count persistent inventory display height in visual rows when `wraptext` is enabled.
- Use the existing curses wrapping helpers, `curses_num_lines()` and `curses_break_str()`, for wrapped inventory text.
- Align wrapped continuation lines under the item text after the inventory letter prefix.
- Suppress horizontal panning controls while wrapped mode is active.
- Clamp row offsets after resize or option changes so stale scroll positions remain valid.

## Testing

- Built successfully in an existing build-ready environment.
- Verified manually in curses that wrapped persistent inventory entries respect terminal width when `wraptext` is enabled.
- Verified that existing clipped/panned behavior remains available when `wraptext` is disabled.
- Ran `git diff --check`.

### DIFF
--- a/win/curses/cursinvt.c
+++ b/win/curses/cursinvt.c
@@ -29,6 +29,13 @@ struct pi_line {
 };
 static struct pi_line zero_pi_line;
 
+#define PERMINV_ITEM_PREFIX 4 /* " x) " inventory letter prefix */
+
+static unsigned pi_line_textwidth(struct pi_line *, unsigned);
+static unsigned pi_line_rows(struct pi_line *, unsigned);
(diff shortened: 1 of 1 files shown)


nhmall:
The plan is to review this pull request this coming week (June 14-21). It requires testing on a few different platforms.

PR #1579, opened 2026-06-09
MS-DOS: several fixes for the tiled map

chasonr:
VESA mode: Set the viewport size correctly so the position bar does not overlay the map.

VESA mode: Correctly set the size to which tiles are stretched when overview mode (F4) is selected.

Both VESA and 16 color modes: Pass correct parameters to vga_userpan and vesa_userpan, so the pan keys (CTRL+arrow) work correctly. Some emulations don't report CTRL with up or down arrows, so accept CTRL-home, CTRL-page up, CTRL page down and CTRL-end.

16 color mode: Set the panning direction so CTRL-left and CTRL-right pan in the same direction as 3.4.3.

### DIFF
--- a/sys/msdos/pckeys.c
+++ b/sys/msdos/pckeys.c
@@ -14,7 +14,7 @@
 #include "pcvideo.h"
 
 boolean pckeys(unsigned char, unsigned char);
-static void userpan(boolean);
+static void userpan(enum vga_pan_direction pan);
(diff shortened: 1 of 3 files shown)


nhmall:
This is in the pipeline.

nhmall:
A new DOS binary is available at https://www.nethack.org.

PR #1590, opened 2026-06-13
build: serialize makedefs to fix the -jN host-tool race

ingpaschke:
prevent makedefs race with order-only rule.

### DIFF
--- a/sys/unix/Makefile.top
+++ b/sys/unix/Makefile.top
@@ -139,6 +139,12 @@ all:    $(ALLDEP)
 	true; $(MOREALL)
 	@echo "Done."
 
+# Several top-level targets reach the host 'makedefs' tool through their own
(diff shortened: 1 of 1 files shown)


nhmall:
Syntax that is GNU make only wont be accepted in Makefiles. Hints files are fine, but we’ve kept the Makefiles themselves  free of GNU-specific entries.

ingpaschke:
Pushed new version with .NOTPARALLEL: Works on GNU and BSD make, impact on build time is minimal.

nhmall:
.NOTPARALLEL does seem to be documented on GNU make and bsd make.
bsd:
https://man.openbsd.org/make#.NOTPARALLEL

GNU:
https://www.gnu.org/software/make/manual/html_node/Parallel-Disable.html

PR #1594, opened 2026-06-19
Fix: a shopkeeper msg was printed even when deaf

copperwater:
The message in question is '[shopkeeper] says "You be careful with my [item]!"' when you wield a shop-owned item, but it was being printed even when the hero is deaf.

Following other examples of shopkeeper dialogue, the correct thing to do here is not to suppress the message if deaf but instead provide some nonverbal feedback, so that is what I did.


nhmall:
This is in the pipeline now.

PR #1595, opened 2026-06-19
Fix: thrown ball travel did not check for regions

copperwater:
This is an issue that we discovered in TNNT last year when we added a custom region with effects that trigger upon entry: it was possible to bypass those effects by entering the region via a thrown iron ball.

This can be demonstrated by creating a poison gas cloud and then dragging oneself inside the cloud behind a thrown ball: you land in the cloud and are surrounded by poisonous gas, but are unharmed by it.

This commit fixes the iron ball code to call in_out_region when appropriate, which handles the side effects of entering and exiting regions in addition to preventing travel into or out of a hypothetical region that blocks entry or exit.

### DIFF
--- a/src/ball.c
+++ b/src/ball.c
@@ -808,7 +808,8 @@ drag_ball(coordxy x, coordxy y, int *bc_control,
                     miss(xname(uball), victim);
 
             } /* now check again in case mon died */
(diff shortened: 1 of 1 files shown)


nhmall:
This is now in the pipeline for merging.

PR #1597, opened 2026-06-21
Copy bones monster info to recorporealized ghost

copperwater:
This is a reimplementation of commit ec32748 in TNNT by entrez:

    If you use undead turning on a dead hero's corpse from a bones file,
    its ghost will get sucked back into its body when it comes back to
    life.  In such cases, treat the newly revived corpse/recorporealized
    ghost as a "former hero" for the purposes of livelogging bones
    monster kills.

Part of this change was to remove a bit of weirdness in the game - the odd situation where you revive a corpse-ghost combo on a bones pile, and get the messages:

    The human corpse glows iridescently.
    Foo's ghost is suddenly drawn into its former body!
    The human is resurrected!

and subsequently the monster is just a generic human (or elf, etc) without anything indicating it used to be a player. With this change, the monster will retain the name that the ghost had, and its struct ebones prevents the player from renaming it.

This commit also adjusts the "resurrected" message to explicitly use the monster's name if it has one, and to use YMonnam in the unrelated case where the monster was a pet (so reviving an unnamed tame kitten will print as "Your kitten is resurrected!")

We typically try to avoid having small gameplay differences like this in TNNT, but rather than removing the code, I figured I'd see if there was interest in merging it upstream.

### DIFF
--- a/src/zap.c
+++ b/src/zap.c
@@ -892,6 +892,10 @@ revive(struct obj *corpse, boolean by_hero)
     mmflags_nht mmflags = NO_MINVENT | MM_NOWAIT | MM_NOMSG;
     int montype, cgend, container_nesting = 0;
     boolean is_zomb;
+    /* normally use the name on the corpse object, but don't if drawing a
+     * player's ghost back into its body, in which case use the ghost's
+     * (player's) name */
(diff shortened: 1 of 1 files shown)


nhmall:
in the pipeline

PR #1602, opened 2026-06-24
add more options to  herecmdmenu

JodiJodington:
These are small commits that add:
engraving, jumping, praying, and viewing player attributes to the herecmdmenu. I also made the existing "Drop" button do a multi-drop instead of a single drop.
I wasn't sure if all of these were wanted, but I think at least some of these would be very useful.
I also want to point out that this menu is a new way to test if you have intrinsic/extrinsic jumping since the "Jump" entry only shows up if that is the case. I'm not sure if this is desireable but I cannot imagine it's that big of a deal (you can always use `#jump` to test that without taking a turn anyway).

Another idea I had was to let you eat/loot comestibles/containers that were on the pile you are on but not the top pile. I.e the menu would have an "Eat ..." option for each comestible and a "Loot ..." for each container, but the implementation ended up being too messy so I'll make that a separate PR.

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -4409,6 +4409,7 @@ enum menucmd {
     MCMD_LOOK_AT,
     MCMD_PRAY,
     MCMD_ENGRAVE,
+    MCMD_ATTRIBUTES,
     MCMD_ATTACK_NEXT2U,
     MCMD_UNTRAP_HERE,
     MCMD_OFFER,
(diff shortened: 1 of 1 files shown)


JodiJodington:
I just fixed up the commit messages and some typos and added another button that just shows previous messages.
I'm going to stop editing it now to let it sit and be stable, but there are probably more things I'd want to add in future PRs since the menu is quite tiny and there are a lot of hero-related things that could be there.

nhmall:
In the pipeline

JodiJodington:
@nhmall good to hear, thank you!

PR #1612, opened 2026-07-03
Some optimizations of the glyph lookup

chasonr:
* src/glyphs.c: Don't produce names for swallow glyphs that can never appear. This speeds up the building of the glyph index in populate_glyphname_hash_indices.

* src/symbols.c: Build a sorted index for loadsyms so that it can be searched with bsearch.

Benchmarking results, using the MS-DOS port on an emulated 386SX at 16 MHz; times are as measured with clock() from the entry to main() to the appearance of the "Who are you?" prompt:

* Unmodified:  48.5 seconds
* With the loadsyms index: 40.4 seconds
* With impossible swallow glyphs left unnamed: 34.5 seconds

These changes should also be helpful for the Amiga and Atari ST ports.

### DIFF
--- a/src/glyphs.c
+++ b/src/glyphs.c
@@ -373,8 +373,10 @@ compose_glyph_name(int glyph, char *buf, size_t bufsz)
             };
 
             j = glyph - GLYPH_SWALLOW_OFF;
-            cmap = glyph_to_swallow(glyph);
             mnum = j / ((S_sw_br - S_sw_tl) + 1);
+            if (!attacktype(&mons[mnum], AT_ENGL))
+                return 0;
+            cmap = glyph_to_swallow(glyph);
             Snprintf(tmpbuf[3], sizeof tmpbuf[3], "swallow %s %s",
                      monsdump[mnum].nm, swallow_texts[cmap]);
             buf3 = tmpbuf[3];
--- a/src/symbols.c
+++ b/src/symbols.c
@@ -7,6 +7,9 @@
 
 staticfn void savedsym_add(const char *, const char *, int);
 staticfn struct _savedsym *savedsym_find(const char *, int);
+staticfn const struct symparse *search_loadsyms(const char *buf, size_t len);
+staticfn int symparse_compare(const void *s1_, const void *s2_);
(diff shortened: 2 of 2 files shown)


nhmall:
I put this in the pipeline.
Upon doing so, I discovered that one of the compilers used in the CI is complaining about the argument to qsort
symbols.c(917,15): Warning C4090: 'function': different 'const' qualifiers.

I'm not sure if this would be an appropriate fix, or not:
```
diff --git a/src/symbols.c b/src/symbols.c
index b1f459341..a2109ad8c 100644
--- a/src/symbols.c
+++ b/src/symbols.c
@@ -905,9 +905,9 @@ search_loadsyms(const char *buf, size_t len)

     /* Sorted index to loadsyms */
     static boolean first_time = TRUE;
-    static const struct symparse *loadsyms_sorted[SIZE(loadsyms) - 1];
+    static struct symparse *loadsyms_sorted[SIZE(loadsyms) - 1];
     if (first_time) {
-        const struct symparse *sp = loadsyms;
+        struct symparse *sp = (struct symparse *)loadsyms;
         size_t i = 0;
         while (sp->range) {
             loadsyms_sorted[i] = sp;
```

chasonr:
Might want a comment to explain why the const qualifier is being dropped, lest some well-intentioned person should try to put it back.

nhmall:
There is this alternative too:
```
diff --git a/src/symbols.c b/src/symbols.c
index b1f459341..904aa6fd8 100644
--- a/src/symbols.c
+++ b/src/symbols.c
@@ -914,8 +914,8 @@ search_loadsyms(const char *buf, size_t len)
             ++sp;
             ++i;
         }
-        qsort(loadsyms_sorted, SIZE(loadsyms_sorted), sizeof(loadsyms_sorted[0]),
-                symparse_compare);
+        qsort((struct symparse *)loadsyms_sorted, SIZE(loadsyms_sorted),
+              sizeof(loadsyms_sorted[0]), symparse_compare);
         first_time = FALSE;
     }
```

chasonr:
I like that, but I would use (struct symparse **).

nhmall:
> I like that, but I would use (struct symparse **).

yes, right

PR #1617, opened 2026-07-05
    MS-DOS: some optimizations for tile handling

chasonr:
* In both the 16 color and the VESA mode, the tileset image is loaded and split into individual tiles; the tiles are then processed into a form that is compatible with the video mode in use. For 16 color mode, a tile is processed each time it is displayed, leading to slow redrawing. For VESA mode, each tile is processed at startup, leading to long startup times. Both modes are changed so that the tile is processed once, when it is first displayed, and the result is cached.

* Use memcpy when splitting the image into tiles.

* Only load the tileset once. In 16 color mode, for reasons I do not understand, the gr_init function is called twice, leading to delay in startup. This does not happen in VESA mode.

### DIFF
--- a/sys/msdos/vidvesa.c
+++ b/sys/msdos/vidvesa.c
@@ -60,6 +60,8 @@ static unsigned long vesa_DoublePixels(unsigned long);
 static unsigned long vesa_TriplePixels(unsigned long);
 static void vesa_WriteStr(const char *, int, int, int, int);
 static unsigned char __far *vesa_FontPtrs(void);
+static unsigned char *vesa_tile(unsigned);
+static unsigned char *vesa_oview_tile(unsigned);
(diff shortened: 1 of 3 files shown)


nhmall:
The shared win/share/tileset.c changes reportedly cause a compile error on at least one platform, and a warning on another. 

../win/share/tileset.c:34:9: error: 'return' with no value, in function returning non-void [-Wreturn-mismatch]

chasonr:
This is currently only used by MS-DOS, but I think Amiga is compiling it and not actually using it.

nhmall:
This is in the pipeline.

nhmall:
> This is currently only used by MS-DOS, but I think Amiga is compiling it and not actually using it.

The Atari build in m68k-wip branch was getting the error.

PR #1620, opened 2026-07-07
MS-DOS: Define a separate palette for VGA text

chasonr:
In 16 color mode, text colors are mapped onto the colors available in the tileset. The mapping is a compromise, and is not one-to-one: cyan and bright cyan are the same, and magenta and bright magenta are the same.

This change defines a separate palette for use by the text map, and switches palettes on any transition between text and tiled maps.


nhmall:
This is in the pipeline.

PR #1627, opened 2026-07-09
sfstruct: buffered stdio writes on amiga and atari st

ingpaschke:
Enable USE_BUFFERING for the structlevel savefile write path on AMIGA and TOS via a new SFSTRUCT_BUFFERING config define, with a 16 KB setvbuf on the fdopen'd write stream so many small bwrite()s coalesce into few dos.library Write() packets.  Leaves mread() unbuffered.


nhmall:
Shouldn't the define for BWBUFSZ be within an #if defined(SFSTRUCT_BUFFERING) block too, since it appears to have no application outside of that?

`#define BWBUFSZ 16384`

ingpaschke:
Yes, you're right. I updated the PR.

PR #1629, opened 2026-07-09
Use fqname for save (de)compression in plname_from_file

### LINKED ISSUE #1628: Loading a saved game fails when the save folder isn't the game folder
### What happens
Starting NetHack with an existing saved game shows a garbled "Who are you?"
screen with a message like:

```
    freopen of save/501player for compress failed; (2) No such file or directory
```
and the game usually crashes right after.

### When
When the save directory (`VAR_PLAYGROUND`) is a different folder from the one
NetHack switches into at startup (`HACKDIR`). Packaged builds such as Homebrew
set these to different folders.

### Cause
In `src/files.c`, `plname_from_file()` (un)compresses the save file using a
**relative** path, so it's looked up in the wrong folder. Everywhere else the
full path is built with `fqname()`.

### Possible Fix
Use `fqname()` for those two calls, like the rest of the save code:

    nh_uncompress(fqname(gs.SAVEF, SAVEPREFIX, 0));
    ...
    nh_compress(fqname(gs.SAVEF, SAVEPREFIX, 0));

daeho-ro:
Will resolve https://github.com/NetHack/NetHack/issues/1628

### DIFF
--- a/src/files.c
+++ b/src/files.c
@@ -1365,7 +1365,7 @@ plname_from_file(
             gs.SAVEF[sln - xln] = '\0';
     }
 #endif
-    nh_uncompress(gs.SAVEF);
+    nh_uncompress(fqname(gs.SAVEF, SAVEPREFIX, 0));
     if ((nhfp = open_savefile()) != 0) {
         if ((sfstatus = validate(nhfp, filename,
                                 without_wait_synch_per_file,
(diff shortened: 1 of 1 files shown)


pat-rankin:
There's.a reason that fork+exec of compress does not use absolute paths.  Unfortunately I don't remember what that reason is.  Permission issues with the subprocess most likely.

PR #1630, opened 2026-07-09
Add secondary effects of different dragon scales for monsters.

### LINKED ISSUE #1555: New dragon scale mail effects do not work on monsters
My pet Master Mind Flayer just got stoned despite wearing Yellow Dragon Scale Mail.

While some effects are not applicable for monsters (such as slow digestion from white), the ones that should be applicable(stoning resistance is very much a thing for monsters and player alike) do not work either.

SnivyLlama:
Mostly fixes #1555, I'm not sure it is worth implementing the other (unimplemented) dragon scale types.

### DIFF
--- a/src/eat.c
+++ b/src/eat.c
@@ -639,19 +639,20 @@ eat_brains(
                 make_stoned(5L, (char *) 0, KILLED_BY_AN,
                             pmname(pd, Mgender(mdef)));
         } else {
-            /* no need to check for poly_when_stoned or Stone_resistance;
-               mind flayers don't have those capabilities */
-            if (visflag && canseemon(magr))
-                pline("%s turns to stone!", Monnam(magr));
-            monstone(magr);
-            if (!DEADMONSTER(magr)) {
-                /* life-saved; don't continue eating the brains */
-                return M_ATTK_MISS;
-            } else {
-                if (magr->mtame && !visflag)
-                    /* parallels mhitm.c's brief_feeling */
-                    You("have a sad thought for a moment, then it passes.");
-                return M_ATTK_AGR_DIED;
+            /* mind flayers can wear yellow dragon scales and gain stone resistance */
+            if (!resists_ston(magr)) {
+                if (visflag && canseemon(magr))
+                    pline("%s turns to stone!", Monnam(magr));
+                monstone(magr);
+                if (!DEADMONSTER(magr)) {
+                    /* life-saved; don't continue eating the brains */
+                    return M_ATTK_MISS;
+                } else {
+                    if (magr->mtame && !visflag)
+                        /* parallels mhitm.c's brief_feeling */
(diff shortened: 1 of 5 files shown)


youbo0:
There's other intrinsics already implemented for monsters that dragon scales should give.
Notably, they should get speed from blue scales like they already do with speed boots. 
Many monsters also already have drain resistance(black) and infravision(red).

The remaining colors, in order of applicability:
- Green (sickness): Sickness resistant monsters are already implemented... sort of. They are protected from Pestilence and other sickness attacks, but it is done by checking for ghoul and fungi. https://github.com/NetHack/NetHack/commit/017a8687e0b541a8e5848ef3037e25dc819208ca
- Gold (hallucination) is non-applicable for monsters. Well... you could make it protect them from ad_halu hallucination attacks from black lights, though that applies confusion to monsters instead. https://github.com/NetHack/NetHack/blob/05d1ed365fd1b409e46eb1accc9c1897a6c3c1f9/src/uhitm.c#L3897

SnivyLlama:
Black and green dragon scales already work properly last I remember, and infravision doesn't even work for monsters. Blue dragon scales would be easy to implement (just add them to speed boots case) and gold dragon scales could also protect from AD_HALU like you said.

youbo0:
Oh, you're right on Black, Green and Red. Wiki simply says "many monsters have [infravision]" so I assumed it did something and was looking at old commits for mhitm_dise(), sorry. 

The set looks fairly complete now. The only possible improvement I can think of is white(slow digestion) on tame monster nutrition if you want to be really thorough.

SnivyLlama:
For sake of completeness, I may do it when I have spare time (slow digestion is a bit harder since pet hunger is much different from player hunger, so it would be an effect solely on wear/removal) but I don't really think that any player in the position to give a humanoid pet white dragon scale (mail) really cares about the slow digestion.

Otherwise I'll just let this sit.

youbo0:
Not very familiar with this codebase, but would incrementing tame monsters' `hungrytime` once per turn if wearing white scales be an easy solution? Maybe around here since we're already looping through the monsters?
https://github.com/NetHack/NetHack/blob/f41f1aa09a76e6c40aacdb94e31a816426409b9b/src/allmain.c#L229
(Side note: I first looked at dog_hunger(), and I might be misunderstanding, but that one is ran less or more often than once per turn according to the monster's speed, right? Doesn't that causes a bug where slow tame monsters might starve on a turn later than intended?)

But yes, I'm personally quite satisfied with what this PR already adds in terms of gameplay regardless of white; I can't see that one mattering outside of completeness' sake unless the player is trying some weird non-standard self-imposed challenge. Thanks for making this!

ars3niy:
Do pets still vacuum-clean the floor of all corpses regardless of how much they already ate? Cause if they didn't, white scales would be a good buff, less eating more fighting.

PR #1641, opened 2026-07-26
themerms: keep Teleportation hub traps off the wall-adjacent gutter

### LINKED ISSUE #1640: teleportation hub room can yield a map that traps the hero by making stairs inaccessible
I think the teleportation hub themed room can make the stairs on their level inaccessible. Screenshot:

Image

In this map, all three teleportation traps teleport to the left two rooms in the level and there are no paths out.  I encountered this at Dlvl 3 with no way to dig or disable or jump over or circumvent these traps.  Maybe I'm missing some possible gameplay trick to get out, but it feels like a boring game just being stuck in three rooms, and it doesn't seem very NetHack-like.

Do (or should) teleportation hubs have some logic to avoid creating inaccessibility traps like this?  Maybe teleportation hubs should avoid blocking the doorways?

davidbau:
A fix for #1640.

The Teleportation hub fill scatters 2-4 "fixed-destination" teleport traps across a room's whole floor. An unlucky room can end up with traps on every tile in front of its only doorway, all teleporting to the far side of the level, leaving no trap-free way through the room. (See #1640 for an example. Level connectivity is built assuming every room is freely passable, so a bad arrangement of fixed-teleport traps can make the level impassable.)

The change places the traps on the room's interior only, never the perimeter "gutter" of floor tiles next to a wall. A clear gutter always leaves a trap-free walk around the room's edge between every door. The trap count is capped at the number of interior tiles, so small rooms place fewer traps and a room with no interior places none.

### DIFF
--- a/dat/themerms.lua
+++ b/dat/themerms.lua
@@ -264,8 +264,17 @@ themeroom_fills = {
    {
       name = "Teleportation hub",
       contents = function(rm)
-         local locs = selection.room():filter_mapchar(".");
-         for i = 1, 2 + nh.rn2(3) do
+         -- Only trap the room's interior, never the "gutter" of floor
+         -- tiles next to a wall.  Keeping the gutter clear guarantees a
+         -- trap-free walk around the perimeter between every door, so these
+         -- fixed-destination teleporters can't ring a doorway, seal the
+         -- room, and strand the hero on the wrong side of the level.
+         local floor = selection.room():filter_mapchar(".");
+         local locs = floor - floor:negate():grow("all");
+         -- never ask for more traps than there are interior tiles; a room
+         -- with no interior (smaller than 3x3, which ordinary rooms never
+         -- are) simply gets none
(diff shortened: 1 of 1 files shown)


ars3niy:
I believe one can have this weaker set of constraints instead:

- a teleporter in the corner of the room cannot have another teleporter grid-bug-adjacent to it (i.e. vertically or horizontally);
- a "grid-bug-connected" set of teleporters, i.e. a set where a grid bug can travel from any one to any other while staying within the set, cannot contain two teleporters that are on the edge of the room and not grid-bug-adjacent to each other;
- there cannot be more than 7 teleporters in the room.

I think this works for non-rectangular rooms as well, with the modification that such rooms have more than four corners.

davidbau:
interesting.  yes, i think that as long as traps don't bisect a room in the wrong way it wouldn't cause this problem.

But also I think: clustering the teleportation hubs in the center of the room "looks more intentional," which might be a nice effect. I imagine it might be the layout a dungeon architect might pick it if really designing a hub.

PR #1649, opened 2026-08-04
MS-DOS: Complete highlighting of piles and pets

chasonr:
* Support decals in all VESA modes and in 16 color mode.
* Support decals in the overview modes as well as the normal map mode.
* Provide correct highlighting for vga_redrawmap and vesa_redrawmap.
* Provide the hilite_pet and hilite_pile flags in the default configuration file.


nhmall:
That's great! This is in the pipeline.

PR #1654, opened 2026-08-09
X11: Fix resizing bug in TTY status window, and implement all attributes

chasonr:
If the TTY status window is in use, and the main window is resized, the various components get smushed together and the status window becomes unreadable.

This change reimplements the TTY status window, using the existing generic Window class and rendering the text onto it.

Several new features are added as well:
* all text attributes: bold, dim, italic, blink, underline and inverse
* 3 line status
* weapon, armor and terrain indicators
* a hit point bar


paxed:
Looks good.

One nit: hitpointbar doesn't obey hitpoint status hilites.

chasonr:
Reimplemented the hit point bar to use the hit point colors.

chasonr:
I've just discovered, quite by accident, that the unmodified TTY status window dereferences a null pointer if any of armorstatus, weaponstatus or terrainstatus is set in .nethackrc. This pull request fixes that, and also implements the corresponding fields in the TTY status window. (The fancy status is unchanged.)

paxed:
This is in the pipeline.

chasonr:
I grepped for those and didn't find any in the final commit. The ones in commit b044725 marked those items that I had actually seen on the status bar, and so had some confidence that they were correctly implemented.

nhmall:
Yes, I was reading the PR and noticed them. I removed the question altogether after reviewing the commits in GitHub, so I deleted the question, but not in time. I should have just left it with a follow-up edit, because I likely confused people in doing so. The original question I asked was: "What do the `/*@@@*/` comments signifiy"

nhmall:
I noticed that no link to the implementation commits made it onto the pull request in GitHub, likely because it was closed prior to being pushed. I'm going to re-open the pull request and then close it again, because I'm curious what GitHub will do in that case.

Please ignore the re-open and re-close.

nhmall:
Conclusion: It did not help.

chasonr:
Make that commit 725bb3d, then. Just to maintain the record after I delete the branch.

Issue #1508, opened 2026-05-03
X11 Unicode font initialisation fails fatally when fonts are not available


OldUser101:
#923 adds Unicode support via the `ENHANCED_SYMBOLS` flag, when enabled and appropriate fonts are not available the `XQueryFont` invocation [here](https://github.com/NetHack/NetHack/blob/c2a7bdb3690432bc83fd193399f25fa82e963d9a/win/X11/winmap.c#L1851) seems to trigger an exception handler rather than using a fallback.

I don't know much of X11, but it would appear this is the case. Disabling `ENHANCED_SYMBOLS` prevents this.

```
Warning: Cannot convert string "-*-fixed-medium-*-*-*-15-*-*-*-*-*-*-*" to type FontStruct
X Error: code 15 (BadName (named color or font does not exist)), request 45, minor 0, serial 68
Program initialization has failed.
To report this error, contact root or games.
X Error
  Generating more information you may report:

[00] /nix/store/q5f7ci8rq3km6mknndp0b4acb9lh9043-nethack-x11-5.0.0/games/lib/nethackdir/nethack(NH_panictrace_libc+0x47) [0x63f2b86fffc7]
[01] /nix/store/q5f7ci8rq3km6mknndp0b4acb9lh9043-nethack-x11-5.0.0/games/lib/nethackdir/nethack(NH_abort+0x8c) [0x63f2b85eaa0c]
[02] /nix/store/q5f7ci8rq3km6mknndp0b4acb9lh9043-nethack-x11-5.0.0/games/lib/nethackdir/nethack(panic+0x1f7) [0x63f2b85eaea7]
[03] /nix/store/q5f7ci8rq3km6mknndp0b4acb9lh9043-nethack-x11-5.0.0/games/lib/nethackdir/nethack(+0x327516) [0x63f2b87dd516]
[04] /nix/store/5m91jqg1526jzsahrgmd37k4ml3nc5l4-libx11-1.8.13/lib/libX11.so.6(_XError+0x13c) [0x700ad3df1a4c]
[05] /nix/store/5m91jqg1526jzsahrgmd37k4ml3nc5l4-libx11-1.8.13/lib/libX11.so.6(+0x4a110) [0x700ad3dee110]
[06] /nix/store/5m91jqg1526jzsahrgmd37k4ml3nc5l4-libx11-1.8.13/lib/libX11.so.6(+0x4a1dd) [0x700ad3dee1dd]
[...]

Issue #569, opened 2021-08-08
Permanent inventory menu causes a segfault in the X11 window port.


NullCGT:
Environment: Ubuntu Linux 5.8.0-63 Generic Kernel.
Window Manager: i3

Enabling the permanent inventory option results in a blank menu which takes focus from the game, making gameplay impossible. Interacting with it causes a segfault.

This is not noted in the X11 issues document.

pat-rankin:
It works for me on OSX.  I won't claim that it works well because having it overlap the map window is a nuisance.  But if you resize and/or move them so that they don't overlap, it's functional.

When you say blank, do you mean completely blank or a window with the OK/Cancel/Select/&c buttons but no inventory items?

If completely blank, how do you interact with it?  Or more generally, how do you interact with it?  If you launch it synchronously from a terminal, you might get a panic trace displayed there.  (Again, it works for me....)

Clicking on the map window ought to shift focus back to that.  That might be needed when it first pops up but after that focus should stay with the map unless you drag the inventory window around or give it focus in order to scroll it.

NullCGT:
The window exists, but the OK/Cancel/Select/&c buttons have no text in them, and none of the inventory items are displayed. By interacting with it, I mean I tried clickng on a few of the blank buttons to see what would happen. I wasn't able to click on the map to bring focus back to it either.

NullCGT:
The window exists, but the OK/Cancel/Select/&c buttons have no text in them, and none of the inventory items are displayed. By interacting with it, I mean I tried clickng on a few of the blank buttons to see what would happen. I wasn't able to click on the map to bring focus back to it either.

paxed:
Works for me too on Debian testing.

NullCGT:
Huh. Thank you for checking. Perhaps there's some sort of issue in my X11 settings somewhere.

nhmall:
Could you post the details of the segfault to show whether it is in NetHack code or within the X11 library code?

NullCGT:
I'm not sure whether this is the same segfault, but I just managed to get a segfault in Arch Linux with X11.

```
Warning: Cannot convert string "-*-fixed-medium-*-*-*-15-*-*-*-*-*-*-*" to type FontStruct

Signal received.
  Generating more information you may report:

[00] src/nethack(+0x10ca0b) [0x55bcf95d5a0b]
[01] src/nethack(+0x10c9a6) [0x55bcf95d59a6]
[...]

PR #1658, opened 2026-08-17
X11: Render text with Xft

chasonr:
Adding Xft support gives these advantages:
* Support for TrueType and OpenType fonts. The user can select any font configured on their computer.
* Support for the Unicode supplementary planes.

At this writing, only a few text input widgets still use the bitmap fonts. These are proving to be tough to convert and I'm leaving them for another day.

### DIFF
--- a/include/winX.h
+++ b/include/winX.h
@@ -11,6 +11,9 @@
 
 #ifdef USE_XFT
 #include <X11/Xft/Xft.h>
+typedef XftFont X11_Font;
+#else
+typedef XFontStruct X11_Font;
 #endif
 
 #ifndef COLOR_H
@@ -529,6 +532,7 @@ extern void X11_set_attrs(Widget, unsigned);
 extern void X11_set_highlight(Widget, boolean);
 extern void X11_set_percent(Widget, unsigned, Pixel);
 extern void X11_blink_labels(void);
+extern int X11_font_height(X11_Font *);
 
 /*
  * These are for widgets that use the enhanced label services only for text
--- a/include/winX.h
+++ b/include/winX.h
@@ -382,7 +382,7 @@ extern Boolean nhCvtStringToPixel(Display *, XrmValuePtr, Cardinal *,
 extern void get_window_frame_extents(Widget, long *, long *, long *, long *);
 extern void get_widget_window_geometry(Widget, int *, int *, int *, int *);
 extern char *fontname_boldify(const char *);
-extern Dimension nhFontHeight(Widget);
+extern Dimension nhFontHeight(Widget, int);
 extern char key_event_to_char(XKeyEvent *);
 extern void msgkey(Widget, XtPointer, XEvent *, Boolean *);
 extern void highlight_yn(boolean);
--- a/include/winX.h
+++ b/include/winX.h
@@ -84,7 +84,7 @@ struct text_map_info_t {
         square_ascent, /*   placement of changes.        */
         square_lbearing;
 
-#ifdef ENHANCED_SYMBOLS
+#if defined(ENHANCED_SYMBOLS) && !defined(USE_XFT)
     XFontStruct *font;
 #endif
 };
@@ -131,7 +131,12 @@ struct line_element {
 };
 
 struct mesg_info_t {
+#ifdef USE_XFT
+    Pixel fgpixel;   /* Color for text drawing */
+#else
+    GC gc;           /* GC for text drawing */
     XFontStruct *fs;                 /* Font for the window. */
+#endif
     int num_lines;                   /* line count */
     struct line_element *head;       /* head of circular line queue */
     struct line_element *line_here;  /* current drawn line position */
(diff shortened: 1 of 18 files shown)


paxed:
Perhaps I'm missing some fonts, but the status lines are all squished together without spaces between eg the attributes, so it looks like "St:17Dx:17Co:18" etc

chasonr:
Proportional fonts often have very narrow space characters. I've made that spacing equal to half the line height.

nhmall:
Package-builders, for at least one distro, have encountered problems with startup failures on the NetHack X11 binary if certain fonts or prerequisites aren't available. Past discussion of the issue seems to have implicated the use of XQueryFont().

Evolution of the X11 window port to use Xft will hopefully resolve those issues and make things better for package-maintainers.

References:
[Issue #1508](https://github.com/NetHack/NetHack/issues/1508)
[Issue #569](https://github.com/NetHack/NetHack/issues/569)
https://github.com/NixOS/nixpkgs/pull/516017#pullrequestreview-4216274108

nhmall:
Just for information, while testing this pull request on **macOS**, I had to complete the following, to get a successful build:
```
brew install libxft
brew install fontconfig
brew install freetype
```

Also, this follow-up patch:
```
diff --git a/sys/unix/NewInstall.unx b/sys/unix/NewInstall.unx
index 9ced987c1..538b18497 100644
--- a/sys/unix/NewInstall.unx
+++ b/sys/unix/NewInstall.unx
@@ -98,6 +98,9 @@ additional interfaces. See below.
 |          |       |                 |                                      |
 |          |       |                 | One possible way:                    |
 |          |       |                 |    brew install xquartz              |
+|          |       |                 |    brew install libxft               |
+|          |       |                 |    brew install fontconfig           |
+|          |       |                 |    brew install freetype             |
 |----------+-------+-----------------+--------------------------------------|
 | MacOS    | Qt    | WANT_WIN_QT     | You will need to obtain and          |
 |          |       |                 | install Qt if you want Qt            |
diff --git a/sys/unix/hints/macOS.500 b/sys/unix/hints/macOS.500
index 06646dab0..36436ed99 100755
--- a/sys/unix/hints/macOS.500
+++ b/sys/unix/hints/macOS.500
@@ -291,7 +291,9 @@ X11CFLAGS += -Wno-variadic-macros
 ifdef USE_XFT
 CFLAGS += -DUSE_XFT
 X11CFLAGS += $(shell pkg-config --cflags xft fontconfig)
+X11CFLAGS += $(shell pkg-config --cflags freetype2)
 WINX11LIB += $(shell pkg-config --libs xft fontconfig)
+WINX11LIB += $(shell pkg-config --libs freetype2)
 endif
 ifdef USE_XPM
 CFLAGS += -DUSE_XPM

```
I then built (almost cleanly) with:
```
sys/unix/setup.sh sys/unix/hints/macOS.500
make WANT_WIN_X11=1 WANT_WIN_CURSES=1 WANT_WIN_TTY=1 WANT_MACSOUND=1 resp=1 bundle
```

chasonr:
I've replicated the crash described in #1508, as described in that issue, and submitted a fix as a separate pull request. The Xft-enabled configuration should indeed not have that problem.

nhmall:
The additional changes in 0c93981192f615c5f87a52d5062c1e2c45c7680f build-tested OK on macOS.

paxed:
A bug with perm_invent: Changing levels refreshed the window and the calculated label sizes seemed to change or something; the lines were cropped strangely or the window gained a horizontal scrollbar. Seemed to be more visible if the used fonts were of different sizes.

chasonr:
I ran a test with perm_invent set and the window did in fact sprout a horizontal scroll bar when first updated. It happened when I picked up a scroll, and if I saved and restarted, it happened again when I changed levels.

Setting some highlights to include inverse might reveal if any labels are being resized.

chasonr:
I set `MENUCOLOR=".*"=nocolor&inverse` as the only menu highlight, and it shows that the labels are gaining some width on the right when the inventory window is first updated.

chasonr:
It appears that when the item widgets are recreated on an already-existing window, they don't get resized when the pixmaps are set up. They remain sized according to the bitmap font loaded by the Xaw library, and that ends up being wrong when we use Xft to size and fill the pixmap. This might be because the parent Form widget is already managed and realized.

I changed this part to use XtCreateWidget, creating the widget in an unmanaged state, so that when it becomes managed, the pixmap is already in place. The widgets seem to get the correct size now.

paxed:
One more thing: Looks like the font on my system is slightly cut off from the bottom in menus; looking at the full options list, the underscore in eg "tiled_map" isn't visible. Another way to see this is to call an object in inventory with underscores.

chasonr:
Confirmed, for the message, TTY status, map and menu windows. Adjusted so that the font height is the greater of `height` or `ascent+descent`. Underscores now display on all windows.

This has a side effect that bugs me slightly: the box-drawing characters on the map are no longer continuous.

The monospaced font (DejaVu Sans Mono) that gets chosen on my system (Linux Mint) gets the underscore cut off; the proportional font (DejaVu Sans) does not. (Do `fc-match mono-10` and `fc-match sans-10`.) I am reluctant to be more specific than "mono-10" and "sans-10", because I don't know what fonts a particular system might have. The changed code does honor the .nethackrc settings for fonts.

paxed:
This is now in the pipeline.

Issue #1656, opened 2026-08-15
Would you consider i18n support in the main repository?


g4b0:
Hello,

I would like to ask a question before writing any code, so I don't repeat what happened with #1478.

## Why

People have been translating NetHack for a very long time: JNetHack (Japanese), NetzHack and NetHack-De (German), Spanish NetHack, KRNethack and NetHackKR (Korean), NetHackJP (Japanese, 5.0), and Ray Chason's "Internationalized NetHack". They keep appearing because people love this game and want to play it in their own language.

I'm Italian. I can read English fine, but after a long day I play more comfortably - more *relaxed* - in my own language, and I know many players who feel the same. I also played NetHack with my son, and I had to translate every message out loud as we went. He would have learned the game much faster if it had spoken Italian to him.

I understand this is not an easy task. It's not just strings: it's grammatical gender, articles, plural rules, and word order that don't work like English. That's exactly why every attempt so far has been so expensive.

## Why in the main repository, and not in yet another fork

Every project in the list above is a separate fork, and almost all of them are dead. The reason looks structural rather than personal: a fork has to carry a full localized copy of the game and rebase it forever. When one maintainer runs out of free time, that language dies - and the next language starts again from zero. Eight attempts, the same outcome.

If the *mechanism* lived here instead, a translation would be **data**: one text file per language, contributed and maintained by the people who actually speak it, without touching core code. The DevTeam would never have to review a translation in a language you don't read - only the mechanism itself. That is the difference between one more abandoned fork and something the community can keep alive.

That's why I'd rather ask you first than build a ninth fork.

## About #1478

I read it, and I think paxed's objections were fair. I would want to address them from the start:

- **no context for the translatable strings** - context-tagged lookups, so a translator can disambiguate identical strings and can tell an informational message from a joke.
- **plural forms lacking** - real plural rules per language, not a singular/plural boolean (Slavic languages need three or four forms).
- **duplication of special level lua code** - no duplicated `dat/*.lua`. The ~990 translatable literals in the Lua levels need a lookup, not a copy of the tree.
- **LLM use** - see below.

## How I would want to work

Small, single-purpose, reviewable commits. Mechanism first. With no language selected the English behaviour stays byte-identical, and no translation content goes into core files.

[...]

PR #1661, opened 2026-08-20
X11: support tabbed menu columns

chasonr:
Set `iflags.menu_tab_sep` and respond to tabs in menu items. Change the default menu font to sans-10.

### DIFF
--- a/include/winX.h
+++ b/include/winX.h
@@ -531,8 +531,10 @@ extern void X11_update_label(Widget);
 extern void X11_set_attrs(Widget, unsigned);
 extern void X11_set_highlight(Widget, boolean);
 extern void X11_set_percent(Widget, unsigned, Pixel);
+extern void X11_set_column_widths(Widget, const int *, unsigned);
 extern void X11_blink_labels(void);
 extern int X11_font_height(X11_Font *);
+extern int X11_column_width(Display *, X11_Font *, const char *, size_t);
 
 /*
  * These are for widgets that use the enhanced label services only for text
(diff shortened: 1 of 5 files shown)


chasonr:
Marking this draft for the moment.

It has one glitch that bugs me somewhat: headings get aligned with the item names rather than the left margin. That isn't the thing I'm after right now. I checked the Qt interface to see if it has the same problem, which it does, but also noticed that the menu had been changed to a fixed width font. This was commit c96a9ff. I specifically remembered supporting proportional fonts in Qt 4; it began its existence in nethack-i18n, which, as discussed in #1656, has to support tabbed columns.

So the problem that led to the Qt menus being made fixed width will apply to X11 (and Win32) also.

chasonr:
Changing the relevant menu (doextlist()) solved that problem, and created another. The command listings at the top become entirely part of the first column, causing the actual help listing to have a very wide first column. This affects both X11 and Qt.

chasonr:
A modest change to the column width algorithm seems to resolve the issue, as far as the X11 interface itself is concerned. The last column of a line is deemed to extend to the right margin, and does not extend the width of that same column in other lines.

I have confirmed that a tabified extended command list displays correctly with a proportional font. But I have reverted the change of NetHack.font_menu. Tabifying the menus will affect the Qt and win32 ports. I will leave that for another pull request.

This pull request, as it now stands, should have no obvious effect on the X11 interface unless you change font_menu to a proportional font, whether in NetHack.ad or in .nethackrc.

paxed:
Is it intentional that the inventory and stats block in the character reroll window are indented? I'm not saying it's necessarily bad, just different from what it was before.

chasonr:
I've done the before and after on that menu and I see what you mean. No, it isn't intentional. But there are cases where we would want to indent a line without a selector; say, in the options menu, which displays options that can't be selected. I would need to know which lines should be indented.

paxed:
This is in the pipeline.

chasonr:
Closing this; the changes are in the mainline now.

PR #1662, opened 2026-08-21
fix: prevent scrollbar flickering in curses interface

0xb1aded:
The scrollbar characters in curses_draw_map() had their color256
field uninitialized. When curses_has_256color() is true,
get_framecolor() reads this garbage value, producing random color
pairs that manifest as black blocks and color flicker at the map
window edges when the view scrolls.


0xb1aded:
Please focus on the state of the scrollbar when hero moves. This example may not be obvious enough, but to limit the file size, I cannot record a very long video.

## *before*

https://github.com/user-attachments/assets/c415f928-f735-4b99-8346-4e06e5dc2d9f

## *after*

https://github.com/user-attachments/assets/67b26d03-3350-4143-b1a2-0195bd7dac09

PR #1663, opened 2026-08-22
Qt and Win32: fix menu columns on some menus

chasonr:
In the core: add tabs to these menus:
- extended command list
- status hilites
- disclose option
- symset selection

Qt and Win32: revise the column width algorithm so that the last column of a line can overhang subsequent columns of other lines.

Qt: change the menu font to proportional and improve the column padding algorithm as follows:
- pad with U+200A HAIR SPACE for finer control of padding
- account for any imprecision in padding width when padding subsequent columns, so that errors cancel and the last column is accurately placed

### DIFF
--- a/src/botl.c
+++ b/src/botl.c
@@ -4539,9 +4539,13 @@ status_hilite_menu(void)
 #endif
         any = cg.zeroany;
         any.a_int = fld + 1;
-        Sprintf(buf, "%-18s", initblstats[i].fldname);
+        if (iflags.menu_tab_sep) {
+            Sprintf(buf, "%s\t", initblstats[i].fldname);
+        } else {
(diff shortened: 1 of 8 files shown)


chasonr:
For Win32, the extended command list without this change looks like this:
nh-menu-original
The font is proportional and the columns do not align.

With tabs in the menu, but no change to the column sizing, we get this:
nh-menu-tabbed
The columns align, but the three lines at the top are counted in the first column and they make that column much too wide.

With the column sizing changed, we get this:
nh-menu-fixed
The three lines at the top are no longer included in sizing the first column. Rather, they overhang the second and third columns.

paxed:
Qt change works for me.

nhmall:
This is in the pipeline.

PR #1681, opened 2026-09-18
polyself: handle a second form change that happens during a polymorph

davidbau:
## Summary

Three defects in `src/polyself.c` share one cause. A polymorph is not a single
step: `polymon()` installs the new form and then applies several hundred lines
of consequences, and several of those consequences can change the hero's form
again before the sequence finishes. `losehp()` calls `rehumanize()` when `u.mh`
reaches zero, and `instapetrify()` and `selftouch()` call `polymon()` directly.
Once that has happened, the rest of the sequence is still acting on the form it
started with.

| symptom | cost to the player |
|---|---|
| `del_light_source()` is asked to remove a light source that was never created | `impossible()`, "Program in disorder!", and a request to mail the DevTeam; a hard stop under `fuzzer_impossible_panic` |
| `encumber_msg()` and `retouch_equipment()` run twice for one polymorph | a cross-aligned artifact blasts twice and rolls damage twice |
| `break_armor()` keeps stripping by the superseded form's rules | a **human**'s water walking boots are removed while they stand in lava; fatal |

Three commits, one per defect. They touch different lines, none depends on
another, and they cherry-pick cleanly in any order:

| commit | change | lines |
|---|---|---|
| `1aacf75f1` | five early returns in `polymon()`, at each call that can change the form | +25 −5 |
| `e21c87732` | move the hero's light-source transition into a helper called from every place that installs a form | +31 −23 |
| `2dc124672` | three re-checks inside `break_armor()`, at sub-block boundaries | +15 −0 |

Combined: `src/polyself.c` +70 −25 and `src/timeout.c` +1 −3, nothing else.

Each defect has a recorded before-and-after pair on the same seed, datetime and
keystream, steppable frame by frame in a browser with no build required, plus
two controls that isolate the mechanism. Those and the full analysis are here:

https://github.com/davidbau/nethack-bugreport/tree/main/bugs/10-polyself-reentrant-form-changes

The reasoning for each part of the change, and why it is correct, follows
below.

---

## The proposed invariant for polymorphs

> Every effect applied during a form change must be appropriate to the form the
> hero has **now**, not to the form the code was told to install.

Re-entrancy itself is not what causes these bugs, and it is not avoidable
here: dropping armour can put the hero in water, dropping an artifact can end
levitation, and damage taken while polymorphed reverts the form. What causes
them is that state captured before a nested form change is still used after
it: the form `polymon()` was asked to install, the form `break_armor()`
decided to strip by, and the caller that still owes the new form a light
source.

Two obligations follow, one for each side of the boundary.

**If you change the form, leave the world consistent with it before you
return.** This is the ownership question, and it is bug 07. The light source is
part of what it means to be a glowing monster, so it belongs to the code that
installs the form, not to a caller several frames up who will get around to it
later. As long as it lives in the caller, there is a window in which "the form
glows" and "a light source exists" disagree, and anything that reverts the form
inside that window is looking at an inconsistent world.

**If you continue after a call that might have changed the form, check before
you act.** This is bugs 06 and 08. Both are the same mistake at different
scales: `polymon()` continues configuring a form that is gone, and
`break_armor()` continues stripping by a form that is gone.

Neither obligation is new to this file. The DevTeam has already written both
down in places: two `FIXME?` comments say `polymon()` "should return early" if
a nested revert happened, a hand-placed `old_light = 0` patches one instance of
the light-ownership problem, and `timeout.c` deletes the light source by hand
before its own `polymon()` call. Each of those is a local workaround for one
instance of a general rule. The fix below applies the rule.

## How the fix is built

Three commits on
[`bugreport/10-polyself-reentrancy`](https://github.com/davidbau/NetHack/compare/16ff59115315917b93185d026aeefea06db9b0f4...bugreport/10-polyself-reentrancy),
branched from `16ff59115`. Each is also its own branch, and they cherry-pick
cleanly in any order, because they touch different lines and none depends on
another:

| commit | what it does | bundle |
|---|---|---|
| [`1aacf75f1`](https://github.com/davidbau/NetHack/commit/1aacf75f163477f989deb900d3567c3aa47c38b2) | five early returns in `polymon()` | 06 |
| [`e21c87732`](https://github.com/davidbau/NetHack/commit/e21c8773279d77160e29706c3bd4ef8ac4f5ada3) | move light ownership to form installation | 07 |
| [`2dc124672`](https://github.com/davidbau/NetHack/commit/2dc124672878074a8fbf2c2380e23509084df08d) | three re-checks inside `break_armor()` | 08 |

Combined: `src/polyself.c` +70/-27, `src/timeout.c` -2, no other file.
[`proposed-fix.patch`](https://github.com/davidbau/nethack-bugreport/blob/main/bugs/10-polyself-reentrant-form-changes/proposed-fix.patch) is the whole thing as one diff.

### Part 1: give the light source to whoever installs the form

Before, the only place a hero light source was created was `polyself()`'s
`made_change:` block, which runs after `polymon()` has already returned
([`polyself.c:720-730`](https://github.com/NetHack/NetHack/blob/16ff59115315917b93185d026aeefea06db9b0f4/src/polyself.c#L720-L730)).
The patch moves that transition into a helper and calls it from every place
that installs a form:

```c
/* update the hero's light source to match the form set_uasmon() has just
   installed; 'old_light' is emits_light() for the form being replaced.
   Every place which changes gy.youmonst.data calls this immediately, so
   that "the form emits light" and "a light source exists" can never
   disagree -- not even for the duration of a re-entrant callback. */
staticfn void
uasmon_light(int old_light)
{
    int new_light = emits_light(gy.youmonst.data);

    if (old_light != new_light) {
        if (old_light)
            del_light_source(LS_MONSTER, monst_to_any(&gy.youmonst));
        if (new_light == 1)
            ++new_light; /* otherwise it's undetectable */
        if (new_light)
            new_light_source(u.ux, u.uy, new_light, LS_MONSTER,
                             monst_to_any(&gy.youmonst));
    }
}
```

Fixed code: [`polyself.c:197-216`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L197-L216).

There are exactly two places that install a form, and each now captures the
old state immediately before and calls the helper immediately after:

- `polymon()`, around its `u.umonnum = mntmp; set_uasmon();`:
  [`polyself.c:824-827`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L824-L827)
- `polyman()`, around its own `set_uasmon()`:
  [`polyself.c:228-237`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L228-L237)

`polyman()` is the one to notice: `newman()` and `rehumanize()` both go through
it, so covering it covers every way back to human, including the nested
`rehumanize()` that caused the original `impossible()`.

Three things then become dead and are removed: `polyself()`'s `made_change:`
block, the `old_light = 0` workaround at `polyself.c:581`, and `timeout.c`'s
by-hand delete before `polymon(PM_GREEN_SLIME)`. `slimed_to_death()`'s comment,
which said `polymon()` "does not perform `polyself()`'s light source
bookkeeping", is updated, because now it does.

**Why this is correct.** The unsafe window existed because two facts were
maintained in two places at two times. After the patch they are maintained in
one place at one time: the helper runs in the same statement sequence as
`set_uasmon()`, with nothing in between that can be re-entered. Any observer,
including a `rehumanize()` called from deep inside `polymon()`, sees a world
where the form and its light source agree.

**What it fixes beyond the reported bug.** `polymon()` has around twenty
callers outside `polyself.c`: stone-golem petrification from `dokick.c`,
`eat.c`, `trap.c` and `uhitm.c`; `polymon(PM_QUEEN_BEE)`;
`were.c:209 polymon(u.ulycn)`; `timeout.c:493 polymon(PM_GREEN_SLIME)`. None of
them ran `made_change:`, so none of them ever managed the light source at all. A
fire vortex who caught lycanthropy kept its light source forever. Those callers
are now correct without any of them changing, which no amount of strengthening
`made_change:` could have achieved.

**The one deliberate behaviour change.** The light now appears at
`set_uasmon()` rather than after `polymon()` returns. Anything inside
`polymon()` that consults the light map now sees the new form's light instead
of the old state. That is the point: the map matches the body.

### Part 2: stop `polymon()` at each boundary that can change the form

`polymon()` had exactly one early exit, the "cannot become that" `return 0`
([`polyself.c:747`](https://github.com/NetHack/NetHack/blob/16ff59115315917b93185d026aeefea06db9b0f4/src/polyself.c#L747)).
The patch adds five, each immediately after a call that can re-enter
form-changing code:

| after | can revert because | fixed code |
|---|---|---|
| `break_armor()` / `drop_weapon(1)` | `Boots_off()` reaches `spoteffects()`; dropping an `#invoke`d levitation artifact ends levitation | [`polyself.c:902-913`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L902-L913) |
| `expels()` | ends in `spoteffects(TRUE)` | [`polyself.c:953-957`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L953-L957) |
| `dismount_steed(DISMOUNT_POLY)` | `teleds()` reaches `spoteffects()`; `instapetrify()` just above calls `polymon()` directly | [`polyself.c:989-994`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L989-L994) |
| `spoteffects(TRUE)` | `drown()` / `lava_effects()` / `dotrap()` reach `losehp()` | [`polyself.c:1001-1003`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L1001-L1003) |
| `retouch_equipment(2)` | an artifact blast reaches `losehp()` | [`polyself.c:1052-1057`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L1052-L1057) |

Each is the same two lines:

```c
    if (u.umonnum != mntmp)
        return 1;
```

**Why the test is `u.umonnum != mntmp` and not `!Upolyd`.** A nested callback
can install a *different monster* form, not only revert to human:
`instapetrify()` and `selftouch()` both call `polymon()` directly, and the
source's own comment at the fifth boundary describes exactly that case: a hero
wielding a cockatrice corpse, hit by stone-to-flesh, becoming a flesh golem and
then being turned back into a stone golem. `!Upolyd` would miss it. Asking
"is the hero still the form I was asked to install?" catches both.

**Why the return value is 1.** `polymon()` returns 1 for "the polymorph
happened", and it did: the form was installed, and then something replaced it.
`polyself()`, the only caller that inspects the result, does
`(void) polymon(...)`, so the value matters only for the direct callers that
check it, and for them "a polymorph occurred" is the truthful answer.

**Why five and not the two the `FIXME?`s mark.** The `FIXME?`s mark `expels()`
and `spoteffects()`. The other three are the same hazard documented in the
comments of the functions being called: `dropp()` says *"Dropping worn armor
while polymorphing might put hero into water"*, `drop_weapon()` says *"if heart
of ahriman is wielded, we might be losing levitation by dropping it"*,
`steed.c:747` says *"teleds() --> spoteffects() --> pickup()"*, and
`polyself.c:1022` says *"this might trigger a recursive call to polymon()"*.
Fixing only the two annotated boundaries would leave three known paths open.

**What the early returns skip, and why that is safe.** Everything after a
boundary reads `gy.youmonst.data` for a form the hero may no longer have: the
`Passes_walls` trap release, `likes_lava`, the `amorphous`/`is_whirly`/`unsolid`
chain release, web and bear trap release, `check_strangling(TRUE)`, and the
`#monster` capability hints. All of those fail *closed* for a human (a human
is not `amorphous`, does not pass walls, has no breath weapon), so skipping
them loses nothing a human should have had. What the returns do prevent is the
duplicated `encumber_msg()` and `retouch_equipment(2)` that `rehumanize()` has
already performed, which is the reported bug.

### Part 3: re-check inside `break_armor()`

`break_armor()` is a special case of the same problem one level down. It
captures the form at entry
([`polyself.c:1160`](https://github.com/NetHack/NetHack/blob/16ff59115315917b93185d026aeefea06db9b0f4/src/polyself.c#L1160))
and then asks that captured `uptr` what to strip, block by block. Its own
gloves block calls `drop_weapon(0)`, which is one of the re-entrant calls above.
So a guard placed *after* `break_armor()` returns, which part 2 does, is
necessary but not sufficient: by then the damage is done inside.

The patch adds three re-checks, at sub-block boundaries:

```c
        if (gy.youmonst.data != uptr)
            return;
```

- before the shield block, immediately after the gloves block that can revert:
  [`polyself.c:1292-1300`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L1292-L1300)
- before the boots block:
  [`polyself.c:1315-1316`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L1315-L1316)
- before the eyewear block:
  [`polyself.c:1331-1337`](https://github.com/davidbau/NetHack/blob/2dc124672878074a8fbf2c2380e23509084df08d/src/polyself.c#L1331-L1337)

**Why the boundaries and not the tests.** The obvious alternative is to replace
every `nohands(uptr)` with `nohands(gy.youmonst.data)`. That is worse. Each
sub-block pairs an `_off()` call with a `dropp()`, and re-evaluating the form
mid-block could take an item off and then decline to drop it, leaving inventory
half-updated. Returning at a boundary leaves every item either fully worn or
fully removed.

**Why `return` rather than continuing with the new form.** The caller,
`polymon()`, is itself about to stop: part 2's first guard sits immediately
after `break_armor()` returns. And the hero is human again, so the correct
amount of further stripping is none: a human can wear all of it.

## Why you should believe the three fixes are right

Four things a reviewer can check independently.

**Each patch has a recorded before-and-after on the same keystream.** The links
in "The three scenarios" above are the same seed, the same datetime and the same
keys, replayed against an unpatched and a patched build. In each pair the runs
are identical up to the moment of the defect. The RNG stream is identical too:
bug 06's witnesses draw the same `rnd(4)=1` and `d(2,8)=10` form hit points on
both sides, so each patch changes only the behaviour it targets and leaves the
rest of the game untouched.

**Each patch has a control.** Bug 06 ships two routes into the same defect
(`expels()` and `spoteffects()`) and two controls that differ by one ingredient
and blast only once, which separates "sequential call" from "nested call" and
shows the existing recursion guard in `retouch_equipment()` is not what is
broken. Bug 08's patched run still prints `The lava here burns you!` and takes
`d(6,6)`, proving the patched build reached the same lava rather than avoiding
the scenario.

**The three are independent.** Each fix leaves the other two symptoms intact,
and the three commits cherry-pick cleanly onto the pinned commit in any order.
Bug 06's early returns specifically do *not* fix bug 07: that assertion fires
inside `rehumanize()`, several frames below, before `polymon()` gets the chance
to test anything. If they were three views of one defect, one patch would have
silenced all three; it does not.

**Each guard sits where the source already says the hazard is.** Every one of
the five boundaries in part 2, and the entry point in part 3, is a call whose
own comments document that it can re-enter form-changing code. The patch adds no
new theory about what is dangerous; it acts on what the file already says.

## Not fixed here

Two things stay open deliberately.

`break_armor()` reports nothing to its caller about having been invalidated.
Part 3 stops it from doing further harm and part 2 stops `polymon()` afterwards,
so no path is left broken, but the cleaner shape would be for `break_armor()` to
return a "form changed under me" result rather than have both sides infer it.

The wider question of whether a form change should be a transaction at all,
with the consequences applied as a unit that either completes or unwinds, is a
much larger design change. These patches make the existing sequence honest
about its own re-entrancy instead.

## Credit

(Found and analysed by AI agents collaborating on a JavaScript port of NetHack
5.0, under human direction. The recordings, the root-cause analysis and the
patches are Claude Opus 5's; the widened boundary audit, the
`u.umonnum != mntmp` form of the test, and the form-installation-boundary
structure for the light source were proposed in review by Codex GPT-5.6.)

---

Reported as #1682.

-- David Bau (@davidbau)


davidbau:
Updated the PR source branch with commit 97b843554. The reentrancy checks now use a monotonic `uasmon_generation` counter, incremented at the end of every `set_uasmon()` installation, including same-form reinstallations. `polymon()` and `break_armor()` compare saved generations at their reentrant boundaries, closing the reachable ABA route through trap-triggered `polyself()`.

The public formal verification report is here:
https://github.com/davidbau/nethack-bugreport/blob/main/bugs/10-polyself-reentrant-form-changes/REPORT.md

